PR #54, opened 2013-03-05
AccessType("public_method") ignored when deserializing to XmlList, etc.

rh389:
XmlDeserializationVisitor::visitProperty uses reflection to set properties unless none of xmlAttribute, xmlValue, xmlCollection are set. For any remaining properties, reflection will only be used if metadata->setter is null. The latter logic seems more desirable for all cases.

This is presumably a bug? I can't see any reason the following configuration should not be using setItems when deserializing from XML:

```
class Order {
    /**
     * @Type("ArrayCollection<Item>")
     * @XmlList(entry = "item", inline = true)
     * @AccessType("public_method")
     */
    private $items;

    public function setItems($items){
        $this->items = $items;
        //Do something else
    }
}
```

I ran into this when attempting to make sure my Items were aware of my Orders, as Item was the owner of a Doctrine association. Attached is a failing test and fix.


goetas:
was fixed with https://github.com/schmittjoh/serializer/pull/665

PR #73, opened 2013-04-02
Add the PRE_DESERIALIZE event to the Events class

adrienbrault:
Hey, I've just noticed that this event was missing from the Events class


nass600:
Hello there!

I noticed pre_deserialize callback_methods are not available when using Yaml notation.

Is this a bug, an intended behaviour or a missing feature?

PR #89, opened 2013-05-10
XML deserialization visitor does not allow setter accessor with XmlCollection, XmlAttribute, XmlValue. 

zebraf1:
There is an issue with XmlDeserializationVisitor visitProperty method. When I have defined AccessType("public_method") and Accessor(getter="getTestSubjects", setter="setTestSubjectsTest") with XmlList(inline = true, entry = "comment") it will not call setter and will access the property directly through reflection class.

here's my test case which will not pass without this fix: https://gist.github.com/zebraf1/5553947

Couldn't find deserialization test cases, how come?


goetas:

zebraf1:
No problem for me, I have moved to other projects.

PR #110, opened 2013-07-05
Deserialization fail

tquiroga:
In different cases, when JSON is sent this error occurs:
`Warning: array_key_exists() expects parameter 2 to be array, string given in`
or
`Warning: array_key_exists() expects parameter 2 to be array, integer given in`


rdohms:
I'm having this issue related to a specific implementation.

property of objA->token holds instances of objB but when data comes from unserialization what comes in is the "id" string.

I setup an accessor to catch this and instantiate an ObjB but it never gets that far cause this error pops up in the visitor before.

Any ideas?

nmeirik:
+1

adityaseth09:
+1

Jerome1337:
+1

goetas:
solved with https://github.com/schmittjoh/serializer/pull/398

PR #120, opened 2013-07-10
Added PreSerializeEvent to the pre-serialize call-back.

markrunaway:
I needed access to the context in the pre-serialize call-back. The event listeners have access to the context so I thought it reasonable that the call-backs should too?


goetas:

PR #126, opened 2013-07-24
make FormErrorHandler more easily extendable

Briareos:
I wanted to override FormErrorHandler so it would return error messages in the format

```
error:
    message: Parsed error message
    template: Error message template
    parameters: Error message parameters
```

instead of

```
error: Parsed error message
```

and it's pretty hard to do with private members.

### DIFF
--- a/src/JMS/Serializer/Handler/FormErrorHandler.php
+++ b/src/JMS/Serializer/Handler/FormErrorHandler.php
@@ -29,7 +29,7 @@ use JMS\Serializer\XmlSerializationVisitor;
 
 class FormErrorHandler implements SubscribingHandlerInterface
 {
-    private $translator;
+    protected $translator;
 
     public static function getSubscribingMethods()
     {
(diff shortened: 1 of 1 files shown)


tonicospinelli:
+ 1 pls :heart_eyes:

tonicospinelli:
I had to duplicate all code from `FormErrorHandler` just to change how to `convertFormToArray` works, just because those methods and property are private :cry:

goetas:

@schmittjoh i would like to consider re-integrating this soon. Are you ok with it?

goetas:
Not planning to create hidden "dependencies" allowing people to extend "internal" classes.

If you need to create a different form serialization, please implement your form handler. Consider https://en.wikipedia.org/wiki/Composition_over_inheritance

Briareos:
Fully agreed. This PR was from a long time ago :)

goetas:
Sorry for the incredibly long feedback loop. Few months ago I started to do some clean up

PR #135, opened 2013-08-26
Make tests fail for DateTime fields in U format with negative dates

adanlobato:
When dates are prior to Unix Epoch (January 1 1970), timestamps are negative.

When `DateHandler` performs a `DateTime::createFromFormat('U', $negativeTimestamp)` false is returned and and exception is thrown. It seems a PHP bug not supporting these negative timestamps on the `createFromFormat` method.

If you try this example, it works perfectly.

``` php
$a = DateTime::createFromFormat('Y-m-d H:i:s', date('Y-m-d H:i:s', -315619200));
```

I made tests fail, do you think we can create a workaround to patch this PHP issue?

### DIFF
--- a/tests/JMS/Serializer/Tests/Fixtures/Author.php
+++ b/tests/JMS/Serializer/Tests/Fixtures/Author.php
@@ -29,13 +29,24 @@ class Author
      */
     private $name;
 
-    public function __construct($name)
+    /**
+     * @Type("DateTime<'U','UTC'>")
+     */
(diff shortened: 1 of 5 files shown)


evillemez:
Has anyone found a workaround to dealing with negative timestamps and using the `DateTime('U')` format?

goetas:
fixed in php  5.6.24  https://3v4l.org/StV1g

PR #138, opened 2013-08-31
Fixed problem with Warning: array_key_exists() expects parameter 2 to be...

kacperix:
When property of object is type of antoher object and input data is not array then warning of array_key_exists is thrown. 

I've added checking is data is array.
It has been fixed some time ago (https://github.com/thomasquiroga/serializer/commit/a636e3c03485a81388fc2f8b374e33e080df3e20) but in some way gone.


goetas:

Duplicate of #110

Issue #150, opened 2013-09-12
Specify group along with MaxDepth


adrienbrault:
Hey,

I was just thinking of the current limitations, and I think that if you can include/exclude fields based on group and also limit their serialisation depth with groups (`@MaxDepth(..., groups={})` ?), it would be way more flexible.

WDYT ?

goetas:
https://github.com/schmittjoh/serializer/issues/170 is merged now

PR #170, opened 2013-10-18
Can now override groups on specific paths of the graph

adrienbrault:
This is backward compatible, and pretty awesome IMHO :heart_eyes:

Inspired from one of the examples on #150


lopsided:
+1 This is great. Will this work with yml mapping files too?

smilesrg:
+1 for this!

Seb33300:
Any news about this?
Will it be merged?

fadamsen:
I was looking for a way to do exactly this, so +1 from me as well.

dvitharanage:
+1 need this feature for large object graphs of related entities. It should be possible to override child object groups from the parent property. Why isn't this merged yet????

goetas:
Finally this is on master

PR #176, opened 2013-11-06
Throw JsonErrorException when `json_encode()` fails

Josiah:
Currently when the PHP built-in method json_encode() as part of the `JsonSerializationVisitor` fails it does so silently and returns false instead of the JSON encoded string.

To make it simpler to debug situations which cause this error I have added a `JsonErrorException` which is thrown containing the error reported by json_last_error().

New unit test supplied which covers added code, all tests passing.


Josiah:
This PR is ancient, cleaning it up

PR #177, opened 2013-11-06
Added the support for class inheritance and skipping handlers

shiroyuki:
(Originally submitted as #175)

This PR is to add the support to allow the handler to tell the graph navigator to skip itself and proceed with the normal procedure. For instance,

``` php
use PhpOption\None;

class MyHandler implements SubscribingHandlerInterface
{
    public function serializeProxy(JsonSerializationVisitor $visitor, $entity, array $type, Context $context)
    {
        if ($visitor->getRoot() === null) {
            throw new SkipStepException('Skip This');
        }

        return array(/* ... */);
    }
}
```

Also, this PR _supports the class inheritance_ such that we can specify a parent class in `getSubscribingMethods`. For instance,

``` php
class MyHandler implements SubscribingHandlerInterface
{
    {
        return array(
            array(
                'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
                'format' => 'json',
                'type' => 'EntityParentClassOrInterface',
                'method' => 'serializeProxy',
            ),
        );
    }

    public function serializeProxy(JsonSerializationVisitor $visitor, $entity, array $type, Context $context)
    {
        /* ... */
    }
}
```

### DIFF
--- a/src/JMS/Serializer/Exception/SkipStepException.php
+++ b/src/JMS/Serializer/Exception/SkipStepException.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * Copyright 2013 Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace JMS\Serializer\Exception;
+
+/**
+ * @author Kinn Coelho Julião <kinncj@gmail.com>
+ * @author Juti Noppornpitak <shiroyuki@gmail.com>
+ */
+class SkipStepException extends RuntimeException
+{
+}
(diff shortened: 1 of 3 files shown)


bdecarne:
Johannes,

This post is kind of old, but about class inheritance : the @Type annotation could do the job, but it can't be used on the class level. If i try to serialize the object, i cant use the handler define for the parent class...

I have a master class FormFlow, and i try to serialize a object instance of CreateItemFormFlow, subclass of FormFlow. How can i achieve that ?

alcalyn:
By using yaml mapping format ?
Couldn't you create a yaml file in your project which map the class in the "vendor" project ? I know it is possible

goetas:

goetas:
re opening to keep track of it as feature request

goetas:
@Majkl578 what do you think about `SkipStepException` ? 
Whar said by @schmittjoh 
>Similarly, the SkipStepException can be solved by calling accept with a different type name. You could for example, use any arbitrary string as type which your handler responds to and then call accept again with a different type from within your handler. This should also be faster than throwing/catching an exception.

is not true, as the object is already in visiting state, so another call to `accept` will just return null

goetas:
The part of "skipping handlers" has been implemented in https://github.com/schmittjoh/serializer/pull/1238 (it was already possible to implement something as that in the past, but it requited a two-steps approach, while now can be done in a single one).

Regarding the inheritance, I think that @schmittjoh answer it explains how to use the `@Type` annotation to achieve it.

> If you would like to catch a base class with a handler, you can already annotate the respective properties with the @Type annotation hinting the base class and register a handler for the base class. The current handler system was meant as a simple key-value lookup table for fast access. The strings do not necessarily have to correspond to class names.

PR #186, opened 2013-11-26
change inheritance of handlerCallbacks

KampfCaspar:
status quo:
Pre/Post serialization hooks are combined for extended classes, while handlerCallbacks always get overwritten.

proposal:
let handlerCallbacks be inherited if not overwritten (à la JsonSerializer). If overwritten, overwrite handlerCallbacks in toto.


goetas:

PR #187, opened 2013-11-27
No CData

mvrhov:
I've defaulted the cdata to true to keep BC with existing serializer metadata. Although it's a bit annoying setting it to false on most of the string properties.

Added visitSimpleString into the XmlSerialization visitor, so Handlers can use that when serializing into the xml if they know that what is beeing serialized doesn0t need to be inside cdata

This one replaces #119


noetix:
Without `cdata` annotation:

```
<FromSiteName>
    <![CDATA[ K & S Perth ]]>
</FromSiteName>
```

With `cdata=false`:

```
<FromSiteName>K & S Perth</FromSiteName>
```

But expected: 

```
<FromSiteName>K &amp; S Perth</FromSiteName>
```

mvrhov:
Yes, that's expected. please read above. You must specify cdata attribute where invalid xml text is expected.

PR #192, opened 2013-12-03
Deserialisation with DomDocument

ElectricMaxxx:
Hi 
i did it i. 
I completely  replaced the `XmlDeserializationVisitor` (and all methods) the a way that uses the DomDocument instead of the SimpleXmlElement. By this, i think we will have more power in working with a given XML-String and won't get trouble with strings and they line endings, cause i case of `visitString()` i do realy use the text as a `DomTextNode` instead of `(string) $data;`
On the way i changed the needed methods inside of the DateHandler to deserialize DateTime AND (new now) DateInterval objects. This work cool.
I added some comments to all scripts i touched, my bad english is the only thins we need to refactore, but i think some comments are better than no comments ;-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,2 @@
 vendor/
 phpunit.xml
-.idea/
-composer.phar
-composer.lock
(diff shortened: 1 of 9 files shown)


goetas:

(BTW the idea looks great)

PR #197, opened 2013-12-04
access to the data

danielsan80:
Access to the data could be usefull.
What other way to change the data after serialization?


mowcixo:
I just added schmittjoh/serializer#466 related with this.

goetas:

Let's go with https://github.com/schmittjoh/serializer/pull/466 for now

PR #203, opened 2013-12-14
Avoided to get circular reference for empty object in array

spolischook:
If I have tried to serialize array with empty objects (or serialized map exclude all properties), I catch the circular reference.


goetas:

PR #51, opened 2013-02-27
Treat \stdClass as an array (fixes #43 and rel.)

aeoris:
Fixes the following open issues on JMSSerializerBundle as well:
* https://github.com/schmittjoh/JMSSerializerBundle/issues/158
* https://github.com/schmittjoh/JMSSerializerBundle/issues/271

PR #211, opened 2014-01-09
Treats stdClass as array

parnas:
Clearer than https://github.com/schmittjoh/serializer/pull/51 .


alexdesignworks:
Not sure why the test failed, by the patch in commit does solve the problem.

goetas:

Closing the PR, see https://github.com/schmittjoh/serializer/pull/429#issuecomment-238501918 for the reason

PR #237, opened 2014-02-18
XmlList and XmlMap Namespace support

skler:
Adding support for namespace param for XmlList and XmlMap

@XmlList(entry="event", inline = true, namespace="http://mynamespace.net" )


goetas:
this is obsolete since https://github.com/schmittjoh/serializer/pull/301 has been merged

PR #238, opened 2014-02-18
Serialization of nested polymorphic objects

DavidMikeSimon:
Serializer has trouble with nested objects of polymorphic types; it serializes them as their stated base type in an annotation, rather than their real type.

This PR adds tests which exhibit this bug, then adds an additional pre-serialize event subscriber that corrects the issue.


teu:
@schmittjoh Good work guys, a release would be great, lack of this feature causes tons of incidental complexity.

PR #252, opened 2014-03-17
Fix for deserializing xml with unprefixed children in the default namespace.

mlively:
Given the following xml:

```
<test-object xmlns="http://example.com/namespace" xmlns:v2="http://example.com/namespace/v2" color="green">
    <name>Test Name</name>
    <description>Test Description</description>
    <v2:status>New</v2:status>
</test-object>
```

attempts to deserialize the object will result in name and description being null. As near as I can tell, this is due to xpath in simplexml not being able to lookup unprefixed namespaces in element names. Basically xpath('./name') will return empty.

You can fix this by registering the namespace for the xpath with registerXPathNamespace. This is already being done for namespaces declared outside of the root node. I just added logic to also do this for the default namespace where the prefix yields "".

I did verify (with the included tests) that attributes do not exhibit this same issue therefore I did not make changes to that block in the visitor.


goetas:

This has been fixed with 1.2.0

PR #253, opened 2014-03-23
Implemented PathBasedExclusion

adam-g-pullen:
This allows the user to annotate their base root class to exclude
complex paths.

For example given the following graph

``` php
User
 -> Company
   -> Users
     -> Groups
      -> Permissions
        -> Groups
   -> Groups
     -> Permissions
        -> Groups
 -> Groups
   -> Permissions
... and so on
```

The current state of the serializer does an excellent job of prevent recursion. However I am not interested in pulling all the data from the graph.

With this change you can tell the serializer to ignore object paths

``` php
use JMS\Serializer\Annotation as JMS;
/**
 * FinalConcept\TimeTracker\EntitiesBundle\Entity
 *
 * @ORM\Table(name="users")
 * @ORM\Entity(repositoryClass="FinalConcept\TimeTracker\EntitiesBundle\Repository\UserRepository")
 * @JMS\PathBasedExclusion({"company.users", "company.teams", "company.groups", "groups.users", "groups.permissions.groups", "teams.company", "teams.users"})
 */
class User implements UserInterface, \Serializable
{
//..
}.
```

which will give you a nice neat graph like so

``` php
User
  -> Company
  -> Groups
   -> Permissions
```

PathBasedExclusion is designed to work on the root object of the graph. Any PathBasedExclusion annotations found on objects lower down in the graph will be ignored.

Therefore if you want to continue ignoring properties from lower objects you need to add them to the root object

### DIFF
--- a/src/JMS/Serializer/Annotation/PathBasedExclusion.php
+++ b/src/JMS/Serializer/Annotation/PathBasedExclusion.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * Copyright 2014 Adam Pullen <adam@finalconcept.com.au>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


goetas:
Currently there are too many exclusion policies. Since https://github.com/schmittjoh/serializer/pull/673 is merged, and allows a really flexible exclusion strategy, i suggest to use it.

PR #258, opened 2014-04-08
Able to convert properties to lowerCamelCase by default

vslinko:


vslinko:
@schmittjoh ping
I'll close this PR after 2 weeks.

schmittjoh:
Let's close this. You can easily and fast implement something like this in your own application.

PR #259, opened 2014-04-08
fix two dev dependencies that could lead to errors when running the tests

xabbuh:
There were two issues with the old composer.json if you ran ``composer update``:

* The implementation of ``Doctrine\ORM\Proxy\Proxy`` changed between ``doctrine/orm`` 2.3 and 2.4 (doctrine/doctrine2@8272ffd23fd66bac13a0dc074c868a00b82c7707) which caused the ``SimpleObjectProxy`` fixture class not being valid any more.

* The constructor of the ``ConstraintViolation`` class got an additional argument ``$message`` with Symfony 2.2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "propel/propel1": "~1.7",
         "symfony/yaml": "2.*",
         "symfony/translation": "~2.0",
-        "symfony/validator": "~2.0",
+        "symfony/validator": "~2.2",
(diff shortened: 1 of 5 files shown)


goetas:
This PR is probably outdated and many tings are already solved on master. Probably only the changes on `SimpleObjectProxy` can be relevant. 

If you are still interested, please create a new PR

PR #261, opened 2014-04-09
Set accessors only for exposed properties

dmikhinev:
Annotation driver shouldn't check accessors for excluded properties


goetas:
was solved with https://github.com/schmittjoh/serializer/commit/b7b1d80f3695f7f874c11438f9bb601e86065de1

PR #266, opened 2014-05-02
add the ability to mark a property as write only so it can be desrialized to but not be serialized from

yjv:
im using the serializer to desrialize data from an http request but then i want to show this same object as a return from my application without some tf the fields that were deserialized to. so being able to mark them as writeonly would be really nice.


goetas:

PR #269, opened 2014-05-09
Missing attribute 'xml-value' in XML Reference

holtkamp:
As described for the annotations here: http://jmsyst.com/libs/serializer/master/reference/annotations#xmlvalue

Still, an XML Schema (.xsd) would be nice, now had to check the test-cases to find this out. Of course, time is scarce for everyone ;)


goetas:
Tnx!

PR #279, opened 2014-06-04
Support for Embedded Classes

kibao:
Set proper property type if property is embedded class.


jhogendorn:
Is there any chance of this getting merged in? Lack of support for embeddables is problematic for me.

goetas:
Unfortunately with no tests, this can not be merged

PR #284, opened 2014-06-20
Deserialization 'Array to string conversion' better error

tenerd:
Sometimes I do mistake and have @Type("string") for in reality array data. When i run my tests, i get common php notice 'Array to string conversion' without any trace to current object etc (only stacktrace to ... vendor\jms\serializer\src\JMS\Serializer\GenericDeserializationVisitor.php).
I added LogicException with current objects, which hints me source of my error more accurately.


goetas:

Closing the PR, see https://github.com/schmittjoh/serializer/pull/591#issuecomment-238364295

Issue #283, opened 2014-06-19
Erroneous data format for unserializing...


Ninir:
Hi guys,

Since PHP `5.4.29` and `5.5.13` versions are out, a [BC break](http://news.php.net/php.internals/73957) affects the serialization. While it is fixed in Doctrine2 and PHPUnit, it seems that an issue is related to this serializer.

##### Configuration

I am using PHP `5.4.29` and the following config:

```json
"require": {
    "symfony/symfony": "2.3.*",
    "doctrine/orm": "2.4.*",
    "jms/serializer-bundle": "dev-master",
    ...
}
```

I'm using [84f8ef5ca46f02f3001af47ba25fc917c776a247](https://github.com/doctrine/doctrine2/commit/84f8ef5ca46f02f3001af47ba25fc917c776a247) from [branch 2.4](https://github.com/doctrine/doctrine2/commits/2.4) for Doctrine2.
Logging works well in this case. The issue is more about using the Serializer than the rest.

Note that all vendors are up-to-date. Caches are cleared.

##### Test cases

I made 2 tests directly from a SF2 controller. The first is passing, the second is not.

###### Passing test
```php
        $dates = array();
        $dates[] = new \DateTime('-7 days');
        $dates[] = new \DateTime('-6 days');

        $serializer = $this->get('jms_serializer');
        $serializedDates = $serializer->serialize($dates, 'json');
        var_dump($serializedDates);
        var_dump($serializer->deserialize($serializedDates, 'array<DateTime>', 'json'));
        exit;
```

Here, the PHP script stops and successfully displays the 2 var_dump(..) calls.

###### Failing test
```php
        $classes = array();
        $classes[] = new File(__DIR__ . '/TestController.php');
        $classes[] = new File(__DIR__ . '/Controller.php');

        $serializer = $this->get('jms_serializer');
        $serializedClasses = $serializer->serialize($classes, 'json');
        var_dump($serializedClasses);
        var_dump($serializer->deserialize($serializedClasses, 'array<Symfony\Component\HttpFoundation\File\File>', 'json'));
        exit;
```

Here, the PHP script fails at the second var_dump(), throwing:

```
ContextErrorException: Warning: Erroneous data format for unserializing 'Symfony\Component\HttpFoundation\File\File' in /path/to/project/vendor/jms/serializer/src/JMS/Serializer/Construction/UnserializeObjectConstructor.php line 29
```

Do you guys have any hint on this please? (ping @schmittjoh @stof)

[...]

PR #286, opened 2014-06-27
Fix for issue #283

akramfares:
This is a fix for the bug create by the update of PHP 5.4.28 to 5.4.28
This fix is inspired from the one made for PHPUnit: https://github.com/sebastianbergmann/phpunit-mock-objects/commit/1c68f1338f1940deb8265428bb2a7cbc5bc074b5#diff-64dbbc1c21f9be2d92e2b715617ffe34


goetas:

The problem was solved long time ago and now only 5.5 is supported

PR #292, opened 2014-07-04
Fix test suite on travisci

l3l0:
Bascially this PR fixed build at the travisci.

Just be aware about few things:

* I need to skip  testWhitelistedDocumentTypesAreAllowed() test for some version of php cause of  https://bugs.php.net/bug.php?id=67081 It seems that serializer rely on some invalid php behavior which was fixed in newer version of php (>= 5.5.13 and >= 5.4.29). I don't know how to fix this in code, if someone have idea feel free to comment and help me :)
* Library support php 5.3 as we can see in travis.yml - so I need to fix code like `$this` usage in anonymous functions.

Give me know if this PR is acceptable, if yes I can squash commits.


urakozz:
@schmittjoh I think you can close this PR now )

schmittjoh:
Yes, thanks :)

Issue #277, opened 2014-05-26
Serializing array that have one of the element or member of an element an empty object


gogume:
When serializing an array that have one of the element or member of an element an empty object the whole array is transformed into an ArrayObject, is this the expected result?
I've wrote a little test here: 
https://github.com/gogume/serializer/tree/array-with-empty-object
This branch is started from the master.

goetas:
Closing, Sorry for the long feedback loop. If the problem is still here, feel free to comment (pointing out the code that presents the issue or providing a test case) it and I will re-open the issue.

PR #293, opened 2014-07-07
Fix invalid serialization of array with empty object

l3l0:
Try to fixed issue with json serialization  of array with empty object.
It was serialized like `{'0': {}}` and it should be serialized as `[{}]` this PR should fix it without break all other behavior.

Fix issue #277 related with #104

Rely on changes done in #292


goetas:
This is probably outdated, sorry for the terribly long feedback loop

Closing the PR. If you are still interested in working on this feel free to re-open a new PR based on the current master

PR #294, opened 2014-07-08
Fixed handling of un-typed collection references from Doctrine Mongo ODM

gpoole:
The MongoDB ODM supports references where the type of the referenced entity is stored in the DBRef object in the database, so the type information is not available in metadata. When this happens, getAssociationTargetClass returns null and this fix tries to handle that.


goetas:

Closing the PR, since too many things changed on master in the last years. If you are still interested in working on this feel free to re-open a new PR based on the current master

PR #306, opened 2014-07-21
setter ignored in XmlDeserializationVisitor for xmlAttribute and xmlValue

tmilos:
If you mark a property with XmlAttribute and Accessor getter/setter, during serialization the getter method will be called, but during deserialization setter method is not called. Setter is also not called in case property has XmlValue annotation.

Example:

``` php
class Assertion
{
    /**
     * @var int
     * @JMS\Type("string")
     * @JMS\Accessor(getter="getIssueInstantString", setter="setIssueInstant")
     * @JMS\SerializedName("IssueInstant")
     * @JMS\XmlAttribute
     */
    protected $issueInstant;

    /**
     * @return string
     */
    public function getIssueInstantString()
    {
        return Helper::time2string($this->issueInstant);
    }

    /**
     * @return string
     */
    public function getIssueInstantTimestamp()
    {
        return $this->issueInstant;
    }

    /**
     * @param string|int|\DateTime $issueInstant
     * @throws \InvalidArgumentException
     * @return $this|Assertion
     */
    public function setIssueInstant($issueInstant)
    {
        if (is_string($issueInstant)) {
            $this->issueInstant = Helper::parseSAMLTime($issueInstant);
        } else if ($issueInstant instanceof \DateTime) {
            $this->issueInstant = $issueInstant->getTimestamp();
        } else if (is_int($issueInstant)) {
            $this->issueInstant = $issueInstant;
        } else {
            throw new \InvalidArgumentException();
        }
        return $this;
    }

}
```

XML for deserialization

``` xml
<Assertion IssueInstant="2014-07-21T14:37:28Z">
</Assertion>
```

Test

``` php
class AssertionTest extends \PHPUnit_Framework_TestCase
{
    public function testDeserialization()
    {
        $xml = <<<EOT
<Assertion IssueInstant="2014-07-21T14:37:28Z">
</Assertion>
EOT;
        AnnotationRegistry::registerLoader('class_exists');
        $serializer = \JMS\Serializer\SerializerBuilder::create()->build();
        $assertion = $serializer ->deserialize($xml, 'Assertion', 'xml');

        // actual value is "2014-07-21T14:37:28Z" directly set by reflection, rather then setter method called
        $this->assertEquals(1405953448, $assertion->getIssueInstantTimestamp());
}
```

Suggested fix:
Class `JMS\Serializer\XmlDeserializationVisitor`
Method: `visitProperty(...)`
When value `$v` is obtained do

``` php
        if (null === $metadata->setter) {
            $metadata->reflection->setValue($this->currentObject, $v);
            return;
        }
        $this->currentObject->{$metadata->setter}($v);
```

instead of just

``` php
$metadata->reflection->setValue($this->currentObject, $v);
```

like it's done in the fallback on the end of that method.

Eventually, extract those few lines into separate method and prevent duplication of code.

### DIFF
--- a/src/JMS/Serializer/XmlDeserializationVisitor.php
+++ b/src/JMS/Serializer/XmlDeserializationVisitor.php
@@ -224,12 +224,12 @@ class XmlDeserializationVisitor extends AbstractVisitor
                 $nodes = $data->xpath('./@'.$attributeName);
                 if (!empty($nodes)) {
                     $v = (string) reset($nodes);
-                    $metadata->reflection->setValue($this->currentObject, $v);
+                    $this->setPropertyValue($metadata, $v);
                 }
 
             } elseif (isset($data[$name])) {
                 $v = $this->navigator->accept($data[$name], $metadata->type, $context);
-                $metadata->reflection->setValue($this->currentObject, $v);
+                $this->setPropertyValue($metadata, $v);
             }
 
             return;
@@ -237,7 +237,7 @@ class XmlDeserializationVisitor extends AbstractVisitor
 
         if ($metadata->xmlValue) {
             $v = $this->navigator->accept($data, $metadata->type, $context);
-            $metadata->reflection->setValue($this->currentObject, $v);
+            $this->setPropertyValue($metadata, $v);
 
             return;
         }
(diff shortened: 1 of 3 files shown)


goetas:
replaced by https://github.com/schmittjoh/serializer/pull/665 that take advantage of new metadata setter

PR #307, opened 2014-07-21
Fixed issues when using DoctrineObjectConstructor in combination with xm...

SavageTiger:
When deserializing enities from XML format the DoctrineObjectConstructor recieved an SimpleXMLElement assuming it was a proxy object it send it back to Doctrine throwing an EntityNotFound exception.

I fixed it, and added the missing code needed to update existing entities in case the source format was XML.


goetas:

But I agree with @stof to do not include this in the main code

PR #311, opened 2014-07-28
Optional setting to deserialize bigint as strings

tenerd:
Currently im not able to deserialize json with big integers, 
when json contains
{ 'big-int': 1000007546090751}
its deserialized as '1.0000076544307E+15'


goetas:

Closing the PR, this is already possible specifying  JSON serialization options

PR #309, opened 2014-07-25
add / discriminator for classes in the middle of hierarchy

scasei:

PR #296, opened 2014-07-08
Overwrite discriminator settings in metadata if not present in parent class

anho:
Same problem as in #182.

Issue #287, opened 2014-06-28
SimpleObjectProxy is not implement abstract methods of Proxy class


empire:
`Serializer` is dependent on  `doctrine/orm ~2.1`, `Doctrine\Common\Proxy` declare 6 abstract methods , but 2 of them is implemented in SimpleObjectProxy. 

Following error is occur when I test my application.

PHP Fatal error:  Class JMS\Serializer\Tests\Fixtures\SimpleObjectProxy contains 6 abstract methods and must therefore be declared abstract or implement the remaining methods (Doctrine\Common\Proxy\Proxy::__setInitialized, Doctrine\Common\Proxy\Proxy::__setInitializer, Doctrine\Common\Proxy\Proxy::__getInitializer, ...) in /.../vendor/jms/serializer/tests/JMS/Serializer/Tests/Fixtures/SimpleObjectProxy.php on line 41

goetas:
solved with https://github.com/schmittjoh/serializer/commit/f9bb48b14e879ed188708308acc4a6f2ddffc494

Issue #182, opened 2013-11-19
Deserialization fails when discriminator base class extends another class


gregory90:
This code:

```php
abstract class Thing {}

/**
* @Serializer\Discriminator(field = "type", map = {
*    "car": "Car",
*    "moped": "Moped",
* })
**/
abstract class Vehicle extends Thing
{
    /** @Serializer\Type("integer") */
    public $km;

    public function __construct($km)
    {
            $this->km = (integer) $km;
    }
}
    
class Moped extends Vehicle
{
}
    
class Car extends Vehicle
{
}

$jsonData = '{"km":5,"type":"car"}';
$serializer = JMS\Serializer\SerializerBuilder::create()->build();
    
$object = $serializer->deserialize($jsonData, 'Vehicle', 'json');
```

Fails with:

    PHP Fatal error: Cannot instantiate abstract class Vehicle in .../jms/serializer/src/JMS/Serializer/Construction/UnserializeObjectConstructor.php on line 28

Am I missing something?

goetas:
solved by https://github.com/schmittjoh/serializer/pull/382

PR #315, opened 2014-08-03
Tests: Bugs with discriminator inheritance on Doctrine models

xoob:
- Fixes a few units tests so I could run them locally
- Adds tests for a bug with the inheritance of the discriminator property if the doctrine classes extend a non-entity base class
- Adds test for a bug where the discriminator field name does not have the naming strategy applied to it

**Inheritance Bug**

Discriminator settings are not inherited if the if the parent entity class extends a generic base class, for example:

```php
<?
/**
 * Non-entity base class
 */
abstract class AbstractModel {}

/**
 * @ORM\Entity
 * @ORM\InheritanceType("SINGLE_TABLE")
 * ...
 */
abstract class Person extends AbstractModel {}

/**
 * @ORM\Entity
 */
class Student extends Person {}
```

On the other hand, if `Person` does not extend any parent class, the inheritance works as expected:

```php
<?
/**
 * @ORM\Entity
 * @ORM\InheritanceType("SINGLE_TABLE")
 * ...
 */
abstract class Organization {}

/**
 * @ORM\Entity
 */
class School extends Organization {}
```

**Naming Bug**

When the discriminator is automatically added from Doctrine annotations, the field name will not have the naming strategy applied to it. `@ORM\DiscriminatorColumn(name="orgType")` will result in `{"orgType":"school","camel_case":"foo"}` instead of the expected `{"org_type": ...}`.

**Related Issues**

From what I could find, this is related to #166, #182, #287, #296, #309


xoob:
Re-opened as #382 with the minimum changes only

PR #317, opened 2014-08-08
Default `$serializeNull` to false

steveYeah:
`JMS\Serializer\Context::serializeNull` is a boolean, so should default to false, not null.

### DIFF
--- a/src/JMS/Serializer/Context.php
+++ b/src/JMS/Serializer/Context.php
@@ -52,7 +52,7 @@ abstract class Context
     private $exclusionStrategy;
 
     /** @var boolean */
-    private $serializeNull;
+    private $serializeNull = false;
 
     private $initialized = false;
 


schmittjoh:
Please check if this is used by the RestBundle to set some default if not explicitly set (@lsmith77 might know this too).

goetas:
Had a look this week in RestBundle repos, no direct reference to `$serializeNull === null`.

Considering it as safe to merge

bendavies:
@schmittjoh @goetas  
this breaks FOS rest bundle in 1.8 at least.
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.8/View/ViewHandler.php#L308

goetas:
are you sure?  here it looks they are using their own context in https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.8/Context/Context.php and later is converted into a JMS context by the adapter https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.8/Context/Adapter/JMSContextAdapter.php

and since `(bool)null ===false` it should work anyway.

is there a failing build to check?

bendavies:
yes i'm 100% sure as this has just broken my projects functional tests.
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.8/View/ViewHandler.php#L308
->
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.8/Util/ContextHelper.php#L67
not returns `false` instead of `null`.

The conversion is done later.

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.8/View/ViewHandler.php#L501-L504

goetas:
i see now :(

PR #318, opened 2014-08-11
Only parse direction once per direction

jorns:
Solves an InvalidArgumentException when defining two serializers for one
direction:
```yml
// ..
    handler_callbacks:
        serialization:
            json: serializeToJson
            json_ld: serializeToJsonLd
```


goetas:
will be solved with https://github.com/schmittjoh/serializer/pull/515

PR #338, opened 2014-09-17
GraphNavigatorInterface

xavierlacot:
Several visitor classes depend on the `JMS\Serializer\GraphNavigator` class, which is a `final` class. Switching the signatures of these visitors so that they to depend on a `GraphNavigatorInterface` instead of the class itself, allows more flexibility, and makes the `GraphNavigator` component replaceable.

Closes https://github.com/schmittjoh/JMSSerializerBundle/issues/238


goetas:
planned for 1.5

goetas:
done in https://github.com/schmittjoh/serializer/pull/876

Issue #319, opened 2014-08-12
"Can't pop from an empty datastructure" error when multiple serializer calls


damienalexandre:
I need to call the serializer on postSerialize from another serializer call (`serialize` recursion) but I get this error:

```
Can't pop from an empty datastructure
500 Internal Server Error - RuntimeException
vendor/jms/serializer/src/JMS/Serializer/GenericSerializationVisitor.php at line 127 
```

Spent a bit of time searching for the bug and I do not understand. The context is supposed to be fresh each time we call `serialize`, but calling it again in an event manage to corrupt it.

I managed to reproduce a test case with this error (did't do a PR as there is an unknown issue breaking the other tests with this one):

``` php
<?php

namespace JMS\Serializer\Tests\Serializer;

use JMS\Serializer\Context;
use JMS\Serializer\Exception\RuntimeException;
use JMS\Serializer\EventDispatcher\Event;
use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
use JMS\Serializer\GraphNavigator;
use JMS\Serializer\VisitorInterface;
use JMS\Serializer\Tests\Fixtures\Author;
use JMS\Serializer\Tests\Fixtures\AuthorList;

class NestedSerializationTest extends BaseSerializationTest
{
    /**
     * @group nestedss
     */
    public function testSerializerImbrication()
    {
        $this->dispatcher->addSubscriber(new CallingSerializerSubscriber($this->serializer));

        $list = new AuthorList();
        $list->add(new Author('foo'));
        $list->add(new Author('bar'));

        $this->assertEquals('[{"never_get_there"}}]', $this->serialize($list));
    }

    protected function getContent($key)
    {
        return [];
    }

    protected function getFormat()
    {
        return 'json';
    }
}

class CallingSerializerSubscriber implements EventSubscriberInterface
{
    protected $serializer;

    public function __construct($serializer)
    {
        $this->serializer = $serializer;
    }

    public function onPostSerialize(Event $event)
    {
        // could be to add data, or log, send message etc...
        $event->getVisitor()->addData('_extraJson', $this->serializer->serialize(array('foo' => 'bar'), 'json'));
    }

    public static function getSubscribedEvents()
    {
        return array(
            array('event' => 'serializer.post_serialize', 'method' => 'onPostSerialize', 'format' => 'json', 'class' => 'JMS\Serializer\Tests\Fixtures\Author'),
        );
    }
}
```

Thx

dbojdo:
[...]

PR #341, opened 2014-09-18
Allow multiple nested serializer calls

xavierlacot:
In case of serializer recursion, the serializer send an exception: "Can't pop from an empty datastructure", as noted by @damienalexandre in #319 

This PR allows to stack the navigator association to visitors, thus making it possible to nest serializer calls. It contains two commits:
- the first one contains the required changes for the bug to be fixed;
- the second one adds a test which shows up the bug resolution (tested against the fixed testsuite proposed in #292)


ogizanagi:
Any news about this PR ?

schmittjoh:
I'm not sure whether we need to support this use-case, i.e. nested calls. I think it might be better to create another serializer instance if you need this instead of using the same instance.

ogizanagi:
Thanks for the quick answer :smiley: 

This could be out of the scope of the current debate, but may I ask you the proper way to do that, as I need to register my custom handlers too, in the context of a Symfony 2 application ?
This basically means redoing the same work as in the `jms/serializer-bundle`, right ?

However, for my use case, using the `SerializerBuilder` directly, and registering the handlers I need seems sufficient.

damienalexandre:
I find it disturbing that we need a fresh new `Serializer` to perform nested calls. A new `SerializationContext` should be enough. I use `Serializer` as a service most of the time, where one instance is supposed to fit all serialization needs.

Isn't it like saying `Serializer` is not stateless?

digitalkaoz:
yeah ran into the same issue here...sadly the serializer is somewhat stateless and thats a bad design here in my eyes

digitalkaoz:
@schmittjoh why are you blocking this valid use case?

digitalkaoz:
if you have a more complex serializer construction (custom handlers, event subscribers, object constructor etcpp) this is a pain in the ass.

bassrock:
:+1:

bassrock:
@schmittjoh this helps especially when using this with Symfony so that you can reuse the injected serializer in different parts of the same code run.

bassrock:
I created a new branch that mimics this but on the latest code base with a few additions. #453

sfavot:
:+1:

goetas:
Do not know if this is still an issue,  but https://github.com/schmittjoh/serializer/pull/453#issuecomment-237467700 proposes an alternative implementation for this feature

goetas:
Nested serialization calls currently require big changes, for a relatively small benefit. Probably getting a new serializer instance can partially fix the issue.

Not planning to accept the feature for now.

xavierlacot:
So this makes the serializer stateful and unusable as a service, with all the performance and memory consumption downsides.

This is a blocker for "real life" production usage, as explained in the previous comments. This has proven to be blocking in at least 3 or 4 of our client projects, and should be clearly stated in the documentation: "jms-serializer is stateful and doesn't care of performance issues".

patie:
👍

PR #356, opened 2014-10-22
add prefix for default namespace

TPETb:
if you meet an element with default namespace, you still need to register a prefix for it to use xpath.

the reason behind it is that most of today environments use libs based on xpath1.0 which doesn't have "default namespace" entity in its specification while xpath2.0 does.

i'm proposing to use this options as it will both work for xpath1.0 and 2.0 while search without prefix will only work in xpath2.0


goetas:
this has been fixed with `1.2.0`

PR #358, opened 2014-10-29
fixed wrong result generation in visitArray if values are skipped

tecbot:
If the navigator return ```null``` and it should be skipped then the ```rs``` array loses his sequential indexes so you will get an object instead of an array in JSON.


0x46616c6b:
@schmittjoh ping :-)

0x46616c6b:
@schmittjoh ping :-/

goetas:

PR #362, opened 2014-11-07
Fix createFromFormat false with format c

frozenstupidity:
Due to an error in php which returns false when using createFromFormat with format c probably related to the issue https://bugs.php.net/bug.php?id=51950 when the format is c the string is passed into the constructor.

### DIFF
--- a/src/JMS/Serializer/Handler/DateHandler.php
+++ b/src/JMS/Serializer/Handler/DateHandler.php
@@ -106,7 +106,12 @@ class DateHandler implements SubscribingHandlerInterface
     {
         $timezone = isset($type['params'][1]) ? new \DateTimeZone($type['params'][1]) : $this->defaultTimezone;
         $format = $this->getFormat($type);
-        $datetime = \DateTime::createFromFormat($format, (string) $data, $timezone);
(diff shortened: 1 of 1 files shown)


theantichris:
@goetas @frozenstupidity Is the only thing keeping this from being merged the missing tests? This has been open for 2 years and is a pretty big bug not being able to use ISO format.

goetas:
Hi, currently I'm a bit busy, but if you noticed, in the last 3 months the pull request count went from 210 to 41... this will be solved soon too :)

I understand it's a pretty old issue, will arrive to it,  Give me just some time.

theantichris:
@goetas No worries. Just wanted to check in and see if these would be included sometime.

goetas:
The bug on php.net was closed as "not a bug".

Relevant part is: 
> DateTime::ISO8601 is defined to "Y-m-d\TH:i:sO"[1], so obviously
it doesn't accept microsecond fractions. If you expect these to be
given, use another format specifier, see <https://3v4l.org/ADn0l>.

So, I recommend to use the explicit ISO format instead of the 'c' shortcut

PR #368, opened 2014-11-29
Throw exception with clear message in case of int and bool as type

entering:
I'm always using int and bool instead of integer and boolean on types and then get a very generic exception message, this makes easier to find the problem.


goetas:
I'm sorry that the feedback loop for this was so long.

There is not a real reason to add this into the core. It is just a different type name that can be handled as class name... even if weird. Maybe we can make an alias of int and bool.

Closing PR

Issue #376, opened 2014-12-26
Serializer working with parent class


kessiler:
I have an application that uses doctrine 2 (ORM), and use an override in datetimes. (Doctrine\DBAL\Types\Type::overrideType)
In this situation the handlers not serialize the object type, even though the son of datetime.

See the example:

Class A -> ORM\DateTimeType extends \DateTime
Object A -> instance of DateTimeType

To serialize, the return should be a string of type date
Example:  serialize($A); -----> '2014-12-26 16:46:10'

But is an empty object..( {} )

goetas:
this is a more general problem and sub-problem of https://github.com/schmittjoh/serializer/pull/698

PR #377, opened 2014-12-29
Fixed handles working with parent class

kessiler:
Resolved issue #376


goetas:
Fixes #376

goetas:
see https://github.com/schmittjoh/serializer/issues/376

PR #378, opened 2014-12-31
return integer instead of string when the column is datetime and the format is timestamp

scourgen:
It's useful for api development.


goetas:

PR #385, opened 2015-01-20
Fix DoctrineObjectConstructor returning nulls instead of objects

cdonadeo:
Makes DoctrineObjectConstructor use the fallback constructor to create
a new object if the identifier computed from the metadata is not
currently stored in the database.  This previously just returned a null
value.


goetas:
this should be a configurable behaviour, not a default. 
Implemented as it is now is also a BC break

goetas:
fixed with https://github.com/schmittjoh/serializer/pull/751

PR #390, opened 2015-02-05
Generic accessor for psuedo template methods

brettminnie:
@schmittjoh This is a feature that we require in the OPG project, basically it fills a gap between the handler and an accessor, meaning I can re-use an accessor on multiple properties as and where they are needed

I have rebased it against master as you requested.

Here is an example of the accessor being invoked in an annotation

https://github.com/ministryofjustice/opg-core-public-domain-model/blob/master/src/Opg/Core/Model/Entity/Document/Document.php#L91

And here is an example of a generic accessor method
https://github.com/ministryofjustice/opg-core-public-domain-model/blob/master/src/Opg/Common/Model/Entity/Traits/DateTimeAccessor.php

I got tired of rewriting the same code in getters and setters :)


brettminnie:
@schmittjoh sorry I need this to go in before I open the PR on the other branch as this functionality is in there as well

goetas:
33 Commits?

brettminnie:
Yeah I know, busy cleaning both these pr's up at the moment, will resubmit once done

PR #391, opened 2015-02-05
Custom depth on circular reference

brettminnie:
@schmittjoh 

Master rebased in, however functionality in https://github.com/schmittjoh/serializer/pull/390 is also in here, so if we can do both, can you do 390, and then let me rebase this against it again

Thanks


brettminnie:
Making this sane

Issue #367, opened 2014-11-24
Exclude annotation not preventing attempt to find public methods when using AccessType


andrewboerema:
I have discovered an issue where, if you set AccessType("public_method") on a class, Exclude annotations on that class's properties are ignored. Ex:
```
<?php
namespace Application\Fruits;

use JMS\Serializer\Annotation as JMS;

/**
 * Class Bananas
 *
 * @package Application\Fruits
 * @JMS\AccessType("public_method")
 */
class Bananas {
    /**
     * @var array
     * @JMS\Exclude
     */
    private static $acceptedStatuses = array(
        'Approved',
        'Pending',
    );
    ...
}
```

In this case, I get the error "There is neither a public getAcceptedStatuses method, nor a public isAcceptedStatuses method, nor a public hasAcceptedStatuses method in class Application\Model\RedOakPage. Please specify which public method should be used for retrieving the value of the property acceptedStatuses."

It would be my understanding that the more specific Exclude annotation on the property would override the class-level annotation of AccessType. If that isn't true, this isn't a defect, but it should be made clear on the docs for AccessType.

Strate:
+1

Nemo64:
still no fix?

mattjanssen:
PR #397

detook:
+1

goetas:
PR has been merged

PR #397, opened 2015-02-11
#367 Exclude annotation not preventing attempt to find public methods when using AccessType

Strate:
Attempt to find getters and setters only if property included to serialization.

### DIFF
--- a/src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
+++ b/src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
@@ -211,10 +211,10 @@ class AnnotationDriver implements DriverInterface
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


mattjanssen:
:+1: I was just about to create this exact PR.

When using `@AccessType("public_method")` the call to `setAccessor()` fails for properties that don't have a setter/getter. This is regardless of `@Exclude` or `@ExclusionPolicy("ALL")` annotations.

The exception thrown is:
```
There is no public %s method in class %s. Please specify which public method should be used for setting the value of the property %s.
```

mattjanssen:
Issue #367

sagikazarmark:
What's blocking this from being merged? IMO this is a serious issue needs to be addressed ASAP.

goetas:
looks good to me

goetas:
checked again, looks good. Thank you

PR #401, opened 2015-02-19
better json serialized of a formtype

vguardiola:
When I serialize a form to json if the root elements are bad named, ever say "children" and not the getName() value configured in the FormType difinition and when serialized a form with data this data don't serialied

**Before**:
empty  and filled form:

``` Json
{
  "children": {
    "name": [],
    "description": [],
    "typeId": [],
    "active": []
  }
}
```

**After**:
empty form:

``` Json
{
  "card": {
    "name": null,
    "description": null,
    "typeId": null,
    "active": "1"
  }
}
```

filled form:

``` Json
{
  "card": {
    "name": "the name",
    "description": "the desc",
    "typeId": 1,
    "active": "1"
  }
}
```

### DIFF
--- a/src/JMS/Serializer/Handler/FormErrorHandler.php
+++ b/src/JMS/Serializer/Handler/FormErrorHandler.php
@@ -146,13 +146,13 @@ class FormErrorHandler implements SubscribingHandlerInterface
         }
 
         if ($children) {
-            $form['children'] = $children;
+            $form[$data->getName() ?: 'children'] = $children;
         }
 
         if ($isRoot) {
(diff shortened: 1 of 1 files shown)


schmittjoh:
This is tricky because people might depend on the current behavior.

vguardiola:
That's true , but I think this is the right functionality for a Rest API , especially for editing request.

I do not know if there is another possibility that it is a configurable behavior. Certainly I have not researched more.

goetas:
Tried that road in 2.0 but decided to rever as there were edgecases where the solution was not optimal

Issue #337, opened 2014-09-16
serializer.pre_serialize event only thrown on objects/classes


ReservedDeveloper:
The events system is a great piece of functionality for being able to dynamically modify the serialization process. Unfortunately, as currently implemented, there is no way to make modifications to primitives prior to their serialization. 

At a *very* casual glance, it could be as simple as moving the [responsible block](https://github.com/schmittjoh/serializer/blob/0.15.0/src/JMS/Serializer/GraphNavigator.php#L152) (starts ```// Trigger pre-serialization callbacks, and listeners if they exist.```) to a higher line, allowing us to hook in for all types.

Would their be any hindrances/complications in doing so? (assuming all RuntimeExceptions have been checked)

TiS:
+1. I want to be able to sometimes change the type of to-be-serialized value.

For example: my application uses big integers as entity IDs. The problem is, the Javascript does not handle such big integers well. I want to be able to detect, if the integer is bigger than some value and then change the serialized type to string (so it is enclosed in quotes). Currently I have to create an object with the integer value and register a handler for it. But this is not possible, the code bas is just too big, and making such a change in a model just because of serialization problems is overkill.

goetas:
As of 1.2.0 is possible to register handlers on primitives. See https://github.com/schmittjoh/serializer/pull/222#issuecomment-234871449 to see how to create handlers for primitives

Issue #162, opened 2013-10-06
PreSerialization Event not handled if the value is not object


yoshi49535:
If the type of property is not object, "pre_serialization" event is not dispatched.
In other words, the method "SerializationContext::isVisiting" throw an exception before handle event listener, iff the value is not an object. 

This is too limited i think.

In my case, I created Hash Type, which can be either "String hash" or "Hashable instance".
I want to serialize without replacement of the object.

goetas:
see https://github.com/schmittjoh/serializer/pull/610

PR #408, opened 2015-03-02
Dispatch events before looping through data

piwi91:
Events were only dispatched if the data was an object but events should also be dispatched if other types are serialized.
This PR solves issue #337 and #162.

### DIFF
--- a/src/JMS/Serializer/GraphNavigator.php
+++ b/src/JMS/Serializer/GraphNavigator.php
@@ -108,6 +108,21 @@ final class GraphNavigator
             $type = array('name' => 'NULL', 'params' => array());
         }
 
+        // Trigger pre-serialization callbacks, and listeners if they exist.
+        // Dispatch pre-serialization event before handling data to have ability change type in listener
+        if ($context instanceof SerializationContext) {
+            if (null !== $this->dispatcher && $this->dispatcher->hasListeners('serializer.pre_serialize', $type['name'], $context->getFormat())) {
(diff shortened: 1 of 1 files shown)


src/JMS/Serializer/GraphNavigator.php
```diff
@@ -108,6 +108,21 @@ public function accept($data, array $type = null, Context $context)
...
+                $this->dispatcher->dispatch('serializer.pre_serialize', $type['name'], $context->getFormat(), $event = new PreSerializeEvent($context, $data, $type));
+                $type = $event->getType();
+            }
+        } elseif ($context instanceof DeserializationContext) {
+            if (null !== $this->dispatcher && $this->dispatcher->hasListeners('serializer.pre_deserialize', $type['name'], $context->getFormat())) {
+                $this->dispatcher->dispatch('serializer.pre_deserialize', $type['name'], $context->getFormat(), $event = new PreDeserializeEvent($context, $data, $type));
+                $type = $event->getType();
+                $data = $event->getData();
+            }
+        }
```
stof: your change is wrong. You are now dispatching the event before the level handling in the context, which does not work properly anymore

stof:
you should not make source code executable

piwi91:
Ok, to bad that this issue isn't fixed yet. I don't know exactly how to fix this the "right way" so I implemented another solution in my application so I don't need this PR to be approved.
Thanks for mentioning about the executable code. Our samba configuration had an error in it which chmodded all the files to 777 :(

PR #419, opened 2015-03-25
Update AbstractDoctrineTypeDriver.php

afurculita:


goetas:

Closing the PR, do not see a real benefit in this and there is no explanation in the PR

PR #423, opened 2015-04-02
MaxDepth is handled incorrectly with Collections

progrupa:
This fixes handling collections for the purpose of depth calculations. Since arrays are not counted as a depth level, neither should Collections.
The other side effect of this change is that @MaxDepth on an entity inside a collection will be handled correctly as well (before, each Collection contributed +1 to depth, but did not leave PropertyMetadata, therefore was not subtracted when calculating relative depth).


MartinNuc:
+1

vlcekmi3:
+1

kofon:
+1

progrupa:
Could we maybe merge this in?

goetas:

PR #426, opened 2015-04-09
Recursion groups

KmeCnin:


smilesrg:
+1 for the merge!

YoruNoHikage:
I think this pretty much a duplicate of #170, it doesn't work the same way but you get the same result.

goetas:

> I think this pretty much a duplicate of #170, it doesn't work the same way but you get the same result.

Issue #427, opened 2015-04-14
The document type "..." is not allowed


crazycodr:
I've just forked the serializer to find out that the tests are not passing. There seems to be a failure and an error which are both related to the inner entity in the doctype.

Tests failing are:
- JMS\Serializer\Tests\Serializer\XmlSerializationTest::testWhitelistedDocumentTypesAreAllowed
- JMS\Serializer\Tests\Serializer\XmlSerializationTest::testExternalEntitiesAreDisabledByDefault

After an hour or so studying the code, i came to conclusion that the code is fine and the tests are the problem. In both tests, we expect the doctype to be:

    <!DOCTYPE author [<!ENTITY foo SYSTEM "php://filter/read=convert.base64-encode/resource='.basename(__FILE__).'">]>

While, the doctype resolves to the internalSubset instead of the whole DocType. So the value that becomes expected should be:

    <!ENTITY foo SYSTEM "php://filter/read=convert.base64-encode/resource='.basename(__FILE__).'">

Now, i'm not a super fan of XML but i'm pretty sure the code actually works. Can anyone confirm that the tests are really incorrect? If so, i'll PR with the fix asap.

Thanks

PR #428, opened 2015-04-14
Fix to invalid tests for XML entity subsets

crazycodr:
Fix https://github.com/schmittjoh/serializer/issues/427

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 vendor/
 phpunit.xml
-
+.idea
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,3 +1,3 @@
 vendor/
 phpunit.xml
-
+.idea
```
piotrantosik: adding ide path is no good choice - better use global gitignore

PR #429, opened 2015-04-16
Add handler for stdClass

Spea:
As mentioned in #51 and #211 I added a new handler to treat stdClass correctly.

### DIFF
--- a/tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php
+++ b/tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php
@@ -22,6 +22,7 @@ use JMS\Serializer\Context;
 use JMS\Serializer\DeserializationContext;
 use JMS\Serializer\GraphNavigator;
 use JMS\Serializer\Handler\PhpCollectionHandler;
+use JMS\Serializer\Handler\StdClassHandler;
 use JMS\Serializer\SerializationContext;
 use JMS\Serializer\Tests\Fixtures\DateTimeArraysObject;
 use JMS\Serializer\Tests\Fixtures\Discriminator\Car;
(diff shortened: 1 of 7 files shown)


Spea:
ping :)

Spea:
@schmittjoh do you need something else to get this PR merged? would be really nice

alexdesignworks:
@schmittjoh
Can we please have this merged. There are like 10 of issues open regarding this.

hshn:
+1

hidenorigoto:
:+1:

qckanemoto:
+1

mathroc:
:+1:

kcivey:
:+1:

tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php
```diff
@@ -29,9 +30,11 @@
 use JMS\Serializer\Tests\Fixtures\Garage;
 use JMS\Serializer\Tests\Fixtures\InlineChildEmpty;
 use JMS\Serializer\Tests\Fixtures\NamedDateTimeArraysObject;
+use JMS\Serializer\Tests\Fixtures\ObjectWithStdClass;
 use JMS\Serializer\Tests\Fixtures\Tree;
 use JMS\Serializer\Tests\Fixtures\VehicleInterfaceGarage;
 use PhpCollection\Sequence;
+use stdClass;
```
stof: This bundle does not import classes of the global namespace currently

goetas:

Closing. This is a feature that can be integrated with a simple class handler (exactly how is done here) using user-land code

mathroc:
@goetas it can indeed be integrated in user-land code but it's odd not to have it in the core bundle. I don't get why `stdClass` does not deserves to be supported directly, it's one of the core structure of PHP, almost on the same level as `array`
it could obviously be made as an external package but it would still be a PITA to require this package and then register the handler. 

Do you mind explaining why you think it should not be supported ?

thank you

goetas:
It is an specific use case and i prefer to have in the core as less things as possible.

The serialization/deserialization of `stdClass` is a special case that some people can have, but IMO is a bad practice.

The model should be typed, express behaviors, expose logic (not implementation).... and `stdClas` is a way to expose inner implementation.
For the same reason many libraries are getting rid of "configuration arrays"   preferring classes. 

So i have two suggestions here: 
- get rid of `stdClass` in your code
- implement this class handler in a user package

Issue #430, opened 2015-04-17
Erroneous data format for unserializing


tmilos:
Issue that will fix schmittjoh/JMSSerializerBundle#394

tmilos:
Merged by  #431

PR #431, opened 2015-04-17
Erroneous data format for unserializing #430

tmilos:


schmittjoh:
Can you add that to the composer.json, too?

tmilos:
Sure, but what about composer.lock? Should I just leave it? 
I didn't add it now, since it's already present in dependencies, so I didn't touch composer.json

PR #434, opened 2015-04-21
softened prefix check effectively forcing prefix for default namespace

maZahaca:


schmittjoh:
Could you add a test for what this is doing?

discordier:
I guess he wants the default namespace prefix (which is the first one in `$registeredNamespaces`) enforced on ALL elements which is plain wrong IMO as by doing so, he is removing the possibility to set an `xsi:schemaLocation` or `xsi:noNamespaceSchemaLocation` from an listener which is needed in some cases. Every element will get prefixed with the just created prefix.
This change also seems to create a new namespace prefix alias for the default namespace on every element which is definately not a wise idea.

maZahaca:
I cannot add tests, because that commit was added by another developer, but anyway that fix works for me

goetas:

PR #20, opened 2013-01-10
Added support for an "array" format

tystr:
This PR adds support for an "array" format, providing the functionality to serialize to and deserialize from arrays.

PR #435, opened 2015-04-22
Add toArray and fromArray methods to the serializer

tystr:
This PR implements the solution suggested by @schmittjoh to add additional methods for dealing with array input/output instead of adding a new format for arrays.

https://github.com/schmittjoh/serializer/pull/20#issuecomment-94781669

TODO
- [x] Add some test cases

### DIFF
--- a/src/JMS/Serializer/Serializer.php
+++ b/src/JMS/Serializer/Serializer.php
@@ -121,6 +121,49 @@ class Serializer implements SerializerInterface
         return $visitorResult;
     }
 
+    public function toArray($data, SerializationContext $context = null)
+    {
+        if (null === $context) {
+            $context = new SerializationContext();
+        }
+
+        $context->initialize(
+            'json',
+            $visitor = $this->serializationVisitors->get('json')->get(),
+            $this->navigator,
+            $this->factory
+        );
+
+        $visitor->setNavigator($this->navigator);
+        $this->navigator->accept($visitor->prepare($data), null, $context);
+
+        return (array) $visitor->getRoot();
+    }
+
+    public function fromArray($data, $type, DeserializationContext $context = null)
+    {
+        if (null === $context) {
+            $context = new DeserializationContext();
+        }
+
+        $context->initialize(
+            'json',
+            $visitor = $this->deserializationVisitors->get('json')->get(),
+            $this->navigator,
+            $this->factory
+        );
+
+        $visitor->setNavigator($this->navigator);
+        $navigatorResult = $this->navigator->accept($data, $this->typeParser->parse($type), $context);
+
+        // This is a special case if the root is handled by a callback on the object iself.
+        if ((null === $visitorResult = $visitor->getResult()) && null !== $navigatorResult) {
+            return $navigatorResult;
+        }
+
+        return $visitorResult;
+    }
+
     /**
      * @return MetadataFactoryInterface
      */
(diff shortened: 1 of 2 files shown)


tests/JMS/Serializer/Tests/Serializer/ArrayTest.php
```diff
@@ -0,0 +1,1307 @@
+<?php
```
schmittjoh: Let's not replicate the entire test suite. One simple test per method should be ok.

src/JMS/Serializer/Serializer.php
```diff
@@ -121,6 +121,49 @@ public function deserialize($data, $type, $format, DeserializationContext $conte
...
+            'json',
+            $visitor = $this->serializationVisitors->get('json')->get(),
+            $this->navigator,
+            $this->factory
+        );
+
+        $visitor->setNavigator($this->navigator);
+        $this->navigator->accept($visitor->prepare($data), null, $context);
+
+        return $visitor->getRoot();
```
schmittjoh: We need some special handling for the ``ArrayObject`` here that the JSON visitor returns sometimes. Could you add a test for that, too?

tystr:
I've just returned from vacation. I'll wrap this up this week.

davidwindell:
:+1:

tystr:
@schmittjoh I've simplified the tests for the array methods down to a couple of very rudimentary tests and added handling for `ArrayObject`.

src/JMS/Serializer/Serializer.php
```diff
@@ -121,6 +121,49 @@ public function deserialize($data, $type, $format, DeserializationContext $conte
...
+            'json',
+            $visitor = $this->serializationVisitors->get('json')->get(),
+            $this->navigator,
+            $this->factory
+        );
+
+        $visitor->setNavigator($this->navigator);
+        $this->navigator->accept($visitor->prepare($data), null, $context);
+
+        return (array) $visitor->getRoot();
```
schmittjoh: We should do this for the entire array (including children) as these might have an array object, too.

tests/JMS/Serializer/Tests/Serializer/ArrayTest.php
```diff
@@ -0,0 +1,87 @@
...
+        $this->serializer = new Serializer(
+            new MetadataFactory(new AnnotationDriver(new AnnotationReader())),
+            new HandlerRegistry(),
+            new UnserializeObjectConstructor(),
+            new Map(array('json' => new JsonSerializationVisitor($namingStrategy))),
+            new Map(array('json' => new JsonDeserializationVisitor($namingStrategy)))
+        );
+    }
+
+    public function testToArray()
```
schmittjoh: Could you add a test how we behave if for example a string or another scalar is passed? We should probably throw an exception if the result is not an array.

rdohms:
:+1:

rdohms:
Where are we with this? Any ideas when a release with it will come out?

schmittjoh:
I've made a few tweaks, and merged this. We'll need to give this a bit of time in master before thinking about a release.

davidwindell:
Great, thanks @tystr and @schmittjoh for getting this done!

davidwindell:
Just a quick update, this has been used by our app in production since 1 May and all good so far!

rdohms:
can we get this in a release?

rdohms:
Aha, you did, v1.0.0 now supports this. :+1:

PR #439, opened 2015-04-29
evented object construction

digitalkaoz:
This PR adds the ability to hook into the ObjectConstruction during Deserialization via Events.

simple create a Subscriber like this:

```php
class MyClassConstructionSubscriber implements EventSubscriberInterface
{
    /**
     * @inheritdoc
     */
    public static function getSubscribedEvents()
    {
        return [
            ['event' => Events::OBJECT_CONSTRUCTION, 'class' => MyClass::class, 'method' => 'createMyClass']           
        ];
    }

    public function createMyClass(ObjectConstructionEvent $event)
    {
        $event->setObject(new MyClass());
    }
}
```

and register the normal way like the docs say: http://jmsyst.com/libs/serializer/master/event_system


schmittjoh:
How is that different/superior to an ObjectConstructor?

digitalkaoz:
i can split up my object construction for different classes into a stable and flexible event system. In my current project we have a bunch of different object constructors (i chained them), an evented system makes more sense in my eyes.

schmittjoh:
The way how this is implement does not necessarily require it to be in the main library. 

I would suggest to keep this in a separate repository for the moment as I'm not sure how many people really need this kind of flexibility, and it does have a performance impact/adds more extension points/complexity.

digitalkaoz:
keeping it in a seperate library wont work as i dont get the EventManager (without hacky reflections) into the EventedObjectConstructor.

i think the performance impact is not really existent, as if no event is subscribed to object_construction the event_dispatcher "fails fast".

the complexity is not really an issue in my eyes, the normal user wont recognize the feature is there.

would love to see it in the main library :) cause it would enable me to delete my version of it...

schmittjoh:
Right now, I'm not ready to add this to the library, but I'm happy to add a way to retrieve the event dispatcher or make that injectable if that helps.

digitalkaoz:
retrieving the eventdispatcher e.g. from the context would help :)

digitalkaoz:
in the end im interested what your concerns are? i thought it would be useful, and its a minimal inversive

goetas:
A nicer idea: https://github.com/schmittjoh/serializer/issues/777 IMO

PR #444, opened 2015-05-07
permit access on variables for inherited implementations

ujhgj:


goetas:
closing, probably already outdated and anyway prefer to keep the variable private

PR #447, opened 2015-05-10
Deserialize exception

ivan1986:
If deserialize from json/xml with wrong structure exception hasn't info about path where structure is wrong - it's very hard to debug

this add new exception, contains path as "comments[0].author.full_name" to unexpected value


ivan1986:
also #110 is fixed

schmittjoh:
I like the idea of adding a path where we throw hard errors, but this seems to alter behavior at the same time. Data that was previously acceptable will now generate exceptions; this is not something that I want to do.

I would like to scale this down a bit, and only improve the errors that we were already previously throwing.

ivan1986:
Hmm, DeserializeException extents RuntimeException - all old cache works
all checks is not strict - if send number string to integer all works, 
prev behavion convertion array to scalar generate php error and it's fixing
current change convert non numeric string to integer/float - before return 0, now throw exception, but i think is correct behavion

ivan1986:
@schmittjoh ping
what also need for merge?
current php errors and RuntimeException replaced to DeserializeException
I found only incorrect numeric data

ivan1986:
@schmittjoh ping

goetas:

VasekPurchart:
@goetas I see you are writing this to more PRs, can you please add to your text mention, that it would be great if the author (or anyone) who opens a new PR, mentions it also in the old one, so subscribers get the notification? Thanks :)

goetas:
Whoever was involved in the discussion should receive the notification from github, so no need for explicit mentioning.

ivan1986:
@goetas yes, i still interested this PR

goetas:
The problem that I see with this feature is a lot of code in the core, only for a debug feature.

Any alternative approach?

ivan1986:
i think it's not only for debug
it's for correct and full error responce - this bundle use for get data from api
and not lot of code - exception, index metadata and formalize errors

PR #453, opened 2015-05-21
Rebase of serialization nesting onto the latest

bassrock:
This is a rebase of #341 on the latest code base. It also adds some accessor methods to remove properties via a callback after it has been serialized.

### DIFF
--- a/src/JMS/Serializer/GenericSerializationVisitor.php
+++ b/src/JMS/Serializer/GenericSerializationVisitor.php
@@ -24,13 +24,27 @@ use JMS\Serializer\Metadata\PropertyMetadata;
 
 abstract class GenericSerializationVisitor extends AbstractVisitor
 {
+    private $navigatorStack;
     private $navigator;
+    private $rootStack;
     private $root;
+    private $dataStackStack;
     private $dataStack;
     private $data;
 
+    public function __construct($namingStrategy)
+    {
+        parent::__construct($namingStrategy);
+        $this->navigatorStack = new \SplStack;
+        $this->dataStackStack = new \SplStack;
+        $this->rootStack = new \SplStack;
+    }
+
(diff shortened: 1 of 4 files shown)


digitalkaoz:
+1

crisu83:
+1 as we use the same instance for serialization across our Lumen application through DI.

bassrock:
@schmittjoh Any chance of merging this in? Just rebased with the latest from master again.

bassrock:
@goetas Since you look to be helping now. Any chance of merging this in?

goetas:
I see some problems with this PR that makes it "not reday to be merged". First, you have removed tests. Second,  this approach complicates all the visitors  adding code to handle this specific usecase. 

Have you considered an approach of recreating only the visitor instance for each seralization call? It should give that same result without changing visitors code...

bassrock:
@goetas Im not sure what you mean when you say I removed tests. I haven't removed any tests. 

How would you recreate the visitor instance for each serialization call?

goetas:
>  Im not sure what you mean when you say I removed tests. I haven't removed any tests.

as you can see on https://github.com/schmittjoh/serializer/pull/453/files there are no tests

> How would you recreate the visitor instance for each serialization call?

https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Serializer.php#L68 contains the list of visitors.

Currently visitors are pre-instantiated in the serialization builder. Doing some kind of lazy-instantiation/callback  will allow to instantiate the visitor right when needed.

It is a bit more complex feature, but is a general solution to the problem exposed here

goetas:
Nested serialization calls currently require big changes, for a relatively small benefit. Probably getting a new serializer instance can partially fix the issue.

Not planning to accept the feature for now.

PR #462, opened 2015-06-11
fix json output (from [] to {} if empty) of form error

jhkchan:
For form errors, right now there are two kinds of json outputs:

```php
"key": []
```

or

```php
"key": {
  "errors": [
    ...
  ]
}
```

It depends on whether the form has errors. This would introduce inconsistency in data format for json and could be troublesome for client side to handle (such as on iOS app). This fix would make the output consistent as follows:

```php
"key": {}
```


schmittjoh:
Can you add a test?

jhkchan:
I will add.

jhkchan:
I am done. Please check.

PR #466, opened 2015-06-16
Added public `hasData` function to check if a data key already have been added.

mowcixo:
This pull request is related with schmittjoh/serializer#197 but it's less aggressive.

In order to reduce try/catch blocks I think it's important to expose data in some way, at least to know if a key already exists.

This will reduce many try/catch blocks or "added flags".

### DIFF
--- a/src/JMS/Serializer/GenericSerializationVisitor.php
+++ b/src/JMS/Serializer/GenericSerializationVisitor.php
@@ -162,7 +162,7 @@ abstract class GenericSerializationVisitor extends AbstractVisitor
      */
     public function addData($key, $value)
     {
-        if ($this->hasData($key)) {
+        if (isset($this->data[$key])) {
             throw new InvalidArgumentException(sprintf('There is already data for "%s".', $key));
         }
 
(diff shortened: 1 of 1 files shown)


src/JMS/Serializer/GenericSerializationVisitor.php
```diff
@@ -162,12 +162,23 @@ public function visitProperty(PropertyMetadata $metadata, $data, Context $contex
      */
     public function addData($key, $value)
     {
-        if (isset($this->data[$key])) {
+        if ( $this->hasData($key) ) {
```
stof: the extra spaces should be removed

stof: and I think the isset check could still be done directly here rather than adding an extra method call

mowcixo: right.

mowcixo: Yes, thinking on a multiple item serialize, the method accessor is a bad idea.

goetas:
replaced by https://github.com/schmittjoh/serializer/pull/625 that integrates tests

mowcixo:
Perfect!

PR #470, opened 2015-06-23
[Feature] - Issue 424 - adding ExcludeForGroups 

leberknecht:
Adding functionality to exclude a property if a specifiy exclusion-group is set but expost it per default.
```php
    use JMS\Serializer\Annotation\ExcludeForGroups;

    class BlogPost
    {
        private $id;

        private $title;

        /** @ExcludeForGroups({"list"}) */
        private $comments;

        private $createdAt;
    }
```
If you now set the `list` exclusion-group on serialization, the `comments` property will be excluded::

    $serializer->serialize(new BlogPost(), 'json', SerializationContext::create()->setExclusionGroups(array('list')));
    //will output $id, $title and $createdAt.

This can be handy if you just want to reduce your response without setting explicit groups for everything you want to expose.


lotas:
:+1:

smilesrg:
+1 for this feature

afurculita:
@leberknecht how about replacing `@ExcludeForGroups({"list"})` with something like `@Exclude(groups={"list"})`? And adding a new type of ExclusionPolicy: maybe "GROUP"?

leberknecht:
I like the idea! Atm the @Exclude is not handled as the other exclusion-policies, so maybe we should change that. Not sure if this will break BC....anyways => i think giving @exclude a group parameter solves my problem, i'll create a new PR for that :)

PR #482, opened 2015-08-19
Switch the Twig integration to use non-deprecated APIs

stof:
This makes the library compatible with Twig 2.0, and avoids deprecation warnings in 1.21+

Note that I changed the way the ``serialization_context`` function is compiled. there is no way to wrap this static function into a method of the extension. This will save a method call each time this function is used.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,12 +19,15 @@
         "doctrine/annotations": "1.*",
         "doctrine/instantiator": "~1.0.3"
     },
+    "conflict": {
+        "twig/twig": "<1.12"
+    },
     "suggest": {
         "symfony/yaml": "Required if you'd like to serialize data to YAML format."
     },
     "minimum-stability": "dev",
     "require-dev": {
-        "twig/twig": ">=1.8,<2.0-dev",
+        "twig/twig": "~1.12|~2.0",
         "doctrine/orm": "~2.1",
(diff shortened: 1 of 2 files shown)


Soullivaneuh:
Thanks. :+1: 

@schmittjoh can you merge it please?

Soullivaneuh:
@stof I think `twig/twig` requirement should be bumped to `1.12`.

A `conflict` key would be better too.

stof:
@Soullivaneuh updated

Nakard:
When can we expect this change will get tagged?

oste:
@stof @schmittjoh This change causes error b/c the string is not callable. The changes in #484 work for me.

INSEAD-asim:
Guys, is this already published? which version has this fix?

stof:
@INSEAD-asim only the dev version has it. It is not yet released

INSEAD-asim:
@stof , Thanks for update. Any roadmap when it will release?

stof:
no idea. I'm not the maintainer of this project. I've been asking for a release since 3 weeks

INSEAD-asim:
OK. no issues. I hope somebody else will reply.

lucascherifi:
@schmittjoh can you tag it please ?

yalg:
I agree for a tag please :)

PR #483, opened 2015-08-25
Fix deprecated Twig classes usage

Soullivaneuh:


Soullivaneuh:
Bad repository target, already existing PR. Closing.

PR #487, opened 2015-09-06
Fix issue where error is thrown for non callable

oste:
Running HHVM and seeing this below issue with the new changes for twig. This reverts some code to make it callable again.

Fatal Error: Argument 2 passed to Twig_Function::__construct() must be an instance of callable


stof:
See #490 for the proper fix (i.e. without the extra method calls for each usage of the function in templates)

PR #491, opened 2015-09-09
Fix deprecated ValidatorInterface usage

### LINKED ISSUE #438: Deprecated usage of ValidatorInterface
Concerned file: https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php#L30

Related notice:

``` php
 * @deprecated since version 2.5, to be removed in 3.0.
 *             Use {@link \Symfony\Component\Validator\ValidatorInterface} instead.
```

As this class is passed on the constructor, don't really know how to manage it.

Maybe create a legacy class? Where this class is used?

Regards

Soullivaneuh:
Fixes #438.

ping @stof for review.

@schmittjoh: In which configuration file this service is configured?

### DIFF
--- a/src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
+++ b/src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
@@ -21,14 +21,23 @@ namespace JMS\Serializer\EventDispatcher\Subscriber;
 use JMS\Serializer\EventDispatcher\Event;
 use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
 use JMS\Serializer\Exception\ValidationFailedException;
-use Symfony\Component\Validator\ValidatorInterface;
+use Symfony\Component\Validator\ValidatorInterface as LegacyValidatorInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 class SymfonyValidatorSubscriber implements EventSubscriberInterface
 {
     private $validator;
 
-    public function __construct(ValidatorInterface $validator)
+    public function __construct($validator)
     {
+        if (!$validator instanceof ValidatorInterface && !$validator instanceof LegacyValidatorInterface) {
+            throw new \InvalidArgumentException('Argument 1 of '.__METHOD__.' must be an instance of Symfony\Component\Validator\Validator\ValidatorInterface or Symfony\Component\Validator\ValidatorInterface.');
+        }
+
+        if (!$validator instanceof ValidatorInterface) {
+            @trigger_error('Passing Symfony\Component\Validator\ValidatorInterface instance to '.__METHOD__.' is deprecated since version 1.1 and will be removed on 2.0.');
+        }
+
         $this->validator = $validator;
     }
 
@@ -50,7 +59,11 @@ class SymfonyValidatorSubscriber implements EventSubscriberInterface
         $validator = $this->validator;
         $context->attributes->get('validation_groups')->map(
             function(array $groups) use ($event, $validator) {
-                $list = $validator->validate($event->getObject(), $groups);
+                if ($validator instanceof ValidatorInterface) {
+                    $list = $validator->validate($event->getObject(), null, $groups);
+                } else {
+                    $list = $validator->validate($event->getObject(), $groups);
+                }
 
                 if ($list->count() > 0) {
                     throw new ValidationFailedException($list);
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,8 @@
         "symfony/validator": "~2.0",
         "symfony/form": "~2.1",
         "symfony/filesystem": "2.*",
-        "phpunit/phpunit": "~4.0"
+        "phpunit/phpunit": "~4.0",
+        "symfony/phpunit-bridge": "^2.7.4"
     },
     "autoload": {
         "psr-0": {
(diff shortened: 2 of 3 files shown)


stof:
> In which configuration file this service is configured?

not needed. The service id does not change. there is nothing to configure

src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
```diff
@@ -21,14 +21,23 @@
...
     private $validator;
 
-    public function __construct(ValidatorInterface $validator)
+    public function __construct($validator)
     {
+        if (!$validator instanceof ValidatorInterface && !$validator instanceof LegacyValidatorInterface) {
+            throw new \InvalidArgumentException('Argument 1 of '.__METHOD__.' must be an instance of Symfony\Component\Validator\Validator\ValidatorInterface or Symfony\Component\Validator\ValidatorInterface.');
+        }
+
+        if ($validator instanceof LegacyValidatorInterface) {
```
stof: check for ``!$validator instanceof ValidatorInterface`` instead. Otherwise you will get the deprecation warning all the time (the class implements both interfaces in Symfony for BC reasons)

src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
```diff
@@ -50,7 +59,11 @@ public function onPostDeserialize(Event $event)
         $validator = $this->validator;
         $context->attributes->get('validation_groups')->map(
             function(array $groups) use ($event, $validator) {
-                $list = $validator->validate($event->getObject(), $groups);
+                if ($validator instanceof ValidatorInterface) {
+                    $validator->validate($event->getObject(), null, $groups);
```
stof: you are missing the assignment here

stof:
you should add tests for this class

Soullivaneuh:
> you should add tests for this class

I agree, but tests are already failing on master. BTW, Travis should be set up.

@schmittjoh What should I do for this?

stof:
@Soullivaneuh Travis is running. It is just unable to set the commit status. See #495

Soullivaneuh:
@stof yes, and as you can see, test are already failling: https://travis-ci.org/schmittjoh/serializer/builds/80216387

They need to be fixed before writing my own.

stof:
@Soullivaneuh this does not prevent writing new tests for these class, which would be passing (only 2 tests are failing in the testsuite, not the whole testsuite)

Soullivaneuh:
Ideed. Was thinking travis broke before ending all the test. Seems not.

Soullivaneuh:
@stof @schmittjoh done.

stof:
@Soullivaneuh this is not actually testing the BC layer: the only way to test it is to force installing an old Symfony version.
You should have tests using always the old API (marked as ``@group legacy``, and skipped when the deprecated API is not available, to make the testsuite ready for Symfony 3 support) and tests using always the new API (and skipped on old Symfony Validator versions when the interface is not available)

Soullivaneuh:
Okay @stof, it would be better to copy the class and make it legacy.

stof:
well, you could indeed have 2 test cases rather than having 2 series of tests in the same testcase, but it would make it harder to notice that we have 2 series of tests for someone editing tests of the class later

Soullivaneuh:
> 2 series of tests for someone editing tests of the class later

Note sure if we should update legacy tests in this case.

Done for double test cases.

stof:
> Note sure if we should update legacy tests in this case.

IMO we should: the tests are there to ensure that the feature works fine against both APIs, to ensure compatibility with Symfony 2.3 too

Soullivaneuh:
@stof Thinking of your words:

> this is not actually testing the BC layer: the only way to test it is to force installing an old Symfony version.

Actually, with a single class like I did before, both will be tested if Travis is correctly configured with different symfony versions AND lowest deps.

stof:
@Soullivaneuh but not locally, making it easier to miss it. and it would also be a pain to configure tests as legacy, as your test would be tested the legacy case only in some environments but not in others

Soullivaneuh:
> as your test would be tested the legacy case only in some environments but not in others

I don't think it's needed too. We are testing our usage case, not if legacy `ValidatorInterface` works well on any Symfony version. It's a job for Symfony mainteners, isn't it?

BTW, actually it's spitted onto two classes, should be ok now.

stof:
@Soullivaneuh we are testing whether the serializer code works againt both interfaces, which is the job of the serializer library.

Symfony maintainers cannot check whether the interface works btw: an interface never works as it has no implementation for its methods. You can only write tests for concrete code, not for interfaces. Tests are covering the implementation.

Soullivaneuh:
> an interface never works as it has no implementation for its methods.

A agree, but we are taking an interface as an argument, not writing one. Am I wrong?

stof:
@Soullivaneuh you are writing code against 2 different interfaces. So you should have tests covering both. This is why I vote for having a testsuite using both interfaces when they are both available, rather than only the newer one.

Soullivaneuh:
Okay, that is done so.

goetas:
Is this still valid? if not i'm going to close it in some weeks

goetas:
replaced by https://github.com/schmittjoh/serializer/pull/491

PR #496, opened 2015-09-16
removed annotation reference from comment

sloba88:
If you use @Expose in a comment to explain annotation it will still be read as annotation and in this case it throws error that you can't use @Expose on class


goetas:
@ is used to explain better to use the annotation "expose"

PR #497, opened 2015-09-20
Correct public method existance check

ujhgj:
method_exists() also checks inherited public methods and ReflectionClass does not. That makes using of inheritance a bit complicated with serializer.
Commit fix this.


stof:
this needs a test covering this use case, to ensure it does not get broken in the future.

goetas:
since `getDeclaringClass` is used, this should not happen 

https://3v4l.org/01UlT

PR #503, opened 2015-10-05
Added handling zero DateInterval

taavit:
When DateInterval is zeroed current handler returns 'P', which cannot be converted back to DateInterval object. Adding simple zero value for any part solved my issue.


courentin:
This PR already has tests.
Can someone remove the composer.lock conflict and merge this PR .. ?

PR #510, opened 2015-10-19
Manage empty array for serializer

### LINKED ISSUE #124: Configure whether serializing empty array.
The serializer currently allows to configure whether to [serialize null values](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Context.php#L184). 
I think one more option is necessary: to configure whether to serialize empty arrays/strings. Otherwise it's a little incosistent that null values are not serialized but empty arrays are which is currently the case. But both basically mean the same: there is no information available.

So I'd suggest to add a method `setSerializeEmpty` which when passed false, will not serialize empty arrays (`array()`) and empty strings (`''`). So it does not have to do with the php `empty()` function because we want to serialize `0` and `false`.

### LINKED ISSUE #183: Yml serializer don't serialize empty arrays
```php
    public function testEmptyArray()
    {
        $serializer = SerializerBuilder::create()->build();
        $ymlContent = $serializer->serialize(array('t' => array()), 'yml');
        echo $ymlContent; // prints t:

        // Symfony Yaml component
        print_r(Yaml::parse($ymlContent));
        print_r(Yaml::parse('t: []'));
    }
```

Running the test get:

```
t:
Array
(
    [t] => 
)
Array
(
    [t] => Array
        (
        )

)
```

Is this expected behavioral?  Any way to configure it to serialize empty array as []? 

Or is bug and should try to fix it in the code and open PR?

Soullivaneuh:
Fixes #183.
Fixes #124.

Actually, empty array would produce this:

```yml
parameters:
    phonenumber:
```

This is not correct. Furthermore, if you parse it with JMS or Symfony Yaml component, you will got `null` that is not the expected return.

With this PR, you will have this:

```yml
parameters:
    phonenumber: []
```


stof:
A test is needed to prevent regressions

Soullivaneuh:
> A test is needed to prevent regressions

I was looking for it but I don't know how to manage tests files.

Look at the test classes for yml, I don't know how to put the expected result to assert. Any clue?

stof:
Well, there is a ``hash_empty`` test (the test itself using it is at https://github.com/schmittjoh/serializer/blob/master/tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php#L764). We just need a similar test for empty arrays

Soullivaneuh:
Ok so I have to manage all file formats. Will work on it.

Actually, I got hard to understand phpunit errors.

Soullivaneuh:
Well, I got this error on master too. Not related.

For me, this PR is ready to merge.

ping @schmittjoh

Soullivaneuh:
ping @schmittjoh 

Could you please take a look? This issue is quite annoying, a quick patch is needed IMHO.

gallart:
:+1:

src/JMS/Serializer/YamlSerializationVisitor.php
```diff
@@ -111,6 +111,11 @@ public function visitArray($data, array $type, Context $context)
                 ->rtrim(false)
                 ->writeln(' {}')
             ;
+        } elseif (0 === count($data)) {
```
schmittjoh: Can you change this to ``empty($data)`` if we specifically check for 0.

tests/JMS/Serializer/Tests/Serializer/xml/array_empty.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<result>
+  <entry/>
```
schmittjoh: Not sure about this. There is no entry after all? If you are not sure, let's skip this test for XML. That's better than adding a potentially wrong test.

Soullivaneuh: If I remember well, I generated this files with your serializer.

I'll manage to skip it.

src/JMS/Serializer/YamlSerializationVisitor.php
Soullivaneuh: I'm not sure `empty($data)` is the good way. We check for an empty array.

All another empty values should have `~` and not `[]`.

Soullivaneuh: Oh I say something wrong, we are on `visitArray`. Will modify it.

Soullivaneuh:
@schmittjoh Review corrected.

Tobion:
How does this PR fix #124? I don't see it.

schmittjoh:
Yes, that seems unrelated, but seems to have been auto-closed.

Soullivaneuh:
@Tobion @schmittjoh My bad, I was thinking it was related.

PR #512, opened 2015-10-24
Switch to PSR-4

afurculita:
PSR-4 offers a much cleaner directory structure.

All the files from src/JMS/Serializer were moved to the root of the src/ directory. Same with the files from tests/Jms/Serializer.


.travis.yml
```diff
@@ -1,11 +1,12 @@
 language: php
 
+sudo: false
+
 matrix:
   allow_failures:
     - php: hhvm
 
 php:
-  - 5.3
```
Soullivaneuh: This should not be removed.

AFAIK, php 5.3 is still supported by this library and this is not related to your PR.

.travis.yml
```diff
@@ -1,11 +1,12 @@
 language: php
 
+sudo: false
```
Soullivaneuh: Not related to the PR.

.travis.yml
afurculita: As of version 1.0, this library does not support php 5.3. Proof: composer.json is set to require php >=5.4 and code like https://github.com/schmittjoh/serializer/blob/1.0.0/src/JMS/Serializer/Serializer.php#L82 is not supported in 5.3.

I know it is not related to my PR, but it broke a lot of tests

.travis.yml
afurculita: I thought this doesn't need its own PR. Should I create a new PR for it?

Soullivaneuh: I'm not the owner of this repo but IMHO yes, it should be on another PR.

goetas:
@stof In your opinion, switching to PSR4 should be considered a BC break?

goetas:
Moving to PSR4 can be done later. 
Merging this will cause a lot of conflicts with currently opened PR

PR #513, opened 2015-10-28
Fix JSON deserialization of inline objects

skolodyazhnyy:
I've been using JMS Serializer and run into some problems while deserializing some objects from JSON. I have two dimensional table in my application, it has rows and cells with simple string values. Table represented by a class which has `$rows` property, it's an array of `Row` instances. Class `Row` has a single property `values` which is an array of strings.

I want to serialize my object into two dimensional JSON array. I'm using `inline` property for `Row` class to expose values in JSON.

```
Table:
[
     ["a", "b", "c"], <- Row
]
```

Without inline property I'm getting not exactly what I want:

```
Table:
[
     {"values": ["a", "b", "c"]}, <- Row
]
```

I didn't manage to find out if there is any test for `GenericDeserializationVisitor` so I can add a test case for my fix. I will really appreciate if you suggest what is the best way to test my change.

Thanks in advance!

### Steps to reproduce

I'm using this mapping:

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<serializer>
    <class name="Test" xml-root-name="table">
        <property name="x" inline="true">
            <type><![CDATA[array<string>]]></type>
        </property>
    </class>
</serializer>
```

Then I'm serializing an object to json:

```php
$test = new Test();
$test->x = ['a', 'b', 'c'];

echo $serializer->serialize($test, 'json');
```

I'm getting this json:

```json
["a", "b", "c"]
```

Then I'm trying to deserialize it back to the object:

```php
$test = $serializer->serialize($data, Test::class, 'json');

var_dump($text->x);
```

I'm getting an empty array.

### DIFF
--- a/tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php
+++ b/tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php
@@ -246,11 +246,11 @@ abstract class BaseSerializationTest extends \PHPUnit_Framework_TestCase
             $this->markTestSkipped('XML can\'t be tested for empty array');
         }
 
-        $data = array('array' => []);
+        $data = array('array' => array());
         $this->assertEquals($this->getContent('array_empty'), $this->serialize($data));
 
         if ($this->hasDeserializer()) {
-            $this->assertEquals($data, $this->deserialize($this->getContent('array_empty')), 'array');
+            $this->assertEquals($data, $this->deserialize($this->getContent('array_empty'), 'array'));
         }
     }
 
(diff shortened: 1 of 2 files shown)


tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php
```diff
@@ -250,7 +250,7 @@ public function testArrayEmpty()
         $this->assertEquals($this->getContent('array_empty'), $this->serialize($data));
 
         if ($this->hasDeserializer()) {
-            $this->assertEquals($data, $this->deserialize($this->getContent('array_empty')), 'array');
+            $this->assertEquals($data, $this->deserialize($this->getContent('array_empty'), 'array'));
```
skolodyazhnyy: This is not really a part of this PR, just a unit test which was failing for me

skolodyazhnyy:
@schmittjoh Do you think you can find some time review this Pull Request? Thanks in advance!

ericrenard:
Hello,
Is there another solution to this bug ? Suggested PR changes fix the bug for me and it's an important one. It's easily reproductible by creating a class with a property with Inline and Type array<object>, then serialize / deserialize => array is null.

goetas:

The documentation says: 

> Note: This only works for serialization, the serializer will not be able to deserialize objects with this annotation. Also, AccessorOrder will be using the name of the property to determine the order.

PR #511, opened 2015-10-20
Fix tests

Soullivaneuh:
As I said on https://github.com/schmittjoh/serializer/pull/510#issuecomment-149468264, there is some failing tests on master.

This PR fixes them.

BTW, you should configure Travis with the repository to prevent it on the future.

PR #515, opened 2015-10-29
Fix multiple handler callbacks in YamlDriver

mpajunen:
Using handler callbacks for multiple formats caused errors in `YamlDriver` because the `direction` key was parsed multiple times. Now `direction` key is only parsed once (per direction).

A basic test case for handler callbacks is included for all configuration formats.

### DIFF
--- a/src/JMS/Serializer/Metadata/Driver/YamlDriver.php
+++ b/src/JMS/Serializer/Metadata/Driver/YamlDriver.php
@@ -203,9 +203,9 @@ class YamlDriver extends AbstractFileDriver
         }
 
         if (isset($config['handler_callbacks'])) {
-            foreach ($config['handler_callbacks'] as $direction => $formats) {
+            foreach ($config['handler_callbacks'] as $directionName => $formats) {
+                $direction = GraphNavigator::parseDirection($directionName);
                 foreach ($formats as $format => $methodName) {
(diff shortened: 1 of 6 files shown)


mpajunen:
ping @schmittjoh

Would be nice to see this merged, the fix is simple and obvious. If the new tests need some more work, I'll be happy to improve them.

The test failures seem unrelated and are probably fixed by #511 in any case.

mpajunen:
Rebased, tests are now green as well.

src/JMS/Serializer/Metadata/Driver/YamlDriver.php
```diff
@@ -204,8 +204,8 @@ protected function loadMetadataFromFile(\ReflectionClass $class, $file)
 
         if (isset($config['handler_callbacks'])) {
             foreach ($config['handler_callbacks'] as $direction => $formats) {
+                $direction = GraphNavigator::parseDirection($direction);
```
stof: as this store a different thing, it would be even better to use a different variable name instead of overwriting the variable

mpajunen: Loop variable name changed to `$directionName`.

goetas:
will be part of v1.3, tnx

mpajunen:
Nice to see this merged finally, thanks.

PR #521, opened 2015-11-16
Fix wrong overriding of the virtual property

Nek-:
Currently, when there is a private property inside the class and you don't access to it, you can't defined a virtual property (the test I added fails on the master branch) that will replace the mapping of the property. This is a problem when you uses groups or choose different policies.

The real problem is from this class: https://github.com/schmittjoh/metadata/blob/master/src%2FMetadata%2FClassMetadata.php#L53 and an eventual fix would be that:

``` php
    public function addPropertyMetadata(PropertyMetadata $metadata)
    {
        if (!isset($this->propertyMetadata[$metadata->name])) {
            $this->propertyMetadata[$metadata->name] = $metadata;
        }
    }
```

The problem is that it would be a huge BC Break because it doesn't allow anymore to override a metadata (that could maybe be useful ? I really don't know).

So I added a little fix in this PR, as the problem is present only on annotation mapping, I added the fix only in the `AnnotationDriver`. It's not awesome but it work and it's efficient.

### DIFF
--- a/tests/JMS/Serializer/Tests/Fixtures/VitualPropertyIsPriority.php
+++ b/tests/JMS/Serializer/Tests/Fixtures/VitualPropertyIsPriority.php
@@ -1,0 +1,96 @@
+<?php
+
+/*
+ * Copyright 2013 Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


tests/JMS/Serializer/Tests/Fixtures/VitualPropertyIsPriority.php
```diff
@@ -0,0 +1,96 @@
...
+ * limitations under the License.
+ */
+
+namespace JMS\Serializer\Tests\Fixtures;
+
+use JMS\Serializer\Annotation\Groups;
+use JMS\Serializer\Annotation\VirtualProperty;
+use JMS\Serializer\Annotation\SerializedName;
+
+class VitualPropertyIsPriority
```
tontonjulien: typo

valanz: Oopsy there is a typo

goetas:
related to https://github.com/schmittjoh/serializer/issues/896

Nek-:
@goetas want me to take care of this fix again ?

goetas:
No , thanks, in the branch 2.0 is already fixed.

goetas:
The comment was more for me to keep track of tickets to close when 2.0 will be closed

PR #527, opened 2015-11-21
Fix: No short array syntax on PHP5.3

localheinz:
This PR

* [x] converts a short array to a long array using as it breaks the build on PHP5.3

See https://travis-ci.org/schmittjoh/serializer/jobs/92420623#L265.

Related to #511.

Used

```
$ php-cs-fixer fix --verbose --fixers=long_array_syntax tests
```

for that.

### DIFF
--- a/tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php
+++ b/tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php
@@ -246,7 +246,7 @@ abstract class BaseSerializationTest extends \PHPUnit_Framework_TestCase
             $this->markTestSkipped('XML can\'t be tested for empty array');
         }
 
-        $data = array('array' => []);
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Oops. :-)

rpiotaix:
If i'm not mistaken, the composer.json file states that the minimum PHP version supported is PHP 5.4

Soullivaneuh:
@rpiotaix You're right.

Ener-Getick:
Should be closed @localheinz as this bundle doesn't support php 5.3

localheinz:
@rpiotaix 

You're right, which means that PHP5.3 should be removed from [`.travis.yml`](https://github.com/schmittjoh/serializer/blob/master/.travis.yml#L8).

rpiotaix:
@localheinz see #532

PR #539, opened 2015-12-15
[feature] exclusion groups

leberknecht:
Added support for passing an array parameter to the `@Exclude` annotation, refs #470

Example: 

``` php
use JMS\Serializer\Annotation\Exclude;

/**
* @ExclusionPolicy("none")
*/
class BlogPost
{
    private $title;

    /** @Exclude({"list"}) */
    private $comments;
}
```

And in the controller: 

``` php

$serializer->serialize(new BlogPost(), 'json', SerializationContext::create()->setExclusionGroups(array('list')));
```

This will only serialize `$title`.


goetas:
is this just a mirror implementation of http://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#creating-different-views-of-your-objects ?

leberknecht:
Kind of, yes. Where "@groups({"test"})" is whitelisting, "@exclude({"test"})" is blacklisting. Can be handy if you serialize a lot without group, but you want to suppress one specific field in one specific case. The work-around (and/or probably the better solution in the long run for most cases) would be to set proper group for all-other-fields. But still would be helpful to have this option available.

src/JMS/Serializer/Exclusion/ExclusionGroupsExclusionStrategy.php
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * @param $metadataGroups
+     * @return bool
+     */
+    private function shouldSkipGroup($metadataGroups) {
+        if (false === $metadataGroups) {
+            return false;
+        }
+        if (is_array($metadataGroups) && empty($metadataGroups)) {
```
goetas: isn't more appropriate use `count` instead of `empty` ?

src/JMS/Serializer/Exclusion/ExclusionGroupsExclusionStrategy.php
```diff
@@ -0,0 +1,63 @@
...
+            return true;
+        }
+        if (!empty($this->groups) && array_intersect(array_keys($this->groups), $metadataGroups)) {
```
goetas: isn't more appropriate use `count` instead of `empty` ?

src/JMS/Serializer/Metadata/ClassMetadata.php
```diff
@@ -225,6 +226,7 @@ public function serialize()
             $this->accessorOrder,
             $this->customOrder,
             $this->handlerCallbacks,
+            $this->exclusionGroups,
```
goetas: this will cause an error if will be deployed without cache cleaning, see https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Metadata/PropertyMetadata.php#L169 and https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Metadata/PropertyMetadata.php#L137 for a safer implementation

goetas:
Do you have some benchmark for performance impact?

leberknecht:
>Do you have some benchmark for performance impact?

Nope :/ Can you give me a hint how to provide one? when running `php tests/benchmark.php json 200` i get 
```php
 % php tests/benchmark.php json 200 
PHP Fatal error:  Uncaught TypeError: Argument 4 passed to JMS\Serializer\Tests\Fixtures\BlogPost::__construct() must be an instance of JMS\Serializer\Tests\Fixtures\Publisher, none given, called in /home/delf/dev/serializer/tests/benchmark.php on line 34 and defined in /home/delf/dev/serializer/tests/JMS/Serializer/Tests/Fixtures/BlogPost.php:116
```
checking benchmark.php, it indeed looks broken: 
```php
function createObject()
{
    $post = new \JMS\Serializer\Tests\Fixtures\BlogPost('FooooooooooooooooooooooBAR', new \JMS\Serializer\Tests\Fixtures\Author('Foo'), new \DateTime);
```
..is it broken or am i just missing something?

goetas:
This feature can be achieved with https://github.com/schmittjoh/serializer/pull/673


``` php
use JMS\Serializer\Annotation\Exclude;

/**
* @ExclusionPolicy("none")
*/
class BlogPost
{
    private $title;

    /** @Exclude(if="context.attributes.containsKey('list')") */
    private $comments;
}
```

And in the controller: 

``` php

$serializer->serialize(new BlogPost(), 'json', SerializationContext::create()->setAttribute('list', true));
```

This will only serialize `$title`.

PR #541, opened 2015-12-23
Add default SerializerContext to serializer, update builder

alcalyn:
This PR adds a default `SerializerContext` factory and `SerializerContext` factory in `Serializer` class to avoid to pass a context every time we call `serialize`, `deserialize`, `toArray` or `fromArray`.

It also update the serializer builder by adding `setDefaultSerializerContextFactory` and `setDefaultDeserializerContextFactory` methods, allowing to pass a callable which create a context.

Now we can set a **default serializer context** factory by doing this:

``` php

$serializer = \JMS\Serializer\SerializerBuilder::create()
    ->setDefaultSerializationContextFactory(function () {
        return SerializationContext::create()
            ->setSerializeNull(false)
        ;
    })
    ->build()
;
```

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -78,26 +78,16 @@ you can set a ``SerializationContextFactory`` to the Serializer.
 
 Example using the SerializerBuilder::
 
-    $serializer = JMS\Serializer\SerializerBuilder::create()
-        ->setDefaultSerializationContextFactory(new MySerializationContextFactory())
-        ->build()
-    ;
-
-And ``MySerializationContextFactory`` must implements interface
-``JMS\Serializer\ContextFactory\SerializationContextFactoryInterface``::
-
-    use JMS\Serializer\ContextFactory\SerializationContextFactoryInterface;
     use JMS\Serializer\SerializationContext;
 
-    class MySerializationContextFactory implements SerializationContextFactoryInterface
-    {
-        public function createSerializationContext()
-        {
+    $serializer = JMS\Serializer\SerializerBuilder::create()
+        ->setDefaultSerializationContextFactory(function () {
             return SerializationContext::create()
                 ->setSerializeNull(true)
             ;
-        }
-    }
+        })
+        ->build()
+    ;
 
 Then, calling ``$serializer->serialize($data, 'json');`` will use your SerializationContext.
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -70,3 +70,40 @@ are located::
 The serializer would expect the metadata files to be named like the fully qualified class names where all ``\`` are
 replaced with ``.``. So, if you class would be named ``Vendor\Package\Foo``, the metadata file would need to be located
 at ``$someDir/Vendor.Package.Foo.(xml|yml)``. For more information, see the :doc:`reference <reference>`.
+
+Setting default SerializationContext factory
+--------------------------------------------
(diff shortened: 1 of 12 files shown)


alcalyn:
I made it as simple as possible by using php `Closure` for now, then we can use the builder like that:

``` php
$serializer = \JMS\Serializer\SerializerBuilder::create()
    ->setDefaultSerializationContextFactory(function () {
        return \JMS\Serializer\SerializationContext::create()
            ->setSerializeNull(true)
        ;
    })
    ->build()
;

**Comment by nuxwin** [comment]
@alcalyn

Please, don't enforce closure for the factory. Eg, in zf2 service manager, it is recommended to use factory classes which is much better.

**Comment by alcalyn** [comment]
Yep I started to create interfaces `SerializationContextFactoryInterface` and `DeserializationContextFactoryInterface`, but I find it too verbose at first

**Comment by nuxwin** [comment]
At first... Don't worry about verbosity. It is much better to rely on interface ;)

Thanks for your work anyway. I need this.

**Comment by alcalyn** [comment]
Now we can set a default serializer context factory by doing this:

``` php
class SerializationContextFactory implements SerializationContextFactoryInterface
{
    public function createSerializationContext()
    {
        return SerializationContext::create()
            ->setSerializeNull(true)
        ;
    }
}
```

Then:

``` php
$serializer = \JMS\Serializer\SerializerBuilder::create()
    ->setDefaultSerializationContextFactory(new SerializationContextFactory())
    ->build()
;
```

alcalyn:
Last commit allow to use a callable in Serializer builder fr a shorter syntax. Example:

``` php
$serializer = \JMS\Serializer\SerializerBuilder::create()
    ->setDefaultSerializationContextFactory(function () {
        return SerializationContext::create()
            ->setSerializeNull(false)
        ;
    })
    ->build()
;
```

alcalyn:
Merged in Jms Serializer organisation repo.

alcalyn:
What about this PR ? @schmittjoh

goetas:
To me looks a bit too much 500 lines of code, just to have a default implementation of the context... that anyway changes for each usecase

alcalyn:
Mmh in theses 500 lines, its about 200 lines for testing, 120 for license, a bunch of one-method interfaces...

And I just needed it this morning because I don't want to create a wrapper around serializer service to only create a new context with `setSerializeNull(true)` on each serialize.

Maybe another solution is to set `serializeNull` in config instead of context, but also `setGroups` and `enableMaxDepthChecks` could be in config, because as new user, I don't know why a part is in config, and another part in context...

So that's why I think a default context factory in builder is not déconnant :D

nuxwin:
@goetas 

For our project, we have interest too in that PR. Please, merge it ;)

@alcalyn ---> Déconnant ???

alcalyn:
is not irrevelant* :P

goetas:
Personally I can agree with having a context factories, but what is the use case for the callable context factories?

They are already part of a specific implementation, so probably they should stay in your project, while the default factories should stay here... is it?

alcalyn:
the callable context factory is to avoid to create a class to just call `setSerializeNull(true)`, and provides a more "builder" way to add a default context.

goetas:
What do you think about renaming some variables as in https://github.com/schmittjoh/serializer/pull/645 ?

(i have just removed the "default" prefix that was looking redundant to me)

alcalyn:
Okey, its shorter

alcalyn:
I close it in favor to https://github.com/schmittjoh/serializer/pull/645

PR #542, opened 2015-12-27
PHP 7 + Symfony 3.0 Support 

jmcclell:
I can split these up into two PR's if you prefer, figured I'd have you review first and see what you want / don't want. I needed he ability to swap objects midflight so I added that, but I'm in a PHP7+Symfony3 env so I needed to make sure things worked in this environment.

Not yet sure on BC until I see what Travis does.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,11 +32,11 @@
         "doctrine/phpcr-odm": "^1.3",
         "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "propel/propel1": "~1.7",
-        "symfony/yaml": "^2.1",
-        "symfony/translation": "^2.1",
-        "symfony/validator": "^2.2",
-        "symfony/form": "~2.1",
-        "symfony/filesystem": "^2.1",
+        "symfony/yaml": "^2.1|^3.0",
+        "symfony/translation": "^2.1|^3.0",
+        "symfony/validator": "^2.2|^3.0",
+        "symfony/form": "~2.1|^3.0",
+        "symfony/filesystem": "^2.1|^3.0",
         "phpunit/phpunit": "^4.8|^5.0"
     },
     "autoload": {
--- a/src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
+++ b/src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
@@ -21,7 +21,7 @@ namespace JMS\Serializer\EventDispatcher\Subscriber;
 use JMS\Serializer\EventDispatcher\Event;
 use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
 use JMS\Serializer\Exception\ValidationFailedException;
-use Symfony\Component\Validator\ValidatorInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 class SymfonyValidatorSubscriber implements EventSubscriberInterface
 {
--- a/src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
+++ b/src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
@@ -21,14 +21,19 @@ namespace JMS\Serializer\EventDispatcher\Subscriber;
 use JMS\Serializer\EventDispatcher\Event;
 use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
 use JMS\Serializer\Exception\ValidationFailedException;
-use Symfony\Component\Validator\Validator\ValidatorInterface;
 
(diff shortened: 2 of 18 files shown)


goetas:
can you rebase this?

Bukashk0zzz:
@jmcclell ping

stof:
I think you should split things to make the review easier (and allowing to merge the support of newer versions of Symfony separately from a new feature needing discussion)

src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
```diff
@@ -21,7 +21,7 @@
 use JMS\Serializer\EventDispatcher\Event;
 use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
 use JMS\Serializer\Exception\ValidationFailedException;
-use Symfony\Component\Validator\ValidatorInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
```
stof: This breaks support for Symfony 2.3

stof:
You should also remove swap files from the PR (and I suggest you to add an ignore rule on your system to avoid such mistakes in the future, see https://help.github.com/articles/ignoring-files/)

goetas:
What is the use case of exchanging completely the object during serialization? I see it as a little risky

jmcclell:
Wow, this is suddenly getting a lot of attention.

@goetas What do you want me to do with regard o rebasing? Split it into two PRs?
@stof Thanks, I hadn't noticed the swap file. My vimrc is setup to store swap files in a central location, but I had done this work on another laptop and didn't notice. Good idea re: global ignore file.
@goetas The use case at the time was for me to be able to automatically wrap an object in a HATEOAS envelope upon serialization, but I abandoned that approach in the end. You are right, it is a bit risky (especially wrapping due the high-chance of looping endlessly without a proper check in place by the user) but may have some value to others for various reasons. I won't be disappointed  if it isn't accepted as a PR, however, as it likely has little demand.

goetas:
@jmcclell thanks for your answers

Then I'm for closing this PR

src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
jmcclell: @stof True, but 2.3 is EOM this month and EOL in a year. It seems that cutting a release of the Serializer library that supports 3.0 doesn't necessarily need to support 2.3. That being said, do you have a strategy / suggestion for how to fix this to work with both? Wrapping the use statement in a class_exists conditional or something, perhaps?

jmcclell:
@goetas I've separated out the object swap into another PR that you are free to close if you want. What is left here are teh changes I made to support Symfony 3.0/PHP7, including several test fixes.

jmcclell:
I was just waking up when I saw your messages and tried to clean this up. Apparently I wasn't awake enough :) I'll patch it up and make sure it's passing tests this evening so you guys can get the Symfony 3 support in. @stof What are your thoughts on 2.3 support, re: the line comment earlier?

composer.json
```diff
@@ -28,15 +28,15 @@
...
         "propel/propel1": "~1.7",
-        "symfony/yaml": "2.*",
-        "symfony/translation": "~2.0",
-        "symfony/validator": "^2.2",
-        "symfony/form": "~2.1",
-        "symfony/filesystem": "2.*",
-        "phpunit/phpunit": "^4.0|^5.0"
+        "symfony/yaml": "2.*|~3.0",
+        "symfony/translation": "~2.0|~3.0",
+        "symfony/validator": "~2.0|3.0",
```
stof: please don't reset the the lowest bounds. You fixed conflicts in the wrong way

jmcclell: Ah yes, I see they have since updated from when I first made these changes. I'll fix to match what's currently in master for low bounds.

src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php
```diff
@@ -21,14 +21,19 @@
...
-use Symfony\Component\Validator\ValidatorInterface;
 
 class SymfonyValidatorSubscriber implements EventSubscriberInterface
 {
     private $validator;
 
-    public function __construct(ValidatorInterface $validator)
+    public function __construct($validator)
     {
+        if (!$validator instanceof \Symfony\Component\Validator\Validator\ValidatorInterface // 3.x
```
jmcclell: @stof This was the only way I could think of to maintain 2.3 compatibility with the new Validator structure. Well, easily, at least. Not a big fan. Thoughts?

goetas:
moving development to https://github.com/schmittjoh/serializer/pull/682

Issue #343, opened 2014-09-19
Type DateTime and Timestamp (U format)


four43:
We are outputting a DateTime as a timestamp in our application. This result is a string, but it would be great if it was an integer.

``` PHP
/**
 * @JMS\VirtualProperty
 * @JMS\SerializedName("dateCreated")
 * @JMS\Type("DateTime<'U'>")
 */
public function getDateCreated() {
    return $this->dateCreated;
}
```

Returns:

``` javascript
{
    "dateCreated":"1411139100"
}
```

Maybe keep the type of the PHP date call? date('U') returns a number.

Thanks for all your work on this. This is a great library and we're having a lot of fun with it.

-Seth

EDIT: date('U') returns a number, DateTime::format('U') returns a string. _sigh PHP..._

goetas:
> EDIT: date('U') returns a number, DateTime::format('U') returns a string. sigh PHP...

looks to be a PHP behavior

goetas:
Easy workarounds starting from 1.6.0:

```php

/**
 * @Serializer\VirtualProperty(
 *     "dateCreated",
 *     exp="object.getDateCreated().getTimestamp()",
 *     options={@Serializer\Type("integer")}
 *  )
 */
class Foo {
  public function getDateCreated() {
      return $this->dateCreated;
  }
}
```

PR #552, opened 2016-02-11
Fix bug #343 return integer when the column is datetime and the format is timestamp

Bukashk0zzz:


Bukashk0zzz:
There are similar pull request https://github.com/schmittjoh/serializer/pull/378 but there are no test and it old.

Bukashk0zzz:
Decided change pull request to more global

PR #553, opened 2016-02-12
[doc] fix AccessorOrder documentation

aledeg:
Before, the documentation mentionned that in order to call the method in the AccessorOrder annotation, we should use only the method name stripped of its "get" prefix. But it forgot to mention that the stripped method should start with a lower case letter.
I fixed that minor typo.


nicholasruunu:
Yeah, probably wasted a good 40 minutes on this. >.< :P
Doesn't seem like it's pushed to the site yet.

PR #556, opened 2016-03-02
Add support for simple handler inheritance

Ener-Getick:
This PR adds support for simple handler inheritance.

For example if you create an handler to normalize ``\Exception`` instances, and if you try to serialize a ``\LogicException`` instance, your first handler will be executed unless you have a more specific handler to normalize ``\LogicException``.

### DIFF
--- a/src/JMS/Serializer/Handler/HandlerRegistry.php
+++ b/src/JMS/Serializer/Handler/HandlerRegistry.php
@@ -79,10 +79,10 @@ class HandlerRegistry implements HandlerRegistryInterface
 
     public function getHandler($direction, $typeName, $format)
     {
-        if ( ! isset($this->handlers[$direction][$typeName][$format])) {
-            return null;
-        }
-
-        return $this->handlers[$direction][$typeName][$format];
+        do {
+            if (isset($this->handlers[$direction][$typeName][$format])) {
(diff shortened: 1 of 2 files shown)


Ener-Getick:
The build failure is not related.

Ener-Getick:
ping @schmittjoh

maennchen:
I'd like to implement that PR into my Fork. Would you be able to write tests for that and create a new PR on https://github.com/jms-serializer/serializer?

Ener-Getick:
@maennchen I don't think so I did this quickly but I do not have a lot of time to write tests :-/

schmittjoh:
I think you can already do that with a listener for the pre-serialize event that updates the type to the one your handler expects. Did you consider using that?

```
if ($object instanceof \Exception) { $event->setType('MyExceptionHandler'); }
```

Ener-Getick:
@schmittjoh i didn't now that but the advantage of my solution is that you can use a generic handler and then override it for a specific class by having another handler.

goetas:
This feature will cause performance  issues since it has to loop trough the class hierarchy in order to find a handler for the object. This operation is done for each object in the graph.

Closing it since a similar behavior can be already obtained with user-land code.

Issue #559, opened 2016-03-04
Avoiding use of CDATA in @Discriminator field


lordelph:
Is there any way to avoid getting CDATA in an discriminator field?

For example, if I have classes where the base class defines a discriminator like this
```
/** 
* @JMS\Discriminator(field = "type", map = {
 *   "foo": "MyBundle\Entity\FooEntity",
 *   "bar": "MyBundle\Entity\BarEntity"
 * })
*/
```

Serializing a FooEntity as XML would include a type like this

`<type><![CDATA[foo]]></type>`

Since I'm in control of the labels used for types, I'd like to avoid CDATA if possible. Is there a way to do that?

lordelph:
As there didn't seem to be a way of doing it, I've proposed a method in this pull request https://github.com/schmittjoh/serializer/pull/560

PR #560, opened 2016-03-04
Fix for #559 to enable @discriminator fields to not be serialized as CDATA

lordelph:
This adds options to the @discriminator annotation (and its YML and XML configuration equivalents) which allow the use of CDATA in the discriminator field to be disabled. By default, it is will be on to preserve existing behaviour. 

So, now you can do this

```
/** 
 * @JMS\Discriminator(field = "type", cdata=false, map = {
 *   "foo": "MyBundle\Entity\FooEntity",
 *   "bar": "MyBundle\Entity\BarEntity"
 * })
 */
```

Serializing a FooEntity as XML would include a type like this::

`<type>foo</type>`

If the `cdata=true`, or if it is omitted, the original behaviour will use CDATA for those elements

`<type><![CDATA[foo]]></type>`

Feedback on this push request appreciated - in particular, this needs a unit test. How about if I add a new set of polymorphic types, e.g. Animal/Dog/Cat which _don't_ use CDATA, some xml fixtures for them, and additional test methods in BaseSerializationTest?

### DIFF
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -245,6 +245,20 @@ to the least super type::
     class Car extends Vehicle { }
     class Moped extends Vehicle { }
 
+By default, an XML serialization would use CDATA for discriminator field, for example::
+
+    <type><![CDATA[car]]></type>
+  
+You can indicate you don't require CDATA with an additional `cdata=false` in the annotation. So, annotating like this::
+
+    /**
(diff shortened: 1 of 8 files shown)


goetas:
Looks good, but tests are missing.

why not prefix `cdata` as `xmlCdata` ?

- have you considered allowing to set it as xml attribute ?
- have you considered adding an option `xmlNamespace` ?

goetas:
Implemented in https://github.com/schmittjoh/serializer/pull/692 as `@XmlDiscriminator`

PR #572, opened 2016-04-21
Serialize null in positional arrays

### LINKED ISSUE #571: Invalid null serialization in arrays
This code 
https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/GenericSerializationVisitor.php#L104-L106

```php
            if (null === $v && ( ! is_string($k) || ! $context->shouldSerializeNull())) {
                continue;
            }
```

Test
```php
$context = \JMS\Serializer\SerializationContext::create();
var_dump($serializer->toArray([1, null, 2], $context));

$context = \JMS\Serializer\SerializationContext::create();
var_dump($serializer->serialize([1, null, 2], 'json', $context));
```

Output
```
array(2) {
  [0] =>
  int(1)
  [2] =>
  int(2)
}

string(13) "{"0":1,"2":2}"
```

Dropping null value makes the `json_encode` think of the data like the named properties. I think correct condition should be

if (null == $v && is_string($k) && !$context->shouldSerializeNull()) {
   continue;
}

In this case we drop null only in case they are at hash array, not positional.

scaytrase:
Fix #571

### DIFF
--- a/src/JMS/Serializer/GenericSerializationVisitor.php
+++ b/src/JMS/Serializer/GenericSerializationVisitor.php
@@ -101,7 +101,7 @@ abstract class GenericSerializationVisitor extends AbstractVisitor
         foreach ($data as $k => $v) {
             $v = $this->navigator->accept($v, $this->getElementType($type), $context);
 
-            if (null === $v && ( ! is_string($k) || ! $context->shouldSerializeNull())) {
+            if (null === $v && is_string($k) && !$context->shouldSerializeNull()) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


scaytrase:
I'm ready to fix some related tests but I'm not sure which failures belong to the patch as the whole build history is red
https://travis-ci.org/schmittjoh/serializer/builds

maennchen:
@scaytrase You can now reabse on the original master and everything should get green.

scaytrase:
Thanks, all green.

scaytrase:
The given commit actually does not resolve the problem.

It fails a simple test

```php
    public function testNullArraySerialization()
    {
        $array = [0, null, 'a'];
        self::assertSame($array, $this->serializer->toArray($array));
    }
```

How we force an array being normalized as array, not object?

PR #573, opened 2016-04-21
Don't test on 5.3

scaytrase:
5.3 Make the travis failed as the BaseSerializationTest use short array syntax at line 249 and the lower php version is bumped to 5.4.0.
travis composer ignores the 5.3 conflict as it installs deps from lock

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ matrix:
     - php: hhvm
 
 php:
-  - 5.3
   - 5.4
   - 5.5
   - 5.6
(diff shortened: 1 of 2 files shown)


stof:
> travis composer ignores the 5.3 conflict as it installs deps from lock

this means that the lock file is outdated though

scaytrase:
I thought lock does not actually lock the php version. But can it ignore the platform reqs when installing from it?

scaytrase:
Ow, it locks the "platform"

scaytrase:
Looks like `composer update --lock` did the trick in a right way

schmittjoh:
Closing this one as this has been added.

PR #576, opened 2016-04-27
Prevent proxy loading if handler exists

### LINKED ISSUE #575: DoctrineProxySubscriber forces loading of the proxy even if custom handler exist
https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/EventDispatcher/Subscriber/DoctrineProxySubscriber.php#L56

This line of code forces the proxy loading causing additional DB query.

We have a special handler which serializes relation to the set of identifier values to avoid DB query loading, but this handler forces the proxy being loaded all times.

btw, collection has type check, but the pure proxy - does not.

scaytrase:
Also simplified the logic a bit for this use case.

Fix #575

### DIFF
--- a/src/JMS/Serializer/EventDispatcher/Subscriber/DoctrineProxySubscriber.php
+++ b/src/JMS/Serializer/EventDispatcher/Subscriber/DoctrineProxySubscriber.php
@@ -36,16 +36,15 @@ class DoctrineProxySubscriber implements EventSubscriberInterface
         // If the set type name is not an actual class, but a faked type for which a custom handler exists, we do not
         // modify it with this subscriber. Also, we forgo autoloading here as an instance of this type is already created,
         // so it must be loaded if its a real class.
-        $virtualType = ! class_exists($type['name'], false);
+        if (! class_exists($type['name'], false)) {
+            return;
+        }
 
         if ($object instanceof PersistentCollection
             || $object instanceof MongoDBPersistentCollection
             || $object instanceof PHPCRPersistentCollection
         ) {
-            if ( ! $virtualType) {
-                $event->setType('ArrayCollection');
-            }
-
+            $event->setType('ArrayCollection');
             return;
         }
 
@@ -54,10 +53,7 @@ class DoctrineProxySubscriber implements EventSubscriberInterface
         }
 
         $object->__load();
-
-        if ( ! $virtualType) {
-            $event->setType(get_parent_class($object));
-        }
+        $event->setType(get_parent_class($object));
     }
 
     public static function getSubscribedEvents()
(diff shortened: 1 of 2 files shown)


maennchen:
This change will fail one test.

```
There was 1 failure:

1) JMS\Serializer\Tests\Serializer\EventDispatcher\Subscriber\DoctrineProxySubscriberTest::testDoesNotRewriteCustomType
Failed asserting that false is true.

/Users/maennchen/Development/serializer/tests/Serializer/EventDispatcher/Subscriber/DoctrineProxySubscriberTest.php:49
```

```php
public function testDoesNotRewriteCustomType()
{
    $event = $this->createEvent($obj = new SimpleObjectProxy('a', 'b'), array('name' => 'FakedName', 'params' => array()));
    $this->subscriber->onPreSerialize($event);

    $this->assertEquals(array('name' => 'FakedName', 'params' => array()), $event->getType());
    $this->assertTrue($obj->__isInitialized());
}
```

scaytrase:
Yes, the fix is to turn in up to `assertFalse` as proxy is not loaded as the `FakedName` is a kind of custom type which should be handled elsewhere.

Am I correct?

maennchen:
Yes, I think changing to `assetFalse` should be the right way.

schmittjoh:
What happens if the proxy isn't initialized and we do want to serialize its fields. Does this break this use-case, do we have tests on that?

scaytrase:
If the proxy isn't loaded and does not have custom handler (or any fake type), it will be loaded, yes, we do have test on this.
If the proxy isn't loaded and does have a custom handler (or any fake type), it would be passed to it as is. I've fix the test to represent is.

In our case we have a handler, which serialize proxy objects into the set of identifier values, i.e

```php
relation => 1
```

or, for composite keys.

```php
relation => [
   'id_1' => 1
   'id_2' => other_key
]
```

So the proxy initializing is not really needed if custom handler works for relation.

scaytrase:
Rebased

schmittjoh:
So, IIUC we have a BC break for people that have a custom handler and do not explicitly initialize the proxy?

Have you thought about handling this one level higher on the entity where you have the relation instead of on the entity itself?

scaytrase:
IIUC current code should throw an unknown type\handler exception. for this we have no tests AFAIK

scaytrase:
@goetas could you point me with instuctions how can I improve this patch? Loading proxies is really weird when serializing `toOne` doctrine relations to their IDs, as this triggers lazy-loading of many entities.

I'm not really see what we incopatible break here

goetas:
> I'm not really see what we incopatible break here

@scaytrase I'm also not really sure about this PR and with the impact it could have on existing code.
 But I see a value in this PR. I need some time to check that everything works as expected.

Will not be part of 1.4 but i guess 1.5 can be a feasible target release.

Will have to check it manually (and really carefully).

Do you have this code in production on some of your fork?

scaytrase:
I've tested this code locally with out corp code to see it removes thousands of lazy loading queries, but we do not currently have other custom handlers to see the impact on them. If you give me the cases, I think I can propose more tests for them

PR #578, opened 2016-05-05
Fixed test suite on master

goetas:
- libxml2 is not reliable for DTD info, fallback on PHP (not using DOM we gain some performance)
- removed travis tests for php 5.3
- removed lock file (since php7 requires different dependencies)
- enabled travis cache
- php7 sorting algorithm is/was buggy (stable vs unstable sorting) fixed this cases
- updated deprecated usages of phpcr odm tests
- fixed wrong tests with `ObjectWithVirtualProperties` (error become evident on php7)

*note on test changes regarding virtual properties*
`ObjectWithVirtualProperties` has 4 serializable fields (here fields are listed in order as the appear in the `ObjectWithVirtualProperties` class:
1. existField
2. VirtualValue
3. test
4. TypedVirtualProperty

The object has a custom order access rule:
1. prop_name
2. existField
3. foo

Since "prop_name" and "foo" do not exist, the serialization order should be: 
1. existField
2. VirtualValue
3. test
4. TypedVirtualProperty

But all tests were expecting:
1. existField
3. test
2. VirtualValue
4. TypedVirtualProperty

that is wrong since "VirtualValue" appears before "test" in the code


src/JMS/Serializer/Metadata/ClassMetadata.php
```diff
@@ -266,12 +266,13 @@ private function sortProperties()
...
                 $order = $this->customOrder;
-                uksort($this->propertyMetadata, function($a, $b) use ($order) {
+                $currentSorting = $this->propertyMetadata ? array_combine(array_keys($this->propertyMetadata), range(1, count($this->propertyMetadata))) : [];
+                uksort($this->propertyMetadata, function($a, $b) use ($order, $currentSorting) {
                     $existsA = isset($order[$a]);
                     $existsB = isset($order[$b]);
 
                     if ( ! $existsA && ! $existsB) {
-                        return 0;
+                        return $currentSorting[$a] - $currentSorting[$b];
```
goetas: this makes the sorting always "stable"

goetas:
https://github.com/schmittjoh/serializer/pull/573 will become obsolete if this gets merged

schmittjoh:
Thanks for the detailed explanations, makes this much faster to review :)

composer.json
```diff
@@ -16,25 +16,24 @@
...
         "doctrine/orm": "~2.1",
-        "doctrine/phpcr-odm": "~1.0.1",
-        "jackalope/jackalope-doctrine-dbal": "1.0.*",
+        "doctrine/phpcr-odm": "^1.3",
+        "jackalope/jackalope-doctrine-dbal": "^1.0",
         "propel/propel1": "~1.7",
         "symfony/yaml": "2.*",
         "symfony/translation": "~2.0",
-        "symfony/validator": "~2.0",
+        "symfony/validator": "~2.0.0|~2.1.0",
```
stof: this does not allow installing any maintainer version of the package, as it only allows 2. and 2.1. this is pretty bad

composer.json
```diff
@@ -16,25 +16,24 @@
...
         "symfony/form": "~2.1",
         "symfony/filesystem": "2.*",
         "phpunit/phpunit": "~4.0"
```
stof: supporting PHP 7 means you should allow PHPUnit 5 too. version 4 is not officially compatible with PHP 7

goetas: Good point. actually I'm working on making this compatible with HHVM too.
Next days (probably half next week) will do a second PR for it

composer.json
goetas: as i understand this allows versions between 2.0.0 and 2.1.999 is it?

currently the test suite will fail with the validator 2.3.0

composer.json
stof: Well, this should be considered as a bug. Symfony 2.1 is dead since many years (2.3 will die in a few months and it was a LTS release while 2.1 was not).
Btw, this was another drawback of committing a lock file: as the lock file was forcing to use outdated libraries, this library is actually broken since years without the CI catching it...

composer.json
scaytrase: I think that the best option is to allow ~2.3|~3.0 and test against against several versions with travis. And fix if it does not work. See like FOSUB does https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/.travis.yml. At least you can do `--prefer-stable` and `--prefer-lowest` test to not creating a matrix mess for each major dependency

goetas: i agree that here at least 2.3 should be supported, but this requires fixing all the tests that are using `ConstraintViolation` in https://github.com/schmittjoh/serializer/blob/master/tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php

since `ConstraintViolation` is not backward compatible from 2.1 to 2.3

PR #567, opened 2016-04-04
Discriminator Groups

maennchen:
Fix for Issue schmittjoh/JMSSerializerBundle#506

PR #301, opened 2014-07-16
Generate namespaced element on XmlList entries

goetas:
Things done with this patch:
- Merged #237 
- Added documentation for #237 
- `@XmlNamespace(uri="http://example.com/namespace")` acts as default namespace
- `XmlDeserializationVisitor`
 - Removed a lot of XPath stuff in flavor of simpler `SimpleXMLElement::children` method
 - Added `$objectMetadataStack` to access class metadata from a property
- `XmlSerializationVisitor`
 - Removed unnecessary element prefixing when the document/node default namespace match the desired namesapce
 - Removed some duplicated code, adding `createElement` and `setAttributeOnNode` methods

PR #579, opened 2016-05-07
Discriminator Groups

maennchen:
Clone of schmittjoh/serializer#567

### DIFF
--- a/src/JMS/Serializer/Metadata/ClassMetadata.php
+++ b/src/JMS/Serializer/Metadata/ClassMetadata.php
@@ -56,8 +56,9 @@ class ClassMetadata extends MergeableClassMetadata
     public $discriminatorFieldName;
     public $discriminatorValue;
     public $discriminatorMap = array();
+    public $discriminatorGroups = array();
 
-    public function setDiscriminator($fieldName, array $map)
+    public function setDiscriminator($fieldName, array $map, array $groups = array())
     {
         if (empty($fieldName)) {
             throw new \InvalidArgumentException('The $fieldName cannot be empty.');
@@ -70,6 +71,7 @@ class ClassMetadata extends MergeableClassMetadata
         $this->discriminatorBaseClass = $this->name;
         $this->discriminatorFieldName = $fieldName;
         $this->discriminatorMap = $map;
+        $this->discriminatorGroups = $groups;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


maennchen:
@schmittjoh:
> would you mind re-basing the discriminator commit and adjusting the serialize/unserialize method so that it is BC by either throwing an exception or setting a default value?

I don't understand exactly what you want here. Could you explain further?

schmittjoh:
Just a check to handle unserializing an old version of the metadata class that didn't have the property yet. Right now, we will drop the parent metadata or have a notice error I guess.

src/JMS/Serializer/Metadata/ClassMetadata.php
```diff
@@ -251,6 +255,7 @@ public function unserialize($str)
             $this->discriminatorFieldName,
             $this->discriminatorValue,
             $this->discriminatorMap,
+            $this->discriminatorGroups,
             $parentStr
         ) = unserialize($str);
```
schmittjoh: ``$str`` might not have data for the new property yet.

maennchen:
Do we have to provide no BC breaks for Caches? I think that people have to clear caches between deployments.

schmittjoh:
I haven't looked into the consequence, but if we for example do not pick-up some exclusion rules anymore because of such a change and someone forgets to clear his cache, sensitive data might be exposed.

So, in general, yes, we should not break BC and anyway it's easy for us to add this check here.

wins1908:
Hello @schmittjoh , @maennchen 
Could we continue on this pull request?

maennchen:
@schmittjoh: 
I added a check for consistency.

I'm not really happy about the solution. (Or that we have to check caches in general.) What do you think about it?

src/JMS/Serializer/Metadata/ClassMetadata.php
```diff
@@ -34,6 +34,7 @@ class ClassMetadata extends MergeableClassMetadata
     const ACCESSOR_ORDER_UNDEFINED = 'undefined';
     const ACCESSOR_ORDER_ALPHABETICAL = 'alphabetical';
     const ACCESSOR_ORDER_CUSTOM = 'custom';
+    const COUNT_PARAMS = 16;
```
goetas: why do not use this approach https://github.com/schmittjoh/serializer/pull/301/files#diff-2d7dc6b1d453d596671e68f15d9a36d0R158 to have a backward compatibile serializtion ? and does not require all che "count" checks

maennchen: I do not want backward compatible serialization. People have to clear caches between deployments.

maennchen: In my opinion caches are not part of backwards compatibility. One can expect the users to clean caches.

goetas:
> I do not want backward compatible serialization. People have to clear caches between deployments.

I do not see nothing wrong with keeping compatibility, and the proposed solution requires less code and logic.

maennchen:
I think having compatibility to caches is nonsense.

I'm not going to contribute to OS projects with code that does not live up to the standards I set for myself.

If you want the method with some other deserialization handling you're going to have to commit it yourself.

PR #580, opened 2016-05-09
HHVM compatibility

goetas:
- reflection incompatibility between hhvm & php

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ php:
   - hhvm
 
 matrix:
-  allow_failures:
-    - php: hhvm
   fast_finish: true
   include:
     - php: 5.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,25 +2,33 @@ language: php
 
 sudo: false
 
+git:
+    depth: 1
+
 cache:
   directories:
     - $HOME/.composer/cache
 
-php:
-  - 5.5
-  - 5.6
-  - 7.0
-  - hhvm
-
 matrix:
-  fast_finish: true
-  include:
-    - php: 5.5
-      env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
+    include:
+        - php: hhvm
+          sudo: required
+          dist: trusty
+          group: edge
+        - php: hhvm-nightly
+          sudo: required
+          dist: trusty
+          group: edge
+        - php: 5.5
+        - php: 5.6
+        - php: 7.0
+        - php: 5.5
+          env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
+    fast_finish: true
 
 before_script:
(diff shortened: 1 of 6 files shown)


schmittjoh:
Please also check the Scrutinizer issue, we might need to adjust the exception it looks like.

stof:
tests are not green on HHVM

stof:
> allow to use symfony validator 2.2 and newer

Would be great to also allow the new Validator API available in Symfony 2.5+ (which is the only one in 3.0+)

goetas:
@schmittjoh issues on scrutinizer errors are because HHVM introduces a new public property on the reflection object

goetas:
@stof sorry for the mess, moved the validator feature to a separate PR https://github.com/schmittjoh/serializer/pull/583

goetas:
Currently making the serializer fully compatible with HHVM 3.6 is not easy.
With HHVM 3.7, tests are green. 

It looks that 3.6 has a buggy implementation of the `SplStack` class

schmittjoh:
@goetas, yes aware of that, but I was referring to the one about the exception. Did you check that, too?

schmittjoh:
3.6 seems quite outdated, I think we should not try to make that compatible if it has problems.

goetas:
@schmittjoh https://scrutinizer-ci.com/g/schmittjoh/serializer/inspections/481ad1b4-439f-4e33-88f8-0cfa960c1aab/issues/ i see only 3 new issues. can you specify ?

goetas:
> 3.6 seems quite outdated, I think we should not try to make that compatible if it has problems.

totally agree. but travis allows only HHVM 3.6. I'm not aware of a way to install a different version

schmittjoh:
I was referring to this one here: https://scrutinizer-ci.com/g/schmittjoh/serializer/inspections/f09be7fe-fd8f-4b2b-b833-45fd17ee7358/issues/files/src/JMS/Serializer/EventDispatcher/Subscriber/SymfonyValidatorSubscriber.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

Maybe we need to keep the allow_failures for now and wait for an image update. It's unfortunate, but working around bugs of outdated HHVM versions sounds not so good.

goetas:
Ok, then will keep this PR alive till travis do not update HHVM

goetas:
blocked buy https://github.com/facebook/hhvm/issues/7193

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: php
-
-sudo: false
-
+group: edge
+sudo: required
+dist: trusty
```
stof: this should be done only for the HHVM build rather than for PHP builds too, as the docker-based infra is faster to boot. See how it is done in Symfony

goetas: cool! will do

goetas:
Finally is green!

PR #584, opened 2016-05-10
Test on Travis always high and low deps

goetas:
- before this PR, dependencies of this project were uncertain. this tries to make more clarity on what are the minimum requirements for the serializer
- added code coverage on scrutinizer
- performance improvements

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ cache:
   directories:
     - $HOME/.composer/cache
 
+env:
+  - STABILITY=''
+  - STABILITY='--prefer-lowest'
+
 php:
   - 5.5
   - 5.6
@@ -18,7 +22,7 @@ matrix:
 
 before_script:
     - composer self-update
-    - composer install
+    - composer update $STABILITY
 
 script: vendor/bin/phpunit --coverage-clover clover
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,11 @@ matrix:
     - php: hhvm
 
 before_script:
+    - if [[ $TRAVIS_PHP_VERSION = '5.6' && $STABILITY = '' ]]; then PHPUNIT_FLAGS="--coverage-clover clover"; else PHPUNIT_FLAGS=""; fi
     - composer self-update
     - composer update $STABILITY
 
-script: vendor/bin/phpunit --coverage-clover clover
+script: vendor/bin/phpunit $PHPUNIT_FLAGS
 
 after_success:
     - wget https://scrutinizer-ci.com/ocular.phar
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,6 @@ cache:
   directories:
     - $HOME/.composer/cache
 
-env:
-  - STABILITY=''
-  - STABILITY='--prefer-lowest'
-
 php:
   - 5.5
   - 5.6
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -6,6 +6,10 @@ cache:
   directories:
     - $HOME/.composer/cache
 
+env:
+  - STABILITY=''
+  - STABILITY='--prefer-lowest'
```
stof: it would be much better to run a single job for prefer-lowest instead of one per PHP version, to limit the size of the build matrix. Travis allows only 5 concurrent jobs per repo owner (i.e. for all repos of @schmittjoh in this case) so the size of the matrix has an impact

composer.json
```diff
@@ -29,14 +30,14 @@
         "twig/twig": "~1.12|~2.0",
         "doctrine/orm": "~2.1",
         "doctrine/phpcr-odm": "^1.3",
-        "jackalope/jackalope-doctrine-dbal": "^1.0",
+        "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "propel/propel1": "~1.7",
-        "symfony/yaml": "2.*",
-        "symfony/translation": "~2.0",
+        "symfony/yaml": "^2.0.5",
```
stof: IIRC, 2.0.6 also had broken metadata, so the lowest dependency should be higher (I suggest making it at least 2.1, like other components)

composer.json
```diff
@@ -29,14 +30,14 @@
...
+        "symfony/translation": "^2.1",
         "symfony/validator": "^2.2",
         "symfony/form": "~2.1",
-        "symfony/filesystem": "2.*",
-        "phpunit/phpunit": "^4.0|^5.0"
+        "symfony/filesystem": "^2.1",
+        "phpunit/phpunit": "^5.0"
```
stof: you also need to allow phpunit 4.8, as PHPUnit 5 is not compatible with PHP 5.5 (or you should rely on the phar available on Travis)

goetas: thanks a lot, i was trying to find a solution exactly for that problem

schmittjoh:
I'm really not sure about this approach. Previously with the lock file, we have a stable build and if we update the lock, and something breaks we know which dependencies to look at.

Without the lock file, things might break without us changing the code. People will complain about tests already being broken when opening pull-requests and we end up hunting down where it comes from without knowing for which dependencies the tests worked. This highest lowest seems useful before a release, but also very fragile otherwise.

Can we combine both approaches (lock file for normal usage, highest-lowest before a release)?

.travis.yml
goetas: i tried to to optimize things to lower the build time and now is almost half time.
do you know some projects using the approach you suggested? to see exactly what do you mean

stof: see https://github.com/minkphp/Mink/blob/master/.travis.yml for instance

goetas:
>Without the lock file, things might break without us changing the code. 

personally i prefer to do not commit lock file, not committing it allows to understand if the declared dependencies and requirements allows to work correctly with the serializer. if some external dependency broke the serializer, the serializer should declare to be conflicting with that version.
committing the lock file ensure that the dependencies at that point in time were ok, but says nothing about the dependencies of the serializer.
A good point about this is https://github.com/schmittjoh/serializer/pull/578#discussion_r62361873 and https://github.com/composer/composer/issues/2461

stof:
@schmittjoh with the previous lock file, you were testing the library only against dependencies that none of your user would use as you were locking versions unmaintained since more than 2 years. so this was making the testsuite useless (as it would not ensure that things work in a real project) rather than making them fragile.

.travis.yml
goetas: done

PR #585, opened 2016-05-11
Ability to swap object to be serialized in pre-serialize event

jmcclell:
This is a new PR to separate the change originally in #542


stof:
you should create a branch containing only the commits related to Symfony 3 and opening a PR for that, separating this from the new feature (which needs more discussion as it is controversial)

jmcclell:
#542 is the branch with just the Symfony 3 support. This branch just happens to still have those commits but this PR is for the feature. If you guys decide you want it I can rebase the Symfony 3 commits out of it.

goetas:
@jmcclell Tests are red

PR #586, opened 2016-05-11
Fixed type-hinting

jgendera:


goetas:
What is the advantage? Is it for some IDE support? which one?

jgendera:
Yes. When you create serialization subscriber, when adding data via Visitor, PHPStorm displays warnings. i.e. "Expected \JMS\Serializer\scalar, got string"

Example code:
```
public function onPostSerialize(ObjectEvent $event)
{
    $object = $event->getObject();
    $visitor = $event->getVisitor();
    $visitor->addData('url', 'http://example.com/' . $object->getFileName());
}
```

goetas:
:+1:

PR #591, opened 2016-05-22
Assert type in string visitor

z38:
At the moment, calling `$serializer->deserialize('[1, 3, 4]', 'string', 'foo');` using a `GenericDeserializationVisitor` (e.g. JSON) throws a PHP notice. After applying this fix, a proper exception is thrown. The other primitive visitors do not seem to have the same problem, as they just default to `0`/`false` without throwing notices.

### DIFF
--- a/src/JMS/Serializer/GenericDeserializationVisitor.php
+++ b/src/JMS/Serializer/GenericDeserializationVisitor.php
@@ -58,6 +58,10 @@ abstract class GenericDeserializationVisitor extends AbstractVisitor
 
     public function visitString($data, array $type, Context $context)
     {
+        if (is_array($data)) {
+            throw new RuntimeException(sprintf('Expected a string, but got an array: %s', json_encode($data)));
+        }
+
         $data = (string) $data;
 
         if (null === $this->result) {
--- a/tests/JMS/Serializer/Tests/Serializer/JsonSerializationTest.php
+++ b/tests/JMS/Serializer/Tests/Serializer/JsonSerializationTest.php
@@ -209,6 +209,15 @@ class JsonSerializationTest extends BaseSerializationTest
         $this->assertEquals('{"0":{}}', $this->serialize(array(new \stdClass())));
     }
 
+    /**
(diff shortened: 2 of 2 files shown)


goetas:

Closing the PR. this is a type hinting problem, can not put everywhere in the code type guards.

z38:
No problem, thanks for getting back!

I'm not sure I understood you correctly. Would you mind showing me the preferred way to deal with this situation?

goetas:
If you expect a string, your input should be:

```
$this->deserialize('"[1, 3, 4]"', 'string');
```

If you expect an array, your input should be:

```
$this->deserialize('[1, 3, 4]', 'array');
```

The main idea is that you should always know what is the input format, not "guessing" one and reject it because or an exception

z38:
We currently use JMS serializer to deserialize the body of HTTP requests (very similar to [Request Body Listeners](http://symfony.com/doc/master/bundles/FOSRestBundle/request_body_converter_listener.html) in FOSRestBundle). These requests might be malformed.

At the moment any combination of input and type does either silently ignore/convert wrong types:
```
$serializer->deserialize('null', 'integer', 'json');
$serializer->deserialize('null', 'string', 'json');
$serializer->deserialize('"test"', 'integer', 'json');
$serializer->deserialize('5', 'string', 'json');
```
... or does throw an exception:
```
$serializer->deserialize('5', 'array', 'json');
$serializer->deserialize('"test"', 'array', 'json');
$serializer->deserialize('null', 'SomeClass', 'json');
```

```$serializer->deserialize('[]', 'string', 'json');``` seems to be the only combination which throws a PHP notice.

Would you recommend validating user input with a JSON schema before deserialization?

goetas:
re opening, just to keep track of the problem

goetas:
really not sure with this. if we do agree with this approach, then we should do the same with integers and booleans

goetas:
The conversion will raise a `E_NOTICE` that probably will be converted in exception by your framework.

PR #538, opened 2015-12-10
[new] implementing https://github.com/schmittjoh/serializer/issues/537

dbojdo:

PR #592, opened 2016-05-23
Introducing NormalizerInterface

alcalyn:
Reintegrates https://github.com/schmittjoh/serializer/pull/538.
Rebased to take updated tests, remove doc duplication.

The `NormalizerInterface` has the 2 methods `fromArray` and `toArray`.
It could be used instead of the full Serializer class when we need to use only `toArray`.

### DIFF
--- a/src/JMS/Serializer/NormalizerInterface.php
+++ b/src/JMS/Serializer/NormalizerInterface.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * Copyright 2013 Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace JMS\Serializer;
+
+/**
+ * Interface NormalizerInterface
+ * @author Daniel Bojdo <daniel@bojdo.eu>
+ */
+interface NormalizerInterface
+{
+    /**
+     * Converts objects to an array structure.
+     *
+     * This is useful when the data needs to be passed on to other methods which expect array data.
+     *
+     * @param mixed $data anything that converts to an array, typically an object or an array of objects
+     * @param SerializationContext $context
+     *
+     * @return array
+     */
+    public function toArray($data, SerializationContext $context = null);
+
+    /**
+     * Restores objects from an array structure.
+     *
+     * @param array $data
+     * @param string $type
+     * @param DeserializationContext $context
+     *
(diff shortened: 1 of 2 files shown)


goetas:
what is the use case for this? is somebody implementing a different serializer implementation? (even if is possible, i do not remember a single custom implementation.

but if is the case, the pr looks good

alcalyn:
I use it for dependency injection, instead of depending of full serializer, a service can depend on only this `NormalizerInterface`.

And it is easier to test it, by creating a mock of Normalizer with only the 2 methods it contains.

And there is already a `SerializerInterface`, so it seams that a `NormalizerInterface` has its place.

src/JMS/Serializer/NormalizerInterface.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+interface NormalizerInterface
+{
+    /**
+     * Converts objects to an array structure.
+     *
+     * This is useful when the data needs to be passed on to other methods which expect array data.
+     *
+     * @param mixed $data anything that converts to an array, typically an object or an array of objects
+     * @param SerializationContext $context
```
olvlvl: Should be `SerializationContext|null`

src/JMS/Serializer/NormalizerInterface.php
```diff
@@ -0,0 +1,49 @@
...
+     * @return array
+     */
+    public function toArray($data, SerializationContext $context = null);
+
+    /**
+     * Restores objects from an array structure.
+     *
+     * @param array $data
+     * @param string $type
+     * @param DeserializationContext $context
```
olvlvl: Should be `DeserializationContext|null`.

alcalyn: True, but sometimes I remember the Java logic that an object can be null, and then avoid the ugly `|null` for object, which is anyway obvious when reading the method declaration in that case

dbojdo:
@olvlvl Thanks for re-sync and fixes. I hope it will be merged soon.

scaytrase:
The PR is good but some proposals

1. Split current interface into `NormalizerInterface` and `DenormalizerInterface`
2. Gracefully rename methods to `normalize` and `denormalize` accordingly (leaving current ones in serializer as deprecated aliases for new one)

@goetas Use case is interoprability and dependency management. Currently the `toArray` method is not announced via `SerializerInterface`, so when using `SerializerInterface` as a dependency in class we have to do some dirty hacks like

```php
        if ($this->serializer instanceof Serializer) {
            return $this->serializer->toArray($object, $jmsContext);
        }
        return json_decode($this->serializer->serialize($object, 'json', $jmsContext), true);
```

Also interfaces make making adaptors and decoratos more easy. Non-announced methods are lost while decorating

alcalyn:
@scaytrase I had not the need to have two interfaces for each `toArray` and `fromArray` methods, moreother when I use `fromArray`, I often need its sister method, so that's why I choosen to make one interface and avoid to over-decouplate into many interfaces.

Also I see the PR is merged but only one commit has been merged, not the whole feature.

alcalyn:
Ok my bad, in fact the interface has been renamed `ArrayTransformerInterface` (see https://github.com/schmittjoh/serializer/commit/a0b7d10808faea32e36dc87883a032c2e0ebf0d8)

alcalyn:
https://github.com/schmittjoh/serializer/commit/a0b7d10808faea32e36dc87883a032c2e0ebf0d8

PR #595, opened 2016-05-26
Fix xml-attribute-map for the xml driver

romantomchak:
There is a simple test XML reference for class Foo:
```xml
<serializer>
    <class name="Foo">
        <property name="array" type="array" xml-attribute-map="true" />
    </class>
</serializer>
```

```php
class Foo
{
    private $array = [
        'name' => 'firstname',
        'value' => 'lastname'
    ];
}
```

When I try serialize this class I'm expecting to get the result as shown below but instead getting an error that the property **array** cannot be serialized because it contains not allowed characters.
```xml
<result name="firstname" value="lastname"/>
```

After investigation I found the mismatch between xml file property and property in metadata class.
This commit fixed this issue.


goetas:
:+1:

goetas:
Thank you

PR #598, opened 2016-05-27
Implemented MultiType annotation

0xPr0xy:
Created a MultiType annotation in order to be able to deserialize an array using strings as keys and multiple object types as values.
The use case for this annotation is to be compliant with the HAL json spec for the embedded resources.

Example: http://blog.stateless.co/post/13296666138/json-linking-with-hal
Spec: https://tools.ietf.org/html/draft-kelly-json-hal-06


goetas:
Closing. 
Sorry for the long feedback loop.
Not planning to integrate this feature as it is in the near future. If you think I should re-consider this, please provide more specific use cases and unit tests.

PR #603, opened 2016-06-17
[Visitor] Add data manipulation support

egeloen:
Hey!

This PR is mostly the same as #409 with in addition a methdod in order to replace completely the serialized data and it also adds tests for these new functionalities.

I have also refined the PHPUnit constraint in composer due to some warnings triggered related to `getMock` which has been deprecated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "symfony/validator": "^2.2",
         "symfony/form": "~2.1",
         "symfony/filesystem": "^2.1",
-        "phpunit/phpunit": "^4.8|^5.0"
+        "phpunit/phpunit": "^5.0,<5.4.0|^4.8"
     },
     "autoload": {
         "psr-0": {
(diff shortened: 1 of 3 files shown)


Gaditanian:
```php
public function onPreSerializeMethod(ObjectEvent $event) {

    $classMetadata = $event->getContext()->getMetadataFactory()->getMetadataForClass(get_class($event->getObject()));

    unset($classMetadata->propertyMetadata[$key]); // Remove 

  }
```

egeloen:
@Gaditanian Nice, I have not though about dropping the property metadata at pre serialize but we also use this PR for an other use case (replacing the serialized) data. We use it for handlin our relations between objects and inline an object:

```
public function onPostSerialize(ObjectEvent $event)
{
    $visitor = $event->getContext()->getVisitor();
    $data = $visitor->getData();

    if (is_array($data) && count($data) === 1 && (isset($data['id']) || isset($data['slug']))) {
        $visitor->setData(reset($data));
    }
}
```

I know it looks like an hack but I can't find an other way to do it... Do you have a nicer solution for that too? :)

goetas:
implemented in v1.4

PR #607, opened 2016-07-19
YAML mapping fails parseing if both handler_callbacks are defined

Deltachaos:
If you define two ``handler_callbacks`` in the YAML mapping the YamlDriver fails because the variable ``$direction`` from line 206 is overwritten in line 208 by ``GraphNavigator::parseDirection($direction)``.
This dont fails if you only define one item below ``handler_callbacks``, but if you define both it will fail with the error message: ``The direction "1" does not exist``


Deltachaos:
@schmittjoh the build failed but i think this is not because of this change. Seems to be a general problem currently.

goetas:
regarding the build https://github.com/schmittjoh/serializer/pull/608

mpajunen:
There's an open PR to fix this issue: #515

goetas:
Can you add a test?

mpajunen:
#515 also includes a test case already.

Deltachaos:
Yep #515 should be the one that is beeing merged.

PR #611, opened 2016-07-25
Allow to not skip empty not inline array root node

goetas:
The current implementation skips the root node for not inline arrays when serializing into XML.

Given: 

```php
class User {
    /**
     * @Serializer\XmlList(inline = false, entry = "comment")
     * @Serializer\Type("array")
     */
    public $data = [];
}
```

The previous output was: 

```xml
<result/>
```
The new behaviour is: 
```xml
<result>
  <data/>
</result>
```

~~Should this be considered a BC break?~~
probably yes

This change is consistent with the existing JSON and YML implementations since the output is

```json
{
  "data": []
}
```

```yml
data: []
```

### DIFF
--- a/src/JMS/Serializer/Annotation/XmlCollection.php
+++ b/src/JMS/Serializer/Annotation/XmlCollection.php
@@ -38,5 +38,5 @@ abstract class XmlCollection
     /**
      * @var boolean
      */
-    public $skip_when_empty = true;
+    public $skipWhenEmpty = true;
 }
--- a/src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
+++ b/src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
@@ -171,6 +171,7 @@ class AnnotationDriver implements DriverInterface
                         $propertyMetadata->xmlCollectionInline = $annot->inline;
                         $propertyMetadata->xmlEntryName = $annot->entry;
                         $propertyMetadata->xmlEntryNamespace = $annot->namespace;
+                        $propertyMetadata->xmlCollectionSkipWhenEmpty = $annot->skip_when_empty;
                     } elseif ($annot instanceof XmlMap) {
                         $propertyMetadata->xmlCollection = true;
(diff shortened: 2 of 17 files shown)


goetas:
> Should this be considered a BC break?

probably yes, will add an option to configure it

schmittjoh:
It seems very marginal, I would be more inclined to just merge it. It only breaks if you literally compare string values, no?

goetas:
Please wait, will fix it tomorrow. I have some usecases

goetas:
Added configuration optioin that allows to chose the default serialization strategy.

```php
class User {
    /**
     * @Serializer\XmlList(inline = false, entry = "comment", skip_when_empty=false)
     * @Serializer\Type("array")
     */
    public $data = [];
}
```

this will add the new behaviour

goetas:
The last two commits are minor changes: 
https://github.com/schmittjoh/serializer/pull/611/commits/2d664fa95c670451dd98046cdbb9624839ff2f92 
ia a minor changes that solves some edge cases when document contains namespaces.

 https://github.com/schmittjoh/serializer/pull/611/commits/38604116da214a125bcebe4224e931c89c79ff55 simplifies the if/else conditions and checks for namespaces difined at node scope, not only at document scope

src/JMS/Serializer/Annotation/XmlCollection.php
```diff
@@ -34,4 +34,9 @@
      * @var string
      */
     public $namespace;
+
+    /**
+     * @var boolean
+     */
+    public $skip_when_empty = true;
```
schmittjoh: Do we use this casing for other annotations? I think we already use camelCasing, no?

goetas: true, will fix it

goetas: done

goetas:
:tada:

goetas:
thanks! 

any plan for a `1.2.0` release? 

(there are some PR pretty safe to merge that can be good to include in a hypothetical 1.2 release)

PR #612, opened 2016-08-02
Change type hint for TypeParser on Serializer

mloberg:
It seems odd that Serializer takes in an argument for the type parser, but type hints it to TypeParser, which is a final class.

I suggest changing the type hint to AbstractParser (which TypeParser extends). This allows the user to provide their own type parser if they so choose.

### DIFF
--- a/src/JMS/Serializer/Serializer.php
+++ b/src/JMS/Serializer/Serializer.php
@@ -18,6 +18,7 @@
 
 namespace JMS\Serializer;
 
+use JMS\Parser\AbstractParser;
 use JMS\Serializer\Construction\ObjectConstructorInterface;
 use JMS\Serializer\Exception\RuntimeException;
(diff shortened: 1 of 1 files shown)


goetas:
Unfortunately this will not be enough.
On https://github.com/schmittjoh/serializer/blob/b359a1bf7cd05f14be373d4aa5a8e3c26a9c531d/src/JMS/Serializer/Metadata/PropertyMetadata.php#L107 there is another creation of the type parser.

The best solution is to have a `TypeParser` injected everywhere by the `SerializerBuilder` (metadata drivers included).

goetas:
Moving the problem to 2.0

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/743/commits/90389918fff2f18146a0deae89da4005fdf33504

PR #619, opened 2016-08-09
Deserialize exception

ivan1986:
#447

rebase

If deserialize from json/xml with wrong structure exception hasn't info about path where structure is wrong - it's very hard to debug

this add new exception, contains path as "comments[0].author.full_name" to unexpected value

### DIFF
--- a/src/JMS/Serializer/Metadata/IndexMetadata.php
+++ b/src/JMS/Serializer/Metadata/IndexMetadata.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * Copyright 2015 Ivan Borzenkov <ivan.borzenkov@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace JMS\Serializer\Metadata;
+
+/**
+ * Index Metadata used for saving array index in Context metadataStack
+ *
+ * @author Ivan Borzenkov <ivan.borzenkov@gmail.com>
+ */
+class IndexMetadata implements \Serializable
+{
+    public $index;
+
+    /**
+     * IndexMetadata constructor.
+     *
+     * @param $index
+     */
+    public function __construct($index)
+    {
+        $this->index = $index;
+    }
+
+    public function getValue()
+    {
+        return $this->index;
+    }
+
+    public function serialize()
+    {
+        return serialize(array(
+            $this->index,
+        ));
+    }
+
+    public function unserialize($str)
+    {
+        list(
+            $this->index,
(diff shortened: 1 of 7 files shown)


tests/JMS/Serializer/Tests/Serializer/DeserializeExceptionTest.php
```diff
@@ -0,0 +1,218 @@
...
+use JMS\Serializer\Naming\SerializedNameAnnotationStrategy;
+use JMS\Serializer\Serializer;
+use JMS\Serializer\Tests\Fixtures\AuthorList;
+use JMS\Serializer\VisitorInterface;
+use Metadata\MetadataFactory;
+use PhpCollection\Map;
+use Symfony\Component\Translation\IdentityTranslator;
+use Symfony\Component\Translation\MessageSelector;
+
+class DeserializeExceptionTest extends \PHPUnit_Framework_TestCase
```
goetas: why not add this tests in `BaseSerializationTest` to avoid all that stuff in `setUp` ?

src/JMS/Serializer/Metadata/IndexMetadata.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+namespace JMS\Serializer\Metadata;
+
+/**
+ * Index Metadata used for saving array index in Context metadataStack
+ *
+ * @author Ivan Borzenkov <ivan.borzenkov@gmail.com>
+ */
+class IndexMetadata implements \Serializable
```
goetas: why it has to be `Serializable` ?

goetas:
In general I'm not happy with the added complexity, compared to the result

src/JMS/Serializer/Metadata/IndexMetadata.php
ivan1986: as all metadata classes in jms/metadata

goetas: but `IndexMetadata` is newer persisted, is used by you only to track the loop index

the jms metadata classes represents class metadata info that are cached on disk

tests/JMS/Serializer/Tests/Serializer/DeserializeExceptionTest.php
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        if (!isset($outputs[$key])) {
+            throw new RuntimeException(sprintf('The key "%s" is not supported.', $key));
+        }
+
+        return $outputs[$key];
+    }
+
+    protected function getFormat()
```
goetas: what about xml and yml?

tests/JMS/Serializer/Tests/Serializer/DeserializeExceptionTest.php
```diff
@@ -0,0 +1,146 @@
...
+
+use Doctrine\Common\Annotations\AnnotationReader;
+use JMS\Serializer\Context;
+use JMS\Serializer\Exception\RuntimeException;
+use Metadata\MetadataFactory;
+use PhpCollection\Map;
+use Symfony\Component\Translation\IdentityTranslator;
+use Symfony\Component\Translation\MessageSelector;
+
+class DeserializeExceptionTest extends BaseSerializationTest
```
goetas: actually my idea was to add this tests directly into `BaseSerializationTest` and calling them something as `test(notBoolean|notString...)ThrowsException`

goetas:
At the moment I'm not planning to add this feature to the code. It adds complexity and slow down the serialization process (adding the index metadata) just to have some debug info in case of errors

PR #623, opened 2016-08-11
Removed unused imports

olvlvl:
Hi,

This PR is removing some unused imports left behind.


goetas:
Postponed to avoid conflicts with other PR since this does not add any real improvement

PR #626, opened 2016-08-17
Fix serializing nulls in arrays

megazoll:
`null`s in the arrays should not be ignored when serializeNull is `true`.

### DIFF
--- a/src/JMS/Serializer/Context.php
+++ b/src/JMS/Serializer/Context.php
@@ -54,6 +54,9 @@ abstract class Context
     /** @var boolean|null */
     private $serializeNull;
 
+    /** @var boolean */
+    private $serializeInArrayNull = false;
+
     private $initialized = false;
 
     /** @var \SplStack */
@@ -208,6 +211,35 @@ abstract class Context
     public function shouldSerializeNull()
     {
         return $this->serializeNull;
+    }
+
+    /**
+     * Set if NULLs in arrays should be presented (TRUE) or skipped (FALSE)
+     *
+     * @param bool $bool
+     * @return $this
+     */
+    public function setSerializeInArrayNull($bool)
+    {
+        $this->serializeInArrayNull = (boolean) $bool;
+
+        return $this;
+    }
+
+    /**
+     * Returns TRUE when NULLs in arrays should be presented
+     * Returns FALSE when NULLs in arrays should be skipped
+     *
+     * @return bool
+     */
+    public function shouldSerializeInArrayNull()
+    {
+        return $this->serializeInArrayNull;
+    }
+
+    public function shouldSerializeNullForKey($key)
+    {
+        return is_string($key) ? $this->shouldSerializeNull() : $this->shouldSerializeInArrayNull();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


megazoll:
Also fixed yaml serialization for integrity and added some tests for serializing nulls

goetas:
Merging this as it is is not possible. As it is, is a BC break since the array serialization behavior changes between versions.

What about adding an extra `shouldSerializeInArrayNull` ? this will make it BC compatible.

Other suggestions?

Otherwise will have to postpone it to 2.0

scaytrase:
Not sure that fixing bug should be considered a BC break

goetas:
> Not sure that fixing bug should be considered a BC break

valid point....

goetas:
But there will be a lot of broken code around if this gets merged as it is

scaytrase:
The behaviour is inconsistent now. We have

```php
$array = [0, null, 'a'];
$context = SerializationContext::create()->setSerializeNull(true);
```
<table>
  <tr><th>Expression</th><th>var_dump result</th></tr>
  <tr><td>$array</td><td><pre>
array(3) {
  [0] =>
  int(0)
  [1] =>
  NULL
  [2] =>
  string(1) "a"
}</pre></tr>
  <tr><td>$serializer->toArray($array)</td><td><pre>array(2) {
  [0] =>
  int(0)
  [1] =>
  string(1) "a"
}</pre></td></tr>
  <tr><td>json_encode($array)</td><td><pre>string(12) "[0,null,"a"]"</pre></td></tr>
  <tr><td>serializer->serialize($array, 'json', $context)</td><td><pre>string(12) "[0,null,"a"]"</pre></td></tr>
  <tr><td>Yaml::dump($array)</td><td><pre>string(15) "- 0
- null
- a
"</pre></td></tr>
  <tr><td>serializer->serialize($array, 'yml', clone $context)</td><td><pre>string(8) "- 0
- a
"
</pre></td></tr>

</table>

So we have json visitor behaving correct for plain array and other visitors are not.

Having the fix will make the serializer more consistent across formats

goetas:
@megazoll  what do you think?

goetas:
@scaytrase I  agree with you, but Im just worried for all the users that upgrading from 1.3 to 1.4 get their code to be broken.

scaytrase:
@goetas updated the table. Currently json makes the array to be an object, yaml does  not (simply ignores null as toArray).

I think we should ask the community, i.e maintainers of some dependent libraries (FOS Rest, Sonata Admin) 

https://www.versioneye.com/php/jms:serializer/references
https://www.versioneye.com/php/jms:serializer-bundle/references

scaytrase:
IMO
It's not a BC break to have a major version. But we can aggregate more changes in a separate branch, leaving this one stable behind.  Focusing on a separate major (and not upgrading this one) release we can introduce more usefull changes alongside the improving consistency.

goetas:
At the moment moving it to 1.5

Anybody interested in implementing  something as https://github.com/schmittjoh/serializer/pull/626#issuecomment-245957111 ?

megazoll:
@goetas updated. Please review my changes.

goetas:
@megazoll Thanks for your effort, I have realized that this was an old standing bug (in XML was totally not working) and decided to fix it with https://github.com/schmittjoh/serializer/pull/660 (as said is almost identical to your first version.

Thanks again for the contribution and sorry if it took so much time and uncertainty.

PR #628, opened 2016-08-19
Replace deprecated getMock with createMock

megazoll:
Also require `phpunit: ^5.4` and remove support php 5.5 becase phpunit 5 requires php 5.6 and php 5.5 not supported by php team: http://php.net/supported-versions.php

### DIFF
--- a/tests/JMS/Serializer/Tests/BaseTestCase.php
+++ b/tests/JMS/Serializer/Tests/BaseTestCase.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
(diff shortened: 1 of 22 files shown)


.travis.yml
```diff
@@ -19,10 +19,9 @@ matrix:
           sudo: required
           dist: trusty
           group: edge
-        - php: 5.5
```
goetas: we want to support php 5.5 beside of phpunit

megazoll:
Added BaseTestCase to keep compatibility with php 5.5

Issue #500, opened 2015-09-28
MaxDepth exclusion strategy for OneToMany (ArrayCollection) type triggers too many doctrine queries.


vishalmelmatti:
Hi 

I have Node class which has self parent-child  relationship.

```
class Node {

    public $id
    public $title;

/**
 * @ORM\OneToMany(targetEntity="Node", mappedBy="parent")
 * @Expose
 * @MaxDepth(2)
  */
    public $children;
```

}

This should display Top level Node + its children. The problem is, while displaying children properties for each property which is exposed, getter of that property is called. 
So for each children, "getChildren" is also called which triggers separate doctrine query for each child. 

As we only want to display up to level 2  MaxDepth(2),  childrent of children won't be displayed which is fine but still queries for each child gets triggers as their getter is called which impact performance tremendously with more number of children.

So getChildren getter should not be called at MaxDepth(2) level. For that, DepthExclusionStrategy needs to handle it.

I tried doing below changes, it works for me. Please let me know if there is any better way to do it and it would be great to have it fixed in this serializer repository.

```
private function isTooDeep(Context $context, $property = null)
{

    $depth = $context->getDepth();
    $metadataStack = $context->getMetadataStack();

    $nthProperty = 0;
    // iterate from the first added items to the lasts
    for ($i = $metadataStack->count() - 1; $i > 0; $i--) {
        $metadata = $metadataStack[$i];
        if ($metadata instanceof PropertyMetadata) {
            $nthProperty++;
            $relativeDepth = $depth - $nthProperty;

            if ($property && $property->type['name'] == "ArrayCollection") {
             if (null !== $metadata->maxDepth && $relativeDepth >= $metadata->maxDepth) {
                return true;
             } 
            } else {
             if (null !== $metadata->maxDepth && $relativeDepth > $metadata->maxDepth) {
                return true;
             } 
           }
        }
    }

    return false;
}
```

oshanz:
@vishalmelmatti i had same issue.
its not the best, because it can be any other type (T / array<T>). but temporary this solved my case thanks.

akomm:
Just encountered the same Issue.

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/760 and https://github.com/schmittjoh/serializer/commit/136fb0d6da43bc255aa8658414a724acb889f3c1

Issue #411, opened 2015-03-09
Serializing ArrayCollection of an abstract class returns empty objects


halloverden:
I'm using the [Serlalizer Bundle](https://github.com/schmittjoh/JMSSerializerBundle) with Symfony2, but suspect this is a Serializer issue. 

I have an abstract class "Question":

```
/**
 * App\SurveyBundle\Entity\Question
 * 
 * @ORM\Entity
 * @ORM\InheritanceType("JOINED")
 * @ORM\DiscriminatorColumn(name="type", type="integer")
 * @ORM\DiscriminatorMap({0 = "App\SurveyBundle\Entity\Question", 3 = "App\SurveyBundle\Entity\MultipleChoice", 2 = "App\SurveyBundle\Entity\SingleChoice", 1 = "App\SurveyBundle\Entity\LikertScale"})
 * @JMS\Discriminator(field="type", map={0: "App\SurveyBundle\Entity\Question", 3: "App\SurveyBundle\Entity\MultipleChoice", 2: "App\SurveyBundle\Entity\SingleChoice", 1: "App\SurveyBundle\Entity\LikertScale"})
 * @ORM\HasLifeCycleCallbacks
 */
abstract class Question {
    /**
     * Question type
     */
    public function getQuestionType() {
        if ($this instanceof MultipleChoice) {
            return 3;
        } else if($this instanceof SingleChoice) {
            return 2;
        } else if($this instanceof LikertScale) {
            return 1;
        } else {
            return 0;
        }
    } 
}
```

...extended by three sub classes, using this yml annotation file (Example for the "SingleChoice" extended class).

```
App\SurveyBundle\Entity\SingleChoice:
    exclusion_policy: ALL
    properties:
        id:
            expose: true
            type: integer
            groups: [default]
        questionText:
            expose: true
            serialized_name: questionText
            type: string
            groups: [default]
        created:
            expose: true
            type: DateTime
            groups: [default]
        survey:
            expose: true
            type: App\SurveyBundle\Entity\Survey
            groups: [deserialization]
    virtual_properties:
        getQuestionType:
            serialized_name: type
            type: integer
            groups: [default]
```

The abstract class "Question" is used by my "Survey" class, which holds an ArrayCollection of Questions.

```
App\SurveyBundle\Entity\Survey:
    exclusion_policy: ALL
    properties:
        id:
            expose: true
            type: integer
            groups: [default]
        created:
            expose: true
            type: DateTime
            groups: [default]
        name:
[...]

Issue #407, opened 2015-03-01
Serialize doctrine entities load linked entities even if the max depth should stop this behavior


nicodmf:
When serialize doctrine entity, if exclusion is made :
- by max_depth : the linked entities are loaded, the field appears but it is empty if array and doesn't appear if single
- by groups : the linked entities aren't loaded and the field doesn't appear

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/743/commits/94c319db88883757a14ac267115272be030aef28  and https://github.com/schmittjoh/serializer/commit/136fb0d6da43bc255aa8658414a724acb889f3c1

PR #632, opened 2016-08-25
MaxDepth is handled incorrectly with Collections

voda:
This is an updated version of PR #423 with included tests. Original description:

This fixes handling collections for the purpose of depth calculations. Since arrays are not counted as a depth level, neither should Collections.
The other side effect of this change is that @MaxDepth on an entity inside a collection will be handled correctly as well (before, each Collection contributed +1 to depth, but did not leave PropertyMetadata, therefore was not subtracted when calculating relative depth).

fixeds https://github.com/schmittjoh/serializer/issues/148 https://github.com/schmittjoh/serializer/issues/500 https://github.com/schmittjoh/serializer/issues/702 https://github.com/schmittjoh/serializer/issues/272 https://github.com/schmittjoh/serializer/issues/411 https://github.com/schmittjoh/serializer/issues/407

### DIFF
--- a/src/JMS/Serializer/Handler/ArrayCollectionHandler.php
+++ b/src/JMS/Serializer/Handler/ArrayCollectionHandler.php
@@ -19,8 +19,9 @@
 namespace JMS\Serializer\Handler;
 
 use Doctrine\Common\Collections\ArrayCollection;
-use JMS\Serializer\Context;
+use JMS\Serializer\DeserializationContext;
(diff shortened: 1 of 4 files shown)


goetas:
I also found this behavior a bit confusing, but at the end results to be consistent. I will be fine keeping it as it is

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/743/commits/136fb0d6da43bc255aa8658414a724acb889f3c1

PR #635, opened 2016-08-29
Issue543 - Adding DateTimeImmutable support

toby-griffiths:
Duplicated DateTime related tests & amended to DateTimeImmutable.

All 'datetime' group tests passing with this update.


Einenlum:
:+1:

PR #642, opened 2016-09-08
Simple test for plain arrays serialization

### LINKED ISSUE #641: Plain arrays are serialized (normalized) as "objects", ignoring serializeNull
As soon as I have no response in original closed issue #571 , I've open a new one

Given commit https://github.com/schmittjoh/serializer/commit/b0a46c80cf3dbd97205f2e2ffeb9229080899c76 does not solve the problem - plain arrays are still serialized as object:

`test/JMS/Serializer/Tests/Serializer/ArrayTest.php`

```php
    public function testNullArraySerialization()
    {
        $context = SerializationContext::create();
        $array = [0, null, 'a'];
        self::assertSame($array, $this->serializer->toArray($array, $context));
    }
```

```
Failed asserting that Array &0 (
    0 => 0
    2 => 'a'
) is identical to Array &0 (
    0 => 0
    1 => null
    2 => 'a'
).
```

How we force an array being normalized as array, not object? I'll try to file this failing test as pull request if you need it.

As a solution I propose to make this check 
https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/GenericSerializationVisitor.php#L101
look like others 
https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/YamlSerializationVisitor.php#L84-L85

scaytrase:
Test and fix sample

Add keys range check and remove string key requirement (numeric keys are ok for php plain arrays)

* Fix #641

### DIFF
--- a/tests/JMS/Serializer/Tests/Serializer/ArrayTest.php
+++ b/tests/JMS/Serializer/Tests/Serializer/ArrayTest.php
@@ -19,6 +19,7 @@
 namespace JMS\Serializer\Tests\Serializer;
 
 use JMS\Serializer\Handler\HandlerRegistry;
+use JMS\Serializer\SerializationContext;
 use JMS\Serializer\Tests\Fixtures\Author;
 use JMS\Serializer\Tests\Fixtures\AuthorList;
 use JMS\Serializer\Tests\Fixtures\Order;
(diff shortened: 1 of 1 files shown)


goetas:
duplicate of https://github.com/schmittjoh/serializer/pull/626 ?

scaytrase:
A bit. #626 Proposed more complete, but I have additional simple test for plain arrays. I can remove fix and propose test only

scaytrase:
@goetas updated, now it is compatible with #626

goetas:
what about the failing test?

scaytrase:
It would be fixed with #626, I think. That's the demonstration of the bug that currently present

goetas:
aaa, i see

PR #644, opened 2016-09-09
Xml namespaces improvements

goetas:
This PR improves the XML namespace prefix lookup by adding the node to the DOM tree as soon as possible.

Before:

``` xml
<a:root xmlns:a="a">
   <b:node xmlns:b="b">
         <a:subnode xmlns:a="a">
   </b:node>
</a:root>
```

after 

``` xml
<a:root xmlns:a="a">
   <b:node xmlns:b="b">
         <a:subnode>
   </b:node>
</a:root>
```

The new behavior produces a cleaner tree since the namespace "a:subnode"   was already defined on the root, and re-defining it on the node is redundant.


goetas:
This can be considered a partial BC break since many test suite that do xml comparison will fail.

the xml semantic will not change at all, will change just a bit the syntax.

Unfortunately xml parsers are not really semantic :(

planning to move it to  2.0

PR #382, opened 2015-01-06
Fix inheritance of discriminators on Doctrine entities

xoob:
The discriminator configuration in the class metadata of Doctrine entities breaks in cases where the base-entity extends another non-entity class. This PR adds a testcases for that problem as well as a fix.

Fixes the following issues:

  - #166 - _Discriminator related metadata overwritten by subsequent MetadataDriver in DriverChain_
  - #182 - _Deserialization fails when discriminator base class extends another class_
  - #296 - _Overwrite discriminator settings in metadata if not present in parent class_
  - #309 - _add / discriminator for classes in the middle of hierarchy_

Working
-------

Entity inherits from abstract entity.

```
abstract class Organization                         @ORM\Entity  @ORM\InheritanceType("SINGLE_TABLE")  @ORM\DiscriminatorMap(...)
 `- class School extends Organization               @ORM\Entity
```

```php
<?php
  $school = new School();
  $json = $this->serializer->serialize($school, 'json');
?>

--- Expected
+++ Actual
@@ @@
-'{"type":"school"}'
+'{"type":"school"}'
```

Broken
------

Entity inherits from abstract entity inherits from abstract non-entity.

```
abstract class AbstractModel                        (none)
 |- abstract class Person extends AbstractModel     @ORM\Entity  @ORM\InheritanceType("SINGLE_TABLE")  @ORM\DiscriminatorMap(...)
 |   |- class Student extends Person                @ORM\Entity
 |   `- class Teacher extends Person                @ORM\Entity
 `- class Clazz extends AbstractModel               @ORM\Entity
```

```php
<?php
  $student = new Student();
  $json = $this->serializer->serialize($student, 'json');
?>

--- Expected
+++ Actual
@@ @@
-'{"type":"student"}'
+'{}'
```

PR #648, opened 2016-09-12
fix discriminator for classes in the middle of hierarchy

scasei:


goetas:
Something similar should be already on master.

What is this PR for?

scasei:
In my application I have class hierarchy like this:

````php

class A {}

abstract class B extends A {}

class BA extends B {}
class BB extends B {}

````

When I serialize an instance of class BA or BB there is a column 'type' in the json, describing which sub class is serialized. 
When it comes to deserialization the serializer tries to instantiate class B instead of BA or BB.
The PR points the serializer to BA/BB so there is no error like: 'The provided class "B" is abstract, and can not be instantiated'

**Comment by goetas** [comment]
Isn't related to https://github.com/schmittjoh/serializer/pull/382 ?

**Comment by scasei** [comment]
Today I updated serializer to tag 1.3.1 but the errors still occurs.

**Comment by scasei** [comment]
Should I try dev-master?

**Comment by goetas** [comment]
Yes please, was merged Friday

**Comment by scasei** [comment]
Ok, tested with 5ebae63 (dev-master), but problem still exists.

Can you please consider a merge?

**Comment by goetas** [comment]
Can you provide tests?

**Comment by scasei** [comment]
No test, no merge, right?

**Comment by goetas** [comment]
Yep

**Comment by goetas** [comment]
But for me is sospicous that master dev does not work. ..

**Comment by scasei** [comment]
Ok, I'll double check my setup

**Comment by goetas** [comment]
You can check the PR, I think is exactly the feature you need

**Comment by goetas** [comment]
(Remember to clear caches)

**Comment by scasei** [comment]
I'm with you, I check everything again

**Comment by scasei** [comment]
I saw, that I requested the same here: #309  :)

And found that my deserialization call is like the following:

````php
$jms_serializer->deserialize($json_content,'WebCollection','json','AbstractMiddleClassName');
````

So the serializer has to find the right sub class from the 'type'' column in the json content.
The premise of the already merged fixes is, that you pass the right sub class name to deserialization.
Thats why it needs this addional code for my setup.

**Comment by goetas** [comment]
Ah, I see. Your pr is for the deserialization case.

Unfortunately without tests can't be merged

**Comment by scasei** [comment]
aight, will push the test here

**Comment by scasei** [comment]
Update: made a fresh clone of masteer and composer install, then run phpunit and got lots of errors, especially for my wanted feature (testNestedPolymorphicInterfaces).

Win10 / PHP 5.6.25 nts

Is there something I should investigate?

**Comment by goetas** [comment]
>  then run phpunit and got lots of errors, 

what it means? can you share?

**Comment by scasei** [comment]
here is my console output:

````console
MINGW64 /c/Program Files (x86)/wamp/www/serializer (master)
$ phpunit.phar
PHPUnit 5.5.4 by Sebastian Bergmann and contributors.

WWWWWWWW............................WWWWWWWWWWW...............  62 / 399 ( 15%)
...................................F..E..........WW......EEEWW 124 / 399 ( 31%)
WWWWWWWWWWW....WWWWWWWW....................................... 186 / 399 ( 46%)
..........WW............................................FF...F 248 / 399 ( 62%)
FFFFFF.SFFFFFFFFFFFFFFFFFFFFFFFSFFFFFFFFFFFFFFFFFFFFFFFFWFFFFF 310 / 399 ( 77%)
FFFF.FF.FFFFFF.F.FSSSSSFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF 372 / 399 ( 93%)
FFFFFFFFFFFF.FF.FFFFFF.F.FW                                    399 / 399 (100%)




````

**Comment by goetas** [comment]
no idea... travis says master it is green https://travis-ci.org/schmittjoh/serializer/builds/159323741

**Comment by scasei** [comment]
Should I attach the whole output log?

**Comment by goetas** [comment]
ok

**Comment by scasei** [comment]
[output.txt](https://github.com/schmittjoh/serializer/files/469312/output.txt)

**Comment by goetas** [comment]
there are a bunch of errors in your environment... as example: 

`Exception: DateTime::__construct(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone.
```
date time not configured well

```
Doctrine\ORM\Tools\ToolsException: Schema-Tool failed with Error 'An exception occured in driver: could not find driver'`
```
sqlite missing?

im not use your PHP is well configured

scasei:
right, have fixed these errors already. the others could be line ending issues.

goetas:
windows :(

scasei:
So, after correcting the line endings, tests running well.


And thank you for pointing to coding a test, because my problem becomes clearly now:

My input is a collection of classes. 
The difference to #382 is, that #382 uses a parent class 'Garage' to store the information, which type the input is of.
In my use case I have a deserilization context which get the information about the type of the classes at runtime.

````php
class Vehicle {
	private $km;
}

class Car extends Vehicle {}
class Moped extends Vehicle {}


$json = '[{"km":3,"type":"car"},{"km":1,"type":"moped"}]';

$context = new \Own\Serializer\Context\DeserializationContext();
$context->setInnerObject('\Vehicle');

$ret = $this->jms_serializer->deserialize($json,'array','json',$context);
		
````

Questions:
Do you think I should work out the PR or should I trash everything and use static collection class as in #382?

**Comment by goetas** [comment]
`setInnerObject` is something new to me?

what about 
```php
$ret = $this->jms_serializer->deserialize($json,'array<Vehicle>','json');
```

`Vehicle` is abstract right?

goetas:
I suggest you to use simply http://jmsyst.com/libs/serializer/master/reference/annotations#discriminator

goetas:
and anyway, I will avoid to use inheritance if possible

scasei:
here my use case classes, i try if 'array<Vehicle>'  will work

````php
class DeserializationContext extends \JMS\Serializer\DeserializationContext
{

	/**
	 * @var string
	 */
	protected $inner_object;
	
	
	/**
	 * @param string $inner_object
	 */
	public function setInnerObject ($inner_object) {
		
		$this->inner_object = $inner_object;
	}
	
	/**
	 * @return string
	 */
	public function getInnerObject () {
		
		return $this->inner_object;
	}
}

namespace Prima\REST\ClientBundle\Serializer\Handler;

use JMS\Serializer\Handler\SubscribingHandlerInterface;
use JMS\Serializer\GraphNavigator;
use JMS\Serializer\VisitorInterface;
use Prima\REST\ClientBundle\Serializer\Context\DeserializationContext as Context;
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ORM\Mapping\ClassMetadata;

class WebCollectionHandler implements SubscribingHandlerInterface
{
	{
		return array(
				array(
						'direction' => GraphNavigator::DIRECTION_DESERIALIZATION,
						'format' => 'json',
						'type' => 'WebCollection',
						'method' => 'deserializeWebCollection',
				),
		);
	}

	public function deserializeWebCollection(VisitorInterface $visitor, $data, array $type, Context $context)
    {
		
 		$ret = new ArrayCollection();
 		$type['name'] = $context->getInnerObject();
 		 		
 		$key = key($data);
 		if (is_numeric($key)) {
 			
 			foreach ($data as $object) { 			
	 			$ret->add(
	 					$visitor->getNavigator()->accept($object, $type, $context)
	 			);
 			} 			
 		}
 		else {
 			
 			$ret->add(
 					$visitor->getNavigator()->accept($data, $type, $context)
 			);
 		}
 		
 		$refl = new \ReflectionObject($visitor); 		
 		$property = $refl->getParentClass()->getProperty('result');
 		$property->setAccessible(TRUE);
 		$property->setValue($visitor, $ret);
 		
 		
        return $ret;
    }
}
````

**Comment by goetas** [comment]
doing this you are moving the "deserialization" logic into something else... 

Do you use `@discriminator` in the serialization phase? why not use the same in  "deserialization" ?

**Comment by scasei** [comment]
That is my doctrine config file for the classes:
````yml
Prima\Partner\AccountBundle\Entity\User:
    type: entity
    table: partner_account
    inheritanceType: SINGLE_TABLE
    discriminatorColumn:
        name: type
        type: string
    discriminatorMap:
        partner: UserChief
        sub: UserSub

````

**Comment by goetas** [comment]
looking at doctrine documentation. 
> If no discriminator map is provided, then the map is generated automatically. The automatically generated discriminator map contains the lowercase short name of each class as key.

This means that for deserialization, doctrine does some magic, that you have to do on your own later...
But as you can see http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/inheritance-mapping.html#single-table-inheritance doctrine in the example prefers to use explicitly `@DiscriminatorMap`, and I suggest you to do the same.

Doing it, also JMS will work better.

**Comment by scasei** [comment]
> why not use the same in "deserialization" ?
My input has a type column, because of correct serialization with `@discriminator`. But in the deserialization phase the right subclass is not found,

**Comment by scasei** [comment]
> I suggest you to do the same. 

But I use a descriminator column :)

**Comment by goetas** [comment]
as you can see in the example, they uses both. 

```
@DiscriminatorColumn(name="discr", type="string")
@DiscriminatorMap({"person" = "Person", "employee" = "Employee"})
```

**Comment by goetas** [comment]
missed 
```
    discriminatorMap:
        partner: UserChief
        sub: UserSub
```
this means that you can copy & paste 
```
    discriminatorMap:
        partner: UserChief
        sub: UserSub
```
into
```
# Vendor\MyBundle\Resources\config\serializer\Model.ClassName.yml
Vendor\MyBundle\Model\ClassName:
    discriminator:
        field_name: type
        map:
           partner: UserChief
           sub: UserSub
```

scasei:
Ok, expect another abstract class 'FOS\UserBundle\Model\User' as parent of abstract 'Prima\Partner\AccountBundle\Entity\User' .

Then deserialization is definitely broken.
I can provide tests.

scasei:
Pushed test case, would please review?

goetas:
it looks you have used tabs instead of spaces

scasei:
Your right, I run cs fixer:

````bash
$ php-cs-fixer.phar fix ./ --fixers=linefeed,short_tag,indentation
   1) src\JMS\Serializer\Metadata\ClassMetadata.php
   2) tests\JMS\Serializer\Tests\Fixtures\DiscriminatorMiddle\Base.php
   3) tests\JMS\Serializer\Tests\Serializer\BaseSerializationTest.php
Fixed all files in 88.984 seconds, 7.750 MB memory used
````

scasei:
Can I do something else to get merged?

coyl:
+1. The error was not fixed in #382

scasei:
@goetas Is something left unclear, why this patch is useful?

coyl:
@goetas we really need this fix. I'm tired to live in forks =)

coyl:
UP

coyl:
=(

scasei:
@goetas some news?

goetas:
Replaced by https://github.com/schmittjoh/serializer/pull/659 (the test classes were already in the repo, no need to add 120 lines of test classes)

scasei:
thank you!

PR #651, opened 2016-09-20
Boolean visitor

AyrtonRicardo:
Problem: 
- I'm using the serializer in an API, when I receive a request with the boolean fields as "false" or "true" both are `true`, but "false" isn't `true` ( :smile: ). 

as workaround I used `filter_var` in all boolean fields before serialization.

Solution:
- `filter_var` in boolean visitor.


goetas:
Do not understand this PR, shouldn't `'false' === $data` do exactly the same?

AyrtonRicardo:
@goetas yep! `'false' === $data` are exactly the same, the problem was in serialization context, when `$data ? 'true' : 'false'`  "false" as string returns `true` in that context.

I changed the logic in `XmlDeserializationVisitor.php` only because `filter_var` do the same.

goetas:
in `visitBoolean` `$data` has to be a boolean, not as string containing "false"

PR #657, opened 2016-10-21
Handle both int and integer

Aliance:
Ofter required to write `int` not `integer`, because of PHP (habbit). Now serializer thinks that `int` is a kind of an object and warn about it:
> Expected object but got integer. Do you have the wrong @ Type mapping or could this be a Doctrine many-to-many relation?

### DIFF
--- a/src/JMS/Serializer/GraphNavigator.php
+++ b/src/JMS/Serializer/GraphNavigator.php
@@ -115,6 +115,7 @@ final class GraphNavigator
             case 'string':
                 return $visitor->visitString($data, $type, $context);
 
+            case 'int':
             case 'integer':
                 return $visitor->visitInteger($data, $type, $context);
 


goetas:
php allows `int` and `integer` too... seem not harmful

Aliance:
@goetas yeap, u'r right, but many coders used to write down `int`, let's not destroy their habit.

If it is good – tell me please that tests should I complete?

goetas:
ok, you convinced me. Can you add some tests here?

goetas:
https://github.com/schmittjoh/serializer/commit/2cef890a1f4bace48b5f1319b27289bd5e04c41a

Aliance:
Thanx for adding test for me, I would not be able to do that today.
Have you released new tag?

goetas:
Monday is the candidate day. 

But you can try to test `master` on your application :)

Issue #593, opened 2016-05-24
Unexpected results when serializing arrays containing null value elements 


atufkas:
Hi,

I just experienced problems with an unexpected behavior when serializing arrays that contain null value elements, meaning: When _one or more elements_ of the array are set to `null`. Note: I'm not talking about serializing null values at all – I know (need and enabled) this option, but the unexpected behavior affects both scenarios: When serializing null values is enabled as well when it is not.

Example to reproduce the behavior (from a SF controller context):

    $data = array(
        'foo' => null,
        'bar' => array(
            'hal',
            42
        ),
        'baz' => array(
            null
        ),
        'biz' => array(
            null,
            'ford'
        )
    );
        
    $serializer = $this->container->get('jms_serializer');
    $context = SerializationContext::create()->setSerializeNull(true);
    $result = $serializer->serialize($data, 'json', $context);

Now, when dumping the `$result` we get this:

    string(57) "{"foo":null,"bar":["hal",42],"baz":[],"biz":{"1":"ford"}}" 

While the JSON output for `foo` and `bar`looks normal, be sure you noticed the vanished null element for `baz` and strangely indexed second array element of `biz`.

This is obviously not a behavior or bug of `json_encode` itself, simply proven when dumping the result of  `json_encode($data)`, which looks much more like what I would expect:

    string(62) "{"foo":null,"bar":["hal",42],"baz":[null],"biz":[null,"ford"]}"
   
Conclusion: Null value elements in arrays somehow get lost but the array itself – if null values were present – gets converted into an "object" with original numeric indices (with holes were the null values originally resided). Pheeew, a bit complicated to describe, you see it for yourself... 

I would classify this bug (if not my mistake of any kind!) as **severe** since it doesn't allow to deal with null values within arrays at all! That's actually part of my current use case – and I'm pretty sure it is not _that_ exotic.

Totally independently of this specific problem I'd like to say: **Thanks to the author and all contributors for a very helpful piece of software you're sharing with us!**

schmittjoh:
Is this data on an object or do you just have this array structure?

atufkas:
Ah, right, didn't mention this. In this case it's just an array structure I'm building myself...

goetas:
possible solution https://github.com/schmittjoh/serializer/issues/660 (and source of the problem)

PR #660, opened 2016-10-23
Nullable array inconsistency

goetas:
Fixes https://github.com/schmittjoh/serializer/issues/593
Closes https://github.com/schmittjoh/serializer/pull/626
Closes https://github.com/schmittjoh/serializer/pull/36
Closes https://github.com/schmittjoh/serializer/pull/642

Introduced with https://github.com/schmittjoh/serializer/commit/8cb1e3cf549a881575e96dbb64cee264c9337206 containing a non exact test case

The problem looks  to be that the previous test case, having `setSerializeNull(true)`, expects that:

```php
array('foo' => 'bar', 'baz' => null, null)
```
 is serialized into 
```json
// {"foo":"bar","baz":null}
```

instead of

```json
// {"foo":"bar","baz":null,"0":null}
```

@hashnz do you remember the reason behind this test case?


goetas:
Similar to https://github.com/schmittjoh/serializer/pull/626

scaytrase:
Hell yeah this one got fixed! Awaiting for the release

PR #661, opened 2016-10-25
Document the implied 'Default' property group when no group is specified

akoebbe:
Pretty sure the example code should be referencing `details` and not `Default`.

### DIFF
--- a/doc/cookbook/exclusion_strategies.rst
+++ b/doc/cookbook/exclusion_strategies.rst
@@ -86,7 +86,9 @@ Another default exclusion strategy is to create different views of your objects.
 Let's say you would like to serialize your object in a different view depending
 whether it is displayed in a list view or in a details view.
 
-You can achieve that by using the ``@Groups`` annotation on your properties.
+You can achieve that by using the ``@Groups`` annotation on your properties. Any
+property without an explicit ``@Groups`` annotation will be included in a
+``Default`` group, which can be used when specifying groups in the serializer.
 
 .. code-block :: php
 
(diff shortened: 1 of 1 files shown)


goetas:
Pay attention to https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Exclusion/GroupsExclusionStrategy.php#L34

The documentation says that there is a `Default` group  :)

akoebbe:
@goetas Ah ha! Thanks for pointing that out. It wasn't apparent that the `Default` group existed. Where is that in the docs?

goetas:
i guess the document you tried to change, had the intent do document it :)

but an explicit section does not exists probably... interested in adding it?

akoebbe:
Yup. I'll back out the last commit and add an explicit statement about it.

akoebbe:
@goetas let me know if that is clear enough.

doc/cookbook/exclusion_strategies.rst
```diff
@@ -86,7 +86,9 @@ Another default exclusion strategy is to create different views of your objects.
 Let's say you would like to serialize your object in a different view depending
 whether it is displayed in a list view or in a details view.
 
-You can achieve that by using the ``@Groups`` annotation on your properties.
+You can achieve that by using the ``@Groups`` annotation on your properties. Any
+property without an explicit ``@Groups`` annotation will be included in a
+``Default`` group, which can be used when specifying groups in the serializer.
```
goetas: "...groups in the serialization context." ?

goetas:
thanks

PR #667, opened 2016-11-01
Recursion groups

Alexey-Stogny:


goetas:
What is the usecase for this feature?

goetas:
(And should be rebased on the current master, because many of the features you are introducing are already present(

Alexey-Stogny:
@goetas sorry, it was my mistake to create pull request. I took @andreasferber idea, and modified to my needs. This changes give you ability to use only one top-level serialization group, and define nested entities groups in annotations.

goetas:
:+1:

PR #672, opened 2016-11-04
Suggestion to fix problem when visit a array

davidsonalencar:
When visit a array required in a element no required, I reproduce an error `Node no longer exists`. I fixed this problem with this change.

What do you think?


goetas:
please provide a test case for this

goetas:
replaced by https://github.com/schmittjoh/serializer/pull/675 that includes tests

davidsonalencar:
Ok :)

Issue #540, opened 2015-12-17
ConditionalExpose/Exclude annotation


TiMESPLiNTER:
I need to include/exclude some properties based on the content of another property/callback. So what I would like be able to do is this:

``` php
use JMS\Serializer\Annotation as JMS;

/**
 * @JMS\ExclusionPolicy("all")
 */
class Person
{

    /**
      * @var string
      * @JMS\Type("string")
      * @JMS\Expose
      */
     protected $firstName;

    /**
      * @var string
      * @JMS\Type("string")
      * @JMS\Expose
      */
     protected $lastName;

    /**
     * @var boolean
     */
    protected $addressIncluded = true;

    /**
      * @var string
      * @JMS\Type("string")
      * @JMS\ConditionalExpose(method="isAddressIncluded")
      */
     protected $street;

     /**
       * @var string
       * @JMS\Type("string")
       * @JMS\ConditionalExpose(method="isAddressIncluded")
       */
     protected $zipCode;

     /**
       * @var string
       * @JMS\Type("string")
       * @JMS\ConditionalExpose(property="addressIncluded")
       */
     protected $location;

     public function isAddressIncluded()
     {
        return $this->addressIncluded;
     }

    /**
     * @param boolean $addressIncluded
     */
     public function setAddressIncluded($addressIncluded)
     {
        return $this->addressIncluded = $addressIncluded;
     }
}
```

So that the result would be:

``` json
{
    "first_name": "John",
    "last_name": "Doe",
    "street": "2nd street",
    "zip_code": "8000",
    "location": "Zurich"
}
```

or if I set the `includeAddress` property to `false`:

``` php
$person->setIncludeAddress(false);
```

the result should look like this:

``` json
{
    "first_name": "John",
    "last_name": "Doe"
}
```

The callback has to return a `boolean` or the property has to be of type `boolean`.

PR #673, opened 2016-11-04
Implemented dynamic exclusion using symfony expression language

goetas:
This PR allows to use dynamic expressions to decide if exclude or not a field during the serialization process.

Under the hood it uses http://symfony.com/doc/current/components/expression_language.html
```php
class Person
{
    /**
     * @Expose(if="service('some.cool.service').isAllowed(object)")
     */
    public $gender;
    /**
     * @Exclude(if="service('some.cool.service').isAllowed(object)")
     */
    public $gender;
}
```

the "gender" filed will be serialized if the expression `service('some.cool.service').isAllowed(object)` returns false or true depending if the annotation used is exclude or expose.

Closes https://github.com/schmittjoh/serializer/issues/540


src/JMS/Serializer/Serializer.php
```diff
@@ -170,6 +182,9 @@ public function fromArray(array $data, $type, DeserializationContext $context =
 
     private function visit(VisitorInterface $visitor, Context $context, $data, $format, array $type = null)
     {
+        if (null !== $this->expressionEvaluator) {
+            $context->addExclusionStrategy(new DynamicExclusionStrategy($this->expressionEvaluator));
+        }
```
goetas: not really proud on how this is implemented

goetas:
@lsmith77 @schmittjoh @Ener-Getick @xabbuh @willdurand @stof 

Some of your opensource projects relies on the serializer, I would be very grateful if somebody of you can have a look at it and review it a bit?

mvrhov:
This and more is implemented in https://github.com/willdurand/Hateoas#exclusion

goetas:
@mvrhov  I took inspiration from it :)

xabbuh:
@goetas I am sorry, but I am not really familiar with the JMS Serializer internals. So I am not of much help here.

src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
```diff
@@ -158,8 +159,15 @@ public function loadMetadataForClass(\ReflectionClass $class)
                         $propertyMetadata->serializedName = $annot->name;
                     } elseif ($annot instanceof Expose) {
                         $isExpose = true;
+                        if (null !== $annot->if){
+                            $propertyMetadata->excludeIf = "!" . $annot->if;
```
schmittjoh: Probably need to wrap this in parentheses to support something like ``a && b``.

src/JMS/Serializer/Expression/ExpressionEvaluator.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace JMS\Serializer\Expression;
+
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+
+class ExpressionEvaluator implements ExpressionEvaluatorInterface
```
schmittjoh: Is there a need to support another evaluator? Even if there were one, we already have the strategy interface which could be implemented, no?

src/JMS/Serializer/Serializer.php
goetas: @schmittjoh any idea on how to do this in a better way?

src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
goetas: done

goetas:
@schmittjoh can't find any good strategy to throw some exception if the expression language is used for expose/exclude annotation, so i have added this documentation part https://github.com/schmittjoh/serializer/pull/673/commits/65311c48dbf65ef88aa1533306511e4eb968f0dc

enabling it automatically is also not possible because the expression language requires always some configuratuion

goetas:
@schmittjoh i consider this as ready for review

schmittjoh:
Regarding the exception, I think you could add a flag to the class metadata (true/false) whether expressions are used, and then also have a similar flag on the Context whether it is enabled. After loading metadata in the GraphNavigator, you can then compare the flags to see if everything is ok. 

It's a bit of a code smell to hard code something for a single exclusion strategy in those classes, but I think it's worth it in this case since exposing data that should not be exposed can pose a serious security issue, and it's also better for DX. What do you think?

goetas:
@schmittjoh regarding the user experience, I definitely agree with you that is way better throwing an exception in the serialization process instead of just ignoring it.

I have implemented it wit https://github.com/schmittjoh/serializer/pull/673/commits/264f32a6130d6eb6d5552448ac5867e023eda37e but from the code point of view im really not happy... 

What do you think... should we keep it hard-coded as it is now... putting the user experience first, and the the code cleanness ?

src/JMS/Serializer/Context.php
```diff
@@ -106,6 +108,37 @@ public function getExclusionStrategy()
...
     }
 
+    /**
+     * @param string $class
+     * @return bool
+     */
+    public function hasExclusionStrategy($class)
+    {
+        if (isset($this->hasExclusionStrategy[$class])) {
+            return $class;
```
schmittjoh: Should return ``$this->hasExclusionStrategy[$class]`` here.

src/JMS/Serializer/Context.php
```diff
@@ -106,6 +108,37 @@ public function getExclusionStrategy()
...
+            $strategies[] = $this->exclusionStrategy;
+        }
+
+        while (count($strategies)>0) {
+            $strategy = array_shift($strategies);
+
+            if ($strategy instanceof $class) {
+                return $this->hasExclusionStrategy[$class] = true;
+            } elseif ($strategy instanceof DisjunctExclusionStrategy) {
+                $ref = new \ReflectionProperty($strategy, 'delegates');
```
schmittjoh: Better add a getter to that class, maybe even a separate interface like ``DelegatingStrategyInterface`` which other classes could also implement and which exposes a ``getDelegates``.

src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
```diff
@@ -160,10 +160,12 @@ public function loadMetadataForClass(\ReflectionClass $class)
                         $isExpose = true;
                         if (null !== $annot->if) {
                             $propertyMetadata->excludeIf = "!(" . $annot->if . ")";
+                            $classMetadata->usingExpression = true;
```
schmittjoh: You can check this in addPropertyMetadata of the ClassMetadata class, no need to have this in each driver then.

src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
```diff
@@ -158,8 +158,17 @@ public function loadMetadataForClass(\ReflectionClass $class)
                         $propertyMetadata->serializedName = $annot->name;
                     } elseif ($annot instanceof Expose) {
                         $isExpose = true;
+                        if (null !== $annot->if) {
+                            $propertyMetadata->excludeIf = "!(" . $annot->if . ")";
+                            $classMetadata->usingExpression = true;
```
schmittjoh: Better move this to the ``addPropertyMetadata`` method, then we do not have to check this in each driver.

src/JMS/Serializer/Context.php
goetas: my, bad, fixed

src/JMS/Serializer/Context.php
goetas: i prefer to add just the method... `DelegatingStrategyInterface` looks an overkill to me

goetas:
> I think hard-coding is ok for this.

ok, then lets stay with it

> What about deserialization, did you think about that?

It was not my usecase so i have just skipped it. 

Probably for consistency should be implemented something. My personal opinion is that rarely in deserialization you need to skip some field, so the possible scenario are really few

goetas:
@schmittjoh There is one disadvantage in my approach, when using the context factory to add the exclusion strategy... from now on `Context::create()`  become useless and not enough... now starts to be fundamental having access to the context factories to create a new "good" context.

So probably  `Context::create()`  should be deprecated. 

I have already prepared a PR for JMS Serializer Bundle and FOS Rest and will submit it soon.

goetas:
@schmittjoh can you give me your feedback on https://github.com/schmittjoh/serializer/pull/673/commits/aa4c9511b22bb769fa2189e1d5d63a061f143234 ?

Now the creation if contextes is pretty important, and can't be done with a simple static method anymore... what do you think?

schmittjoh:
I've been thinking whether we can change the factory to something that can take a context and populate it with default values where nothing has been explicitly given. I would like to avoid additional complexity for end users if possible.

goetas:
> I've been thinking whether we can change the factory to something that can take a context and populate it with default values where nothing has been explicitly given.

Do not have any really good idea on how to do it. the problem is that exclusion strategies are part of the "context", and a dynamic exclusion strategy needs a wide set of dependencies.

One detail that should be considered is the amount of users using jms serializer via FOS rest bundle, and without FOS.
If they are using FOS rest, this will be fully transparent, since will be really easy to add the context factory via symfony DI container (fos has its own contextes and context factories).

If users are not using FOS, then configuring the expression language is not trivial at all, this means that users will have to figure out how to build contextes anyway.

On the other side, really do not like the static method for creating new contextes. Does not play well with future dependencies (this is more my personal opinion and for the Developer Experience, i'm ready to sacrifice my point of view).

schmittjoh:
We should probably remove exclusion strategies from the context altogether and rather move them to the serializer. The context was intended as a lightweight data object, but with more complex exclusion strategies it is becoming a lot heavier.

With regard to static methods, we can actually deprecate them (they were only needed for older PHP versions). Now, it's just as easy to do ``(new SerializationContext)->somethingHere()``.

goetas:
The only solution i see is adding an extra "if" condition in the `GraphNavigator` (exactly at [here](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/GraphNavigator.php#L227)).

The approach should work. In that way, the "expression" language will become part of the graph navigator implementation.

Is this an acceptable solution?

goetas:
@schmittjoh Implemented as you suggested. 
Now:
- there is an exception if the expression language is used but not configured
- context creation is lightweight

Implementation: 
-  expression language is added to serializer-builder and propagated via constructors to the graph-navigator.
- check is done explicitly in the property loop

This implementation does not require changes on the fos rest bundle :)

goetas:
was time to merge it :)

composer.json
```diff
@@ -38,6 +38,7 @@
         "symfony/validator": "^2.2",
         "symfony/form": "~2.1",
         "symfony/filesystem": "^2.1",
+        "symfony/expression-language": "^2.6|^3.0",
```
Aliance: You add some dependency to `composer.json` but not to `composer.lock`, so I pull the latest master, run `composer install` & then run `./vendor/bin/phpunit`:
> Fatal error: Class 'Symfony\Component\ExpressionLanguage\ExpressionFunction' not found in .../tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php on line 243

I think this should be fixed.

xabbuh: The `composer.lock` file is not committed to git. So you need to run `composer update` instead.

Aliance: My bad, I've seen this file in my IDE and think that it's commited to git.

bblue:
I am happy to see this change, but would it also be possible to add support for dynamic exclusion via the existing exclusion strategies? From a user perspective it seems a strange limitation.

I have tried, but could not get it to work as I'm not familiar enough with the internals of JMS.

I registered my custom exclusion strategy as a service, then made the service a listener for the `serializer.pre_serialize event`. From here I could access the object currently being serialized via `$event->getObject()`

However, any nested objects overwrites whatever is currently the serialized object, and I have not found a way to navigate where the serializer currently is, so that approach was no good... 

But it appears all you need is to make the object of the current property being serialized accessible in custom serializers. That does not sound too hard... or?

Issue #674, opened 2016-11-04
Warning: JMS\Serializer\XmlDeserializationVisitor::visitArray(): Node no longer exists


ama-gi:
I got this error when deserializing after I updated the bundle from 1.3.1 to 1.4.1.

goetas:
thanks for the ticket, already working on it, this is a temporary patch https://github.com/schmittjoh/serializer/pull/672

goetas:
introduced with https://github.com/schmittjoh/serializer/pull/658

goetas:
re-fixed in https://github.com/schmittjoh/serializer/pull/697

PR #675, opened 2016-11-04
Fixed xml arrays with namespaced entry triggers error

goetas:
fixes https://github.com/schmittjoh/serializer/issues/674


goetas:
@davidsonalencar yes, reading the documentation, `$data->count()`  looks for both, namespaced and not namespaced nodes

Issue #666, opened 2016-10-30
Doctrine Proxies and serializer.pre_serialize


scasei:
Listener for 'serializer.pre_serialize' is not called, when class attribute is set and class is a Doctrine Entity.

This example will not trigger, because in EventDispatcher `$class` is 'Proxies__CG__\Prima\Catalog\MediaBundle\Entity\Media' instead of 'Prima\Catalog\MediaBundle\Entity\Media'.

``` php
<?php

namespace Prima\Catalog\MediaBundle\Listener;

use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
use JMS\Serializer\EventDispatcher\ObjectEvent;
use JMS\Serializer\EventDispatcher\PreSerializeEvent;

/**
 * Add data after serialization.
 */
class SerializationListener implements EventSubscriberInterface
{
    /**
     * @var \Sonata\MediaBundle\Provider\FileProvider
     */
    protected $file_provider;


    /**
     * @param \Sonata\MediaBundle\Provider\FileProvider        $file_provider
     */
    public function __construct(
            \Sonata\MediaBundle\Provider\FileProvider $file_provider
    ) {
        $this->file_provider = $file_provider;
    }

    /**
     * {@inheritdoc}
     */
    public static function getSubscribedEvents()
    {
        return array(
            array(
                'event' => 'serializer.pre_serialize', 
                'class' => 'Prima\Catalog\MediaBundle\Entity\Media', 
                'method' => 'onPreSerialize',
            ),
        );
    }

    /**
     * @param PreSerializeEvent $event
     */
    public function onPreSerialize(PreSerializeEvent $event)
    {
        /* @var $object \Prima\Catalog\MediaBundle\Entity\Media */
        $object = $event->getObject();

        $object->setBinaryTransport($this->file_provider->getReferenceFile($object)->getContent());
    }
}
```

goetas:
I guess this can be solved adding your listener on the proxy class too.

scasei:
You are right, this is my workaround. But this should be mentioned somewhere or there could be a solution like detecting this classes. 
Should I provide a PR?

goetas:
Can be fixed probably detecting in some event listener the doctrine proxy class...

goetas:
@scasei Can you check the priority of your listener? `DoctrineProxySubscriber` should do exactly what you need

scasei:
There is no priority set, so its default, or '0'.

[...]

PR #677, opened 2016-11-18
Detect doctrine proxies when serializing a polymorphic type

dragosprotung:
Doctrine Proxies and serializer.pre_serialize schmittjoh/serializer#666 https://github.com/schmittjoh/serializer/pull/677

### DIFF
--- a/src/JMS/Serializer/GraphNavigator.php
+++ b/src/JMS/Serializer/GraphNavigator.php
@@ -151,7 +151,12 @@ final class GraphNavigator
                     // metadata for the actual type of the object, not the base class.
                     if (class_exists($type['name'], false) || interface_exists($type['name'], false)) {
                         if (is_subclass_of($data, $type['name'], false)) {
-                            $type = array('name' => get_class($data), 'params' => array());
+                            if (in_array('Doctrine\Common\Proxy\Proxy', class_implements($data), true)) {
+                                $name = get_parent_class($data);
+                            } else {
+                                $name = get_class($data);
+                            }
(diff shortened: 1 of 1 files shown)


src/JMS/Serializer/GraphNavigator.php
```diff
@@ -151,7 +151,12 @@ public function accept($data, array $type = null, Context $context)
                     // metadata for the actual type of the object, not the base class.
                     if (class_exists($type['name'], false) || interface_exists($type['name'], false)) {
                         if (is_subclass_of($data, $type['name'], false)) {
-                            $type = array('name' => get_class($data), 'params' => array());
+                            if (in_array('Doctrine\Common\Proxy\Proxy', class_implements($data), true)) {
```
goetas: Any alternative idea?

dragosprotung: I chose class_implements over instanceof just so Doctrine does not become a hard dependency.
The proxy is also used in DoctrineProxySubscriber.

goetas: the thing is that `DoctrineProxySubscriber` is an optional dependency, that acts as any other event listener, is not in the "core" of the application, and can be even put as external package.

Currently is part of the official package just for simplicity

dragosprotung: The the only other alternative I see is to iterate over the parents in the event dispatcher and check if there is a listener on them. But I am not sure what impact this can have.

PR #678, opened 2016-11-20
Restrict doctrine annotations version

tcitworld:
To keep php 5.5 compatibility. [Doctrine 1.3](https://github.com/doctrine/annotations/releases/tag/v1.3.0) requires php 5.6 at least.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "jms/parser-lib": "1.*",
         "phpoption/phpoption": "^1.1",
         "phpcollection/phpcollection": "~0.1",
-        "doctrine/annotations": "^1.0",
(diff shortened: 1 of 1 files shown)


goetas:
why this? composer.json is about the minimum version compatibility. doctrine annotations 1.0 should be compatible with 5.5, is it?

if somebody wants doctrine annotations 1.3,  is up to him use the right version of php... or i'm missing something...

goetas:
the problem you are facing is that probably you have generated the lock file using a different php version that what is on the server ... https://github.com/wallabag/wallabag/issues/2617

tcitworld:
Indeed, didn't thought of that. Thanks.

PR #682, opened 2016-12-05
Improved Symfony 3.x compatibility 

goetas:
Replaces https://github.com/schmittjoh/serializer/pull/542 https://github.com/schmittjoh/serializer/pull/491


goetas:
This will require proper configuration of the JMS bundle that will have to pick the right event subscriber based on symfony version

PR #688, opened 2016-12-07
Added clearing previous libxml errors

zerkms:
For the empty document string `libxml_*` error functions return the *previous* error, not the *error caused by last xml operation*.

The simple reproduction script:

https://3v4l.org/RKEb0

```php
$previous = libxml_use_internal_errors(true);
$doc = simplexml_load_string('foo');
libxml_use_internal_errors($previous);

$result = libxml_get_last_error();

$previous = libxml_use_internal_errors(true);
$doc = simplexml_load_string('');
libxml_use_internal_errors($previous);

$result = libxml_get_last_error();

var_dump($doc, $result);
```

What it means is that if there is any "previous error" it would mask the problem with `libxml_get_last_error` returning not what the JMSSerializer expects.

In this PR I'm sending the test (so the travis run here fails deliberately) that should have passed and that demonstrates initial problem (masked with the lack of `libxml_clear_errors()` call).

I additionally sent a php bug report https://bugs.php.net/bug.php?id=73670, to either fix the php documentation or `libxml`-related implementation.

In either case, JMSSerializer should handle it properly.

### DIFF
--- a/src/JMS/Serializer/Metadata/Driver/XmlDriver.php
+++ b/src/JMS/Serializer/Metadata/Driver/XmlDriver.php
@@ -33,6 +33,8 @@ class XmlDriver extends AbstractFileDriver
     protected function loadMetadataFromFile(\ReflectionClass $class, $path)
     {
         $previous = libxml_use_internal_errors(true);
+        libxml_clear_errors();
+
         $elem = simplexml_load_file($path);
         libxml_use_internal_errors($previous);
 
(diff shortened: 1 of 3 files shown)


zerkms:
@goetas this PR has a failing test deliberately: the whole purpose of this PR is to demonstrate a problem.

I did not include a solution for it because for me there is no a really obvious one existing (for me).

So I created the #689 and provided a possible solution that the library maintainers need to assess.

zerkms:
@goetas I think this PR also need to be merged: it contains the unit test for the #689

zerkms:
Oops, too slow :-) I did not have the whole environment set up at a new desktop at home, so could not merge it faster than you :-)

goetas:
commits schmittjoh serializer

something weird...  was sure that this was included in the firts merge... anyway.. now looks good

PR #689, opened 2016-12-07
A possible fix for the #688

zerkms:
Yep, it might look awkward, but I'm open for suggestions.

An issue it fixes: #688


goetas:
what about adding some tests?

zerkms:
@goetas there is a failing test that I added in the referred PR https://github.com/schmittjoh/serializer/pull/688/files

I deliberately split those, so that it was possible to accept either or both separately

goetas:
thanks @zerkms

PR #692, opened 2017-01-05
added support for xml-attributes as discriminators

twtinteractive:
Until now when using Discriminator-maps with XML, the discriminator was hard coded to be a child node. This PR allows to use an attribute instead if desired.

```
 /**
 * @Discriminator(field = "type", map = {"car": "Car", "moped": "Moped"})
 */
 abstract class Vehicle { }
 class Car extends Vehicle { }
 class Moped extends Vehicle { }
 $serializer->serialize(new Car(), 'xml');
```

would result in
```
<vehicle>
    <type>car</type>
</vehicle>
```
now you can do this:

```
/**
 * @Discriminator(field = "type", map = {"car": "Car", "moped": "Moped"})
 * @XmlDiscriminatorAttribute
 */
abstract class Vehicle { }
class Car extends Vehicle { }
class Moped extends Vehicle { }

$serializer->serialize(new Car(), 'xml'); 
```
to get:
```
<vehicle type="car" />
```

### DIFF
--- a/src/JMS/Serializer/GraphNavigator.php
+++ b/src/JMS/Serializer/GraphNavigator.php
@@ -261,6 +261,11 @@ final class GraphNavigator
                 $typeValue = (string) $data->{$metadata->discriminatorFieldName};
                 break;
 
+            // Check XML attribute for discriminatorFieldName
+            case is_object($data) && $metadata->xmlDiscriminatorAttribute && isset($data[$metadata->discriminatorFieldName]):
+                $typeValue = (string) $data[$metadata->discriminatorFieldName];
+                break;
+
             default:
                 throw new \LogicException(sprintf(
                     'The discriminator field name "%s" for base-class "%s" was not found in input data.',
--- a/src/JMS/Serializer/GraphNavigator.php
+++ b/src/JMS/Serializer/GraphNavigator.php
@@ -257,13 +257,13 @@ final class GraphNavigator
                 $typeValue = (string) $data[$metadata->discriminatorFieldName];
                 break;
 
-            case is_object($data) && isset($data->{$metadata->discriminatorFieldName}):
-                $typeValue = (string) $data->{$metadata->discriminatorFieldName};
-                break;
-
             // Check XML attribute for discriminatorFieldName
             case is_object($data) && $metadata->xmlDiscriminatorAttribute && isset($data[$metadata->discriminatorFieldName]):
                 $typeValue = (string) $data[$metadata->discriminatorFieldName];
+                break;
+
+            case is_object($data) && isset($data->{$metadata->discriminatorFieldName}):
+                $typeValue = (string) $data->{$metadata->discriminatorFieldName};
                 break;
 
             default:
--- a/src/JMS/Serializer/Metadata/ClassMetadata.php
+++ b/src/JMS/Serializer/Metadata/ClassMetadata.php
@@ -58,6 +58,8 @@ class ClassMetadata extends MergeableClassMetadata
     public $discriminatorMap = array();
     public $discriminatorGroups = array();
 
+    public $xmlDiscriminatorAttribute = false;
+
     public function setDiscriminator($fieldName, array $map, array $groups = array())
     {
         if (empty($fieldName)) {
@@ -72,6 +74,10 @@ class ClassMetadata extends MergeableClassMetadata
         $this->discriminatorFieldName = $fieldName;
         $this->discriminatorMap = $map;
         $this->discriminatorGroups = $groups;
+    }
+
+    public function setXmlDiscriminator($attribute = false) {
+        $this->xmlDiscriminatorAttribute = $attribute;
     }
 
     /**
@@ -203,6 +209,7 @@ class ClassMetadata extends MergeableClassMetadata
                 $this->discriminatorGroups
             );
             $discriminatorProperty->serializedName = $this->discriminatorFieldName;
+            $discriminatorProperty->xmlAttribute = $this->xmlDiscriminatorAttribute;
             $this->propertyMetadata[$this->discriminatorFieldName] = $discriminatorProperty;
         }
 
--- a/src/JMS/Serializer/Metadata/ClassMetadata.php
+++ b/src/JMS/Serializer/Metadata/ClassMetadata.php
@@ -59,6 +59,7 @@ class ClassMetadata extends MergeableClassMetadata
     public $discriminatorGroups = array();
 
     public $xmlDiscriminatorAttribute = false;
+    public $xmlDiscriminatorCData = true;
 
     public function setDiscriminator($fieldName, array $map, array $groups = array())
     {
@@ -253,6 +254,7 @@ class ClassMetadata extends MergeableClassMetadata
             parent::serialize(),
             'discriminatorGroups' => $this->discriminatorGroups,
             'xmlDiscriminatorAttribute' => $this->xmlDiscriminatorAttribute,
+            'xmlDiscriminatorCData' => $this->xmlDiscriminatorCData,
         ));
     }
 
@@ -285,6 +287,10 @@ class ClassMetadata extends MergeableClassMetadata
 
         if (isset($deserializedData['xmlDiscriminatorAttribute'])) {
             $this->xmlDiscriminatorAttribute = $deserializedData['xmlDiscriminatorAttribute'];
+        }
+
+        if (isset($deserializedData['xmlDiscriminatorCData'])) {
+            $this->xmlDiscriminatorCData = $deserializedData['xmlDiscriminatorCData'];
         }
 
         parent::unserialize($parentStr);
--- a/src/JMS/Serializer/Metadata/ClassMetadata.php
+++ b/src/JMS/Serializer/Metadata/ClassMetadata.php
@@ -76,10 +76,6 @@ class ClassMetadata extends MergeableClassMetadata
         $this->discriminatorGroups = $groups;
     }
 
-    public function setXmlDiscriminator($attribute = false) {
-        $this->xmlDiscriminatorAttribute = $attribute;
-    }
-
     /**
      * Sets the order of properties in the class.
      *
@@ -256,6 +252,7 @@ class ClassMetadata extends MergeableClassMetadata
             $this->discriminatorGroups,
             parent::serialize(),
             'discriminatorGroups' => $this->discriminatorGroups,
+            'xmlDiscriminatorAttribute' => $this->xmlDiscriminatorAttribute,
         ));
     }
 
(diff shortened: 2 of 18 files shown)


goetas:
Thanks for the pull request. Have a look also to https://github.com/schmittjoh/serializer/pull/560

Some considerations: 
 - instead of adding a new annotation (that requires a lot of changes...) what about properties of the `@Discriminator` annotation?
- this then opens possibilities as namespaces... and many other things

fivetide:
Actually adding this to Discriminator, just like in #560 was our first approach, but after discussing this internally we concluded Discriminator would be better off remaining format-agnostic. With additional options in mind, i would like to suggest changing the new annotation to

```
@XmlDiscriminator(attribute=true)
```
which would enable future changes like
```
@XmlDiscriminator(cdata=false)
@XmlDiscriminator(namespace="foo/bar")
```
and changing the other drivers accordingly...

goetas:
hmm.. sounds reasonable. even now we have almost always xml-specific annotations.

So interesting options would be:

```
@XmlDiscriminator(
attribute=true, // default = false
cdata=true, // default = false
namespace="http://xxx"  // default = NULL
)
```

right?

fivetide:
Correct, just one question: I belive the current behaviour is to use cdata, which means we should default cdata to true...?

goetas:
@fivetide you are right, my bad, i have updated my comment

fivetide:
Okay, i will propably get that done on monday, will update the PR then :)

fivetide:
Ok, i updated the PR with the changes we agreed on.

src/JMS/Serializer/Metadata/ClassMetadata.php
```diff
@@ -58,6 +58,9 @@ class ClassMetadata extends MergeableClassMetadata
     public $discriminatorMap = array();
     public $discriminatorGroups = array();
 
+    public $xmlDiscriminatorAttribute = false;
+    public $xmlDiscriminatorCData = true;
```
goetas: Should it be `$xmlDiscriminatorCDATA` or `$xmlDiscriminatorCdata` ?

src/JMS/Serializer/Metadata/Driver/YamlDriver.php
```diff
@@ -257,6 +257,15 @@ private function addClassProperties(ClassMetadata $metadata, array $config)
             $metadata->xmlRootNamespace = (string) $config['xml_root_namespace'];
         }
 
+        if (isset($config['xml_discriminator'])) {
+            if (isset($config['xml_discriminator']['attribute'])) {
+                $metadata->xmlDiscriminatorAttribute = $config['xml_discriminator']['attribute'];
```
goetas: cast to boolean please

src/JMS/Serializer/Metadata/Driver/YamlDriver.php
```diff
@@ -257,6 +257,15 @@ private function addClassProperties(ClassMetadata $metadata, array $config)
...
+            }
+            if (isset($config['xml_discriminator']['cdata'])) {
+                $metadata->xmlDiscriminatorCData = $config['xml_discriminator']['cdata'];
```
goetas: cast to boolean please

src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
```diff
@@ -99,6 +100,9 @@ public function loadMetadataForClass(\ReflectionClass $class)
                 } else {
                     $classMetadata->setDiscriminator($annot->field, $annot->map, $annot->groups);
                 }
+            } elseif ($annot instanceof XmlDiscriminator) {
+                $classMetadata->xmlDiscriminatorAttribute = $annot->attribute;
```
goetas: cast to boolean please

src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
```diff
@@ -99,6 +100,9 @@ public function loadMetadataForClass(\ReflectionClass $class)
...
+                $classMetadata->xmlDiscriminatorCData = $annot->cdata;
```
goetas: cast to boolean please

tests/JMS/Serializer/Tests/Metadata/Driver/xml/Discriminator.ObjectWithXmlAttributeDiscriminatorParent.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<serializer>
+    <class name="JMS\Serializer\Tests\Fixtures\Discriminator\ObjectWithXmlAttributeDiscriminatorParent"
+           discriminator-field-name="type"
+           xml-discriminator-attribute="true"
```
goetas: what is this attribute (i mean `xml-discriminator-attribute`)?

tests/JMS/Serializer/Tests/Metadata/Driver/yml/Discriminator.ObjectWithXmlAttributeDiscriminatorParent.yml
```diff
@@ -0,0 +1,8 @@
+JMS\Serializer\Tests\Fixtures\Discriminator\ObjectWithXmlAttributeDiscriminatorParent:
+    discriminator:
+        field_name: type
+        map:
+            child: JMS\Serializer\Tests\Fixtures\Discriminator\ObjectWithXmlAttributeDiscriminatorChild
+    xml_discriminator:
+        attribute: true
+        cdata: false
```
goetas: what about having a structure as:

```
JMS\Serializer\Tests\Fixtures\Discriminator\ObjectWithXmlAttributeDiscriminatorParent:
    discriminator:
        field_name: type
        map:
            child: JMS\Serializer\Tests\Fixtures\Discriminator\ObjectWithXmlAttributeDiscriminatorChild
        xml_attribute: true
        xml_element:
            cdata: false
            namespace: http://www.w3.org/2005/Atom
```

this is not a big deal, but is just to keep consistency with the current YAML metadata format.
(see http://jmsyst.com/libs/serializer/master/reference/yml_reference the xml_attribute property example)

tests/JMS/Serializer/Tests/Serializer/XmlSerializationTest.php
```diff
@@ -361,6 +363,20 @@ public function testWithoutFormatedOutputByXmlSerializationVisitor()
         $this->assertXmlStringEqualsXmlString($this->getContent('simple_class_object_minified'), $stringXml);
     }
 
+    public function testDiscriminatorAsXmlAttribute()
+    {
+        $xml = simplexml_load_string($this->serialize(new ObjectWithXmlAttributeDiscriminatorChild()));
+        $this->assertEquals('type="child"', trim($xml->xpath('//result/@type')[0]->saveXML()));
```
goetas: instead of "re-loading" the xml into a simplexml object, why not re-use the `$this->getContent('template_name')` method? the whole test case will be: 

```php
$xml = $this->serialize(new ObjectWithXmlAttributeDiscriminatorChild());
$this->assertEquals($this->getContent('xml_discriminator_attribute'), $xml);
$this->assertInstanceOf(
    ObjectWithXmlAttributeDiscriminatorChild::class,
    $this->deserialize(
         $xml,
        ObjectWithXmlAttributeDiscriminatorParent::class
     )
);
```

src/JMS/Serializer/GraphNavigator.php
```diff
@@ -261,6 +261,11 @@ private function resolveMetadata($data, ClassMetadata $metadata)
                 $typeValue = (string) $data->{$metadata->discriminatorFieldName};
                 break;
 
+            // Check XML attribute for discriminatorFieldName
+            case is_object($data) && $metadata->xmlDiscriminatorAttribute && isset($data[$metadata->discriminatorFieldName]):
```
goetas: i will put this statement before the previous one

src/JMS/Serializer/Metadata/ClassMetadata.php
fivetide: capitalizing CD follows the convention used in PropertyMetadata::$xmlElementCData ...

tests/JMS/Serializer/Tests/Metadata/Driver/xml/Discriminator.ObjectWithXmlAttributeDiscriminatorParent.xml
fivetide: sorry, forgot to remove that one...

tests/JMS/Serializer/Tests/Serializer/XmlSerializationTest.php
fivetide: What would be the benefit when comparing to a predefined xml? I assumed the test should only cover if the discriminator is rendered as an attribute, while other aspects of xml-generation (whitespace, xml-header, if the tag is rendered as self-closing...) should not break the test, which might happen when comparing to predefined xml..

tests/JMS/Serializer/Tests/Metadata/Driver/yml/Discriminator.ObjectWithXmlAttributeDiscriminatorParent.yml
fivetide: Having the YAML format being wildly different to annotations and xml feels odd to me, but if you insist, i will change it...

goetas: My idea is just to keep it consistent. I would like to keep consistency within a format mainly. 
Having  consistency across formats is nice to have. 

Since XML and YAML formats have already differences,  and users use only one of them, will try to keep them consistent with the rest of the format itself.

tests/JMS/Serializer/Tests/Serializer/XmlSerializationTest.php
goetas: This is an interesting topic about testing, and how much we want to do "unit" testing...

Your test currently tests only of the type is rendered as attribute... but what if for some other change, make the type to be rendered using both strategies.. I mean as attribute and as element? This test will not detect it.

Of course, testing the whole xml content, has some drawbacks, as example will give some "false" failures if we decide to change indentation or other format specific differences... that generally should not make fail your testcase.

To be honest, I prefer to have a false failure, instead of having a not detected failure.

(this means less "unit" testing, and going more in direction of functional or integration tests... (event if the border between this definitions is really blurry)

What is your opinion on it?

fivetide: Personally and totally subjective, i like to have my tests represent actual requirements and i like to be able to treat them as much. When working with foreign code, i can not determine of a false failure is indeed false, or if there is an actual requirement... But since in this case, it is pretty obvious, i feel okay doing it your way.

src/JMS/Serializer/Metadata/Driver/YamlDriver.php
```diff
@@ -278,6 +278,17 @@ private function addClassProperties(ClassMetadata $metadata, array $config)
                 }
                 $groups = isset($config['discriminator']['groups']) ? $config['discriminator']['groups'] : array();
                 $metadata->setDiscriminator($config['discriminator']['field_name'], $config['discriminator']['map'], $groups);
+
```
goetas: PSR standards, max one empty line

fivetide:
Thanks a lot! :+1:

PR #694, opened 2017-01-09
Use GroupsExclusionStrategy::DEFAULT_GROUP instead default group.

Aliance:
I think using the `GroupsExclusionStrategy::DEFAULT_GROUP` constant is more correct.


goetas:
looks correct... give me just some time to make sure everything works well

Aliance:
Should I fix some cs failures?

goetas:
no, looks something broken with scrutinizer itself...

@schmittjoh ping

Issue #695, opened 2017-01-09
Deserializing XMLList with Namespaces not (always) working as intended


fivetide:
Deserializing XML with only namespaced child-nodes does not work as intended (in #675).

My testcase, derived from testObjectWithNamespacesAndList():

``` 
public function testObjectWithOnlyNamespacesAndList()
    {
        $object = new ObjectWithNamespacesAndList();

        $object->phones = array();
        $object->addresses = array();

        $object->phonesAlternativeB = array();
        $object->addressesAlternativeB = array();

        $object->phonesAlternativeC = array('777', '888');
        $object->addressesAlternativeC = array('A'=>'Street 7', 'B'=>'Street 8');

        $object->phonesAlternativeD = array();
        $object->addressesAlternativeD = array();

        $this->assertEquals(
            $this->getContent('object_with_only_namespaces_and_list'),
            $this->serialize($object, SerializationContext::create())
        );
        $this->assertEquals(
            $object,
            $this->deserialize($this->getContent('object_with_only_namespaces_and_list'), get_class($object))
        );
    }
``` 

with this xml: 
```
<?xml version="1.0" encoding="UTF-8"?>
<ObjectWithNamespacesAndList xmlns="http://example.com/namespace" xmlns:x="http://example.com/namespace2">
  <x:phone><![CDATA[777]]></x:phone>
  <x:phone><![CDATA[888]]></x:phone>
  <x:address id="A"><![CDATA[Street 7]]></x:address>
  <x:address id="B"><![CDATA[Street 8]]></x:address>
</ObjectWithNamespacesAndList>
```

currently fails because of src/JMS/Serializer/XmlDeserializationVisitor.php:153

`0 === $data->count()`

(see https://3v4l.org/n7o8I).

I would like to provide a PR to solve this, but my current approaches both seem subobtimal, either just silencing the warning from #674, or using a temporary error-handler every time $hasNode is determined... 
What do you think?

goetas:
What about https://3v4l.org/EatEF ?

goetas:
see https://github.com/schmittjoh/serializer/pull/697 as alternative

goetas:
solved in https://github.com/schmittjoh/serializer/pull/697

PR #696, opened 2017-01-09
Bugfix for #695, #674

twtinteractive:
During XML deserialization, do not visit property if that node is absent.

fixes Deserializing XMLList with Namespaces not (always) working as intended #695


fivetide:
It seems, despite all tests running green, this does not solve all edge-cases (like ArrayCollection<type>). I will add some more changes for these...

goetas:
implemented in https://github.com/schmittjoh/serializer/pull/697

PR #697, opened 2017-01-09
Deserializing XMLList with Namespaces not (always) working as intended

goetas:
Trying to solve https://github.com/schmittjoh/serializer/issues/695 using XPATH

relates #696, #674

The advantage of this is the use or one single xpath xpression to do the whole job.

no multiple `isset`, `count` or `children`. 

the disadvantage is that xpath should be probably slower

closes https://github.com/schmittjoh/serializer/pull/697  #695

### DIFF
--- a/src/JMS/Serializer/XmlDeserializationVisitor.php
+++ b/src/JMS/Serializer/XmlDeserializationVisitor.php
@@ -150,13 +150,15 @@ class XmlDeserializationVisitor extends AbstractVisitor
             $namespace = isset($classMetadata->xmlNamespaces[''])?$classMetadata->xmlNamespaces['']:$namespace;
         }
 
-        if (0 === $data->count()){
-            $hasNode = false;
+        if (null !== $namespace) {
+            $prefix = uniqid('ns-');
+            $data->registerXPathNamespace($prefix, $namespace);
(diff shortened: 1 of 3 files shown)


goetas:
some kind of benchmark https://3v4l.org/f5r5i

goetas:
The benchmark show that is obvioulsy slower (xpath...) but the solution looks reliable... can not find an alternative way to count child elements  in a node

goetas:
@schmittjoh this is not the fastest option, but looks the most "reliable" ... what do you think?

fivetide:
Another possible solution could be #696, which is meant to fix the root cause to #674, and reverting visitArray to the previous version. The downside is that the SimpleXmlElement is modified to emulate the existence of the absent collection node.

fivetide:
Do you guys see and pro/con for either of these fixes?

goetas:
- https://github.com/schmittjoh/serializer/pull/696  con: hacky, pro: works (faster?)
- https://github.com/schmittjoh/serializer/pull/697  con: slow, pro: clean

to be honest i prefer a clean approach. performances in xml are generally low...

goetas:
xpath is already widely used in the xml deserializer

PR #698, opened 2017-01-11
[RFC] Extra parameter that allows to typehint the first step serialization

goetas:
this will allow:

-  putting custom handlers and virtual types on the first level serialization objects
- solve some issues with json/yaml arrays serialized as hashes
- ?? something else

- What about the `toArray` and `fromArray`?
- What about additional XML options?

### DIFF
--- a/src/JMS/Serializer/Serializer.php
+++ b/src/JMS/Serializer/Serializer.php
@@ -88,15 +88,19 @@ class Serializer implements SerializerInterface, ArrayTransformerInterface
         $this->deserializationContextFactory = new DefaultDeserializationContextFactory();
     }
 
-    public function serialize($data, $format, SerializationContext $context = null)
+    public function serialize($data, $format, SerializationContext $context = null, $type = null)
     {
         if (null === $context) {
             $context = $this->serializationContextFactory->createSerializationContext();
         }
 
+        if (null !== $type) {
+            $type = $this->typeParser->parse($type);
+        }
+
         return $this->serializationVisitors->get($format)
-            ->map(function(VisitorInterface $visitor) use ($context, $data, $format) {
-                $this->visit($visitor, $context, $visitor->prepare($data), $format);
+            ->map(function(VisitorInterface $visitor) use ($context, $data, $format, $type) {
+                $this->visit($visitor, $context, $visitor->prepare($data), $format, $type);
 
                 return $visitor->getResult();
             })
--- a/src/JMS/Serializer/SerializerInterface.php
+++ b/src/JMS/Serializer/SerializerInterface.php
@@ -30,11 +30,12 @@ interface SerializerInterface
      *
      * @param object|array|scalar $data
      * @param string $format
-     * @param Context $context
+     * @param SerializationContext $context
+     * @param string $type
      *
      * @return string
      */
-    public function serialize($data, $format, SerializationContext $context = null);
+    public function serialize($data, $format, SerializationContext $context = null, $type = null);
 
     /**
      * Deserializes the given data to the specified type.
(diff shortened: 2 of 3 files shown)


apapsch:
Why is a type hint needed? I can imagine all applications doing more or less the same thing, along these lines:

```
function make_jms_typehint($it) {
  if (is_array($it) || $it instanceof \ArrayAccess && $it instanceof \Countable) {
     if (count($it) === 0) {
       $innerTypeHint = "";
     } else {
       $innerTypeHint = sprintf("<%s>", make_jms_typehint($it[0]));
     }
     return "array" . $innerTypeHint;
  }
  elseif (is_object($it)) {
    return get_class($it);
  }
  else {
    return null;
  }
}
```

That could also be done by the serializer itself.

schmittjoh:
How about adding this to the context itself?

goetas:
IMO the context should contain information that might be useful during the whole serialization process.

Adding the "first typehint" to the context, will be useful only in the first step of the serialization process.

goetas:
@apapsch your code will not be able to distinguish this 3 cases:

- `array<object>`
- `array<string, object>`
- `array<integer, object>`

apapsch:
`array<object>` looks like it's the same as `array<integer, object>`. Under this assumption, `make_jms_typehint` could always generate the form `array<key_type, value_type>`. Also, the last else block would need to `return gettype($it)`.

What to do in case keys or values have different types? For example, PDOStatement::fetch returns arrays with integer and string keys by default.

apapsch:
> solve some issues with json/yaml arrays serialized as hashes

Even with the typehint `array`, the result in #706 is a hash. This seems to do with the use of `ArrayObject`, which is treated differently to `array` by `json_encode`. Reproduce with:

```
$my = new ArrayObject(["asdf", "asdf"]);
echo json_encode((array) $my);
echo json_encode($my);
```

IMHO, it should print the same both times. To me it looks like a PHP bug. Shouldn't `ArrayObject` behave just like `array`?

`JsonSerializationVisitor` is constructing `ArrayObject`s. `getResult` is passing the root value to `json_encode`. To mitigate my bug, I applied this quick fix:

```
     public function getResult()
     {
-        $result = @json_encode($this->getRoot(), $this->options);
+        $root = $this->getRoot() instanceof \ArrayObject ? (array) $this->getRoot() : $this->getRoot();
+        $result = @json_encode($root, $this->options);
 
         switch (json_last_error()) {
```

Looking at the comments around `ArrayObject` it seems it is used to force hash output, so my quick fix can't be applied directly. But `ArrayObject` could be converted to `array` if the type hint is `array`.

fullbl:
Hi, is there any news on this?

goetas:
@fullbl as you can see, this is targeted to 2.0, that currently has not a specific release plan.

Till the 2.0 release you will have to do some workarounds for it

apapsch:
You might want to use this workaround for the (feature) bug ArrayObject. It's my quick fix:

```
diff --git a/src/JMS/Serializer/JsonSerializationVisitor.php b/src/JMS/Serializer/JsonSerializationVisitor.php
index bda77e3..01f1370 100644
--- a/src/JMS/Serializer/JsonSerializationVisitor.php
+++ b/src/JMS/Serializer/JsonSerializationVisitor.php
@@ -26,7 +26,8 @@ class JsonSerializationVisitor extends GenericSerializationVisitor
 
     public function getResult()
     {
-        $result = @json_encode($this->getRoot(), $this->options);
+        $root = $this->getRoot() instanceof \ArrayObject ? (array) $this->getRoot() : $this->getRoot();
+        $result = @json_encode($root, $this->options);
 
         switch (json_last_error()) {
             case JSON_ERROR_NONE:
```

goetas:
@schmittjoh 

> How about adding this to the context itself?

Sounds a good idea to keep it compatible with current API

goetas:
proceeding in https://github.com/schmittjoh/serializer/pull/743

PR #703, opened 2017-01-17
there is no reason to limit this only to private access level

skydiablo:
changed this to "protected" keep the mind open for extending this class more comfortable.

### DIFF
--- a/src/JMS/Serializer/Handler/DateHandler.php
+++ b/src/JMS/Serializer/Handler/DateHandler.php
@@ -68,7 +68,7 @@ class DateHandler implements SubscribingHandlerInterface
         $this->xmlCData = $xmlCData;
     }
 
-    private function serializeDateTimeInterface(
+    protected function serializeDateTimeInterface(
         VisitorInterface $visitor,
         \DateTimeInterface $date,
         array $type,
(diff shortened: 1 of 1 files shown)


goetas:
if you need a specific date time parsing, i suggest you to implement your own datetime handler 
(implementing only what you have to handle instead of extending this)

The current datetime handler is not supposed to be part of the public api.

goetas:
If you are fixing a bug, a bugfix PR is welcome

skydiablo:
nobody will hurt if we change this functions to protected access level?...

goetas:
It will hurt the project extensibility.

A protected variable is a potential dependency.
 if some application extends the class using that variable, but the
serializer needs to change it, it can't be done because for your
application will be a BC break

skydiablo:
i see... but if we mark this class als internal API, the risk it is on my side!?

goetas:
Yes, but if "you" have a BC issue with your application because of the
serializer , you feel bad about the serializer, not about your self :)

On 18 Jan 2017 12:41 pm, "Volker" <notifications@github.com> wrote:

i see... but if we mark this class als internal API, the risk it is on my
side!?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/schmittjoh/serializer/pull/703#issuecomment-273454321>,
or mute the thread

PR #707, opened 2017-01-24
Add default class for discriminator

vsychov:


goetas:
tests are missing.

can you explain the propose of this feature?

vsychov:
This pull request, added new feature for make possible use default class for discriminator, in case if value not exists in the map. Without pull request it throw exception.

goetas:
without tests, can't be merged

vsychov:
ok, I will add tests

vsychov:
tests added

vsychov:
Hi, when this will be merged?

Issue #713, opened 2017-02-10
Serialized DateTime instances are not valid ISO-8601


xelan:
The [default format of JMS\Serializer\Handler\DateHandler is \DateTime::ISO8601](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Handler/DateHandler.php#L64), which is however [not fully compatible with ISO 8601](http://php.net/manual/en/class.datetime.php). Specificially, the time zone delimiter is missing.

Example:
`2016-04-06T14:39:32+0200`
should be:
`2016-04-06T14:39:32+02:00`

As a result, the serialized string can't be validated via XML schema (xs:dateTime) and the interoperability with other software is also limited.
The PHP documentation suggests to use `DateTime::ATOM` instead of `DateTime::ISO8601` for ISO 8601 compatibility.

In my opinion, the fix would technically be a breaking change, as the serialized XML/JSON would be different from that of previous versions. The deserialization would need to support both variants for backwards compatibility, newly serialized documents should use the standard-compliant DateTime format as default.

What do you think?

xelan:
Deserialization already seems to work with both `DateTime::ISO8601` and `DateTime::ATOM` (Tested with jms/serializer 1.4.2) using JSON and XML.

mvrhov:
This can also fixed when DateHandler instance is created. And the bundle already supports this as you can set the format in your configuration.

xelan:
Thanks for your fast response, @mvrhov. What is the recommended way to set the DateHandler for the serializer? I don't use the bundle.

To fix the issue and provide a standards-compliant default, I created a patch which sets the default to DateTime::ATOM and adapted the test suite accordingly.

goetas:
If you are not using the bundle, then probably you are using the serialization builder.

```php
$builder->configureHandlers(function($h) {
        $h->registerSubscribingHandler(new DateHandler(YOUR_DEFAULT));
        $h->registerSubscribingHandler(new PhpCollectionHandler()); // omit if not using 
        $h->registerSubscribingHandler(new ArrayCollectionHandler()); 
        $h->registerSubscribingHandler(new PropelCollectionHandler()); // omit if not using propel
});
$serializer = $builder->build();

xelan:
Yes, I'm using the SerializerBuilder, so your suggested solution should work. Thank you very much, @goetas!
Is there still interest in changing the default to conform with ISO-8601 later on (e.g. with serializer 2.0)? If so, I could submit my PR.

[...]

PR #714, opened 2017-02-10
Use ISO 8601 compliant datetime format as default

xelan:
| Q             | A
| ------------- | ---
| Branch?       | master / 2.0?
| Bug fix?      | yes (non-conformity to ISO 8601 standard)
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes (adapted tests)
| Related tickets | #713 

Use `DateTime::ATOM` as default in `JMS/Serializer/Handler/DateHandle` instead of `DateTime::ISO8601` to make serialized DateTime instances compatible with ISO 8601.
See http://php.net/manual/en/class.datetime.php

* Allows validation of XML with [XML schema definitions](https://www.w3.org/TR/xmlschema-2/#dateTime)
* Allows validation of JSON with [JSON schema](http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3.1)
* Increases interoperability with other software

The proposed change simplifies the usage of ISO 8601 datetime values, as no custom handler configuration is necessary anymore for standards-compliant behaviour.

I'm not so sure about an additional test for the legacy format. Maybe an additional test should be added for deserializing documents which use the previous `DateTime::ISO8601` format. Should I create an additional fixture class (like `JMS\Serializer\Tests\Fixtures\NamedDateTimeArraysObject` and add a test in `JMS\Serializer\Tests\Serializer\BaseSerializationTest`?


goetas:
fixed in https://github.com/schmittjoh/serializer/pull/743/commits/0d51077a97b82abf3e6e610654baeb29c247e851

PR #718, opened 2017-03-01
Use namespace from bottom of object stack

RLasinski:
I encountered problems with unserializing xml lists in xml lists with a default namespace from the root element.
The cause is, that you use the top element of the stack (newest one) not the bottom element (root).

I also added a test which fails with the previous implementation.

### DIFF
--- a/src/JMS/Serializer/XmlSerializationVisitor.php
+++ b/src/JMS/Serializer/XmlSerializationVisitor.php
@@ -295,7 +295,7 @@ class XmlSerializationVisitor extends AbstractVisitor
             if (null !== $metadata->xmlNamespace) {
                 $element = $this->createElement($elementName, $metadata->xmlNamespace);
             } else {
-                $defaultNamespace = $this->getClassDefaultNamespace($this->objectMetadataStack->top());
+                $defaultNamespace = $this->getClassDefaultNamespace($this->objectMetadataStack->bottom());
                 $element = $this->createElement($elementName, $defaultNamespace);
             }
             $this->setCurrentNode($element);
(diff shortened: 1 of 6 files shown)


src/JMS/Serializer/XmlSerializationVisitor.php
```diff
@@ -454,7 +454,7 @@ private function addNamespaceAttributes(ClassMetadata $metadata, \DOMElement $el
 
     private function createElement($tagName, $namespace = null)
     {
-        if (null === $namespace) {
+        if (null === $namespace || $this->document->isDefaultNamespace($namespace)) {
```
goetas: this is not always true. 

elements should be created in their namespace, if they belong to one. then the dom parser will take care of converting it into the correct xml reppresentation

goetas: what you are trying to solve, i tried it with https://github.com/schmittjoh/serializer/pull/644

src/JMS/Serializer/XmlSerializationVisitor.php
```diff
@@ -295,7 +295,7 @@ public function visitProperty(PropertyMetadata $metadata, $object, Context $cont
             if (null !== $metadata->xmlNamespace) {
                 $element = $this->createElement($elementName, $metadata->xmlNamespace);
             } else {
-                $defaultNamespace = $this->getClassDefaultNamespace($this->objectMetadataStack->top());
+                $defaultNamespace = $this->getClassDefaultNamespace($this->objectMetadataStack->bottom());
```
goetas: your tests are not covering this line

goetas:
can you also rebase your PR ?

goetas:
I went a bit deeper in the analysis. Just add 

`@XmlNamespace(uri="http://example.com/namespace")` on the `JMS\Serializer\Tests\Fixtures\ObjectList` class and your code will work.

Issue #720, opened 2017-03-03
DateInterval supported for serialization but not deserialization


c0ntax:
If you attempt to deserialize a DateInterval string to a DateInterval object, it creates a DateInterval object but ignores the actual string.

I noticed in the DateHandler class the following...

```php
    public static function getSubscribingMethods()
    {
        $methods = array();
        $deserialisationTypes = array('DateTime', 'DateTimeImmutable');
        $serialisationTypes = array('DateTime', 'DateTimeImmutable', 'DateInterval');
...
```

It seems that DateInterval is not being supported for deserialization only. Is there a particular reason for this, or is it just an omission?

PR #721, opened 2017-03-03
#720 - Adding support for DateInterval deserialization

c0ntax:
Added in DateInterval deserialization and associated unit test

Fixes https://github.com/schmittjoh/serializer/issues/720

### DIFF
--- a/src/JMS/Serializer/Handler/DateHandler.php
+++ b/src/JMS/Serializer/Handler/DateHandler.php
@@ -35,7 +35,7 @@ class DateHandler implements SubscribingHandlerInterface
     public static function getSubscribingMethods()
     {
         $methods = array();
-        $deserialisationTypes = array('DateTime', 'DateTimeImmutable');
+        $deserialisationTypes = array('DateTime', 'DateTimeImmutable', 'DateInterval');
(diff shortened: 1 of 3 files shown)


src/JMS/Serializer/Handler/DateHandler.php
```diff
@@ -173,6 +191,18 @@ private function parseDateTime($data, array $type, $immutable = false)
         return $datetime;
     }
 
+    private function parseDateInterval($data)
+    {
+        $dateInterval = null;
+        try {
+            $dateInterval = new \DateInterval($data);
+        } catch (\Exception $e) {
+            throw new RuntimeException(sprintf('Invalid dateinterval "%s", expected ISO 8601 format', $data));
```
goetas: what about adding the previous exception (`$e`) as third argument of `RuntimeException` ?

c0ntax: I was just copying the one example of catching and rethrowing an exception within the project to keep things consistent, but I'm happy to add that as a 3rd param. It makes sense.

c0ntax: Added rethrow of exception

goetas:
@c0ntax thanks!

Issue #242, opened 2014-02-24
Json array serialisation


jaymecd:
I'm kinda wondering why serialised array with empty object is represented as json object with index keys  aka "0": , "1": & etc.

``` php
$this->assertEquals('{"0":{}}', $this->serialize(array(new \stdClass())));
```

Instead of correct array output:

``` php
$this->assertEquals('[{}]', $this->serialize(array(new \stdClass())));
```

The way of encoding child elements, should not affect parent/root one.

jaymecd:
awesome :)

Issue #61, opened 2013-03-18
JSON recursion when first object in root list is empty


kriswallsmith:
To reproduce:

```
$serializer->serialize(array((object) array()), 'json');
```

The error:

> json_encode(): recursion detected

Issue #59, opened 2013-03-10
Can't serialize an array with an empty object


meze:
When I try to serialize an array that contains one empty object into json, it fails with the error:

```
PHPUnit_Framework_Error_Warning : json_encode(): recursion detected
#0 [internal function]: PHPUnit_Util_ErrorHandler::handleError(2, 'json_encode(): ...', '...', 29, Array)
#1 ...\www\vendor\jms\serializer\src\JMS\Serializer\JsonSerializationVisitor.php(29): json_encode(Object(ArrayObject), 0)
```

Here's the test:

``` php
public function testSerializeArrayWithEmptyObject()
{
    $this->assertEquals('[{}]', $this->serialize(array(new Author(null))));
}
```

Looks like it's related to https://github.com/schmittjoh/JMSSerializerBundle/issues/271

PR #730, opened 2017-03-13
Fix stdClass inconsistencies when serializing to JSON

goetas:
Fixes https://github.com/schmittjoh/serializer/issues/61
Closes https://github.com/schmittjoh/serializer/issues/59 
Closes https://github.com/schmittjoh/serializer/issues/242

### DIFF
--- a/tests/JMS/Serializer/Tests/Serializer/JsonSerializationTest.php
+++ b/tests/JMS/Serializer/Tests/Serializer/JsonSerializationTest.php
@@ -267,7 +267,7 @@ class JsonSerializationTest extends BaseSerializationTest
 
     public function testSerializeArrayWithEmptyObject()
     {
-        $this->assertEquals('{"0":{}}', $this->serialize(array(new \stdClass())));
+        $this->assertEquals('[{}]', $this->serialize(array(new \stdClass())));
     }
 
     public function testSerializeRootArrayWithDefinedKeys()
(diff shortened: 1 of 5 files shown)


tests/JMS/Serializer/Tests/Serializer/JsonSerializationTest.php
```diff
@@ -263,7 +265,85 @@ public function testSerializeWithNonUtf8EncodingWhenDisplayErrorsOn()
 
     public function testSerializeArrayWithEmptyObject()
     {
-        $this->assertEquals('{"0":{}}', $this->serialize(array(new \stdClass())));
+        $this->assertEquals('[{}]', $this->serialize(array(new \stdClass())));
```
goetas: this to me sounds a wrong test case.

goetas:
|version|time|
---|---
|master|0.044855713844299|
|pull request|0.038354706764221|

This PR gives also a 15% of speed improvement

tests/JMS/Serializer/Tests/Serializer/JsonSerializationTest.php
goetas: @schmittjoh What do you think?

robwasripped:
With the class `GenericSerializationVisitor` now deprecated, what is the recommended way to ensure that the visitor given in the `serializer.post_serialize` event contains the `setData` method?

goetas:
I really do not suggest to use `setData`. when possible is better IMO to use `vistiProperty(new StaticMedadataProperty(....))`. 

But to answer to your question, using directly `JsonSerializationVisitor` should work well. 
(BTW with the previous implementation, `setData` was fully relying on the internal `json_encode` logic)

PR #732, opened 2017-03-19
PHPdoc

JuanLuisGarciaBorrego:
Parameters type or null


goetas:
As you can see the minimum requirement for this project is not php7

PR #739, opened 2017-04-03
Use generic collection interface in handler

alcaeus:
I came across this when dealing with custom collections introduced in MongoDB ODM 1.1. The `ArrayCollectionHandler` currently only deals with selected collection classes, the ODM's `PersistentCollection` class being one of them.
With the introduction of custom collections in MongoDB ODM, not all persistent collections are of this class; instead they implement a new `PersistentCollectionInterface`, so that interface should be added to the list of collections in the handler.

When adding this, I thought there was no reason why the handler shouldn't just handle the generic collection interface defined in common, so I used that to avoid having to add new interfaces if implementations change. If you'd rather have me just add the new interface, please let me know.

### DIFF
--- a/src/JMS/Serializer/Handler/ArrayCollectionHandler.php
+++ b/src/JMS/Serializer/Handler/ArrayCollectionHandler.php
@@ -32,10 +32,7 @@ class ArrayCollectionHandler implements SubscribingHandlerInterface
         $formats = array('json', 'xml', 'yml');
         $collectionTypes = array(
             'ArrayCollection',
-            'Doctrine\Common\Collections\ArrayCollection',
-            'Doctrine\ORM\PersistentCollection',
-            'Doctrine\ODM\MongoDB\PersistentCollection',
-            'Doctrine\ODM\PHPCR\PersistentCollection',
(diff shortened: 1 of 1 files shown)


goetas:
Unfortunately you can not just remove the old declarations, the match is done using the class name as a string, not `instanceof`.

I'm fine adding `Doctrine\Common\Collections\Collection`, but you have to put back the removed lines.

goetas:
add tests please

alcaeus:
> Unfortunately you can not just remove the old declarations, the match is done using the class name as a string, not `instanceof`.

That is unfortunate, since collections in ODM are potentially generated (when using the custom collections feature) and only share an interface. I'll have to take some time to look at this - closing in the meantime.

goetas:
https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Metadata/Driver/DoctrineTypeDriver.php#L46 can be interesting for you. it is a dynamic type detection definition based on doctrine metadata.

alcaeus:
Thanks, I'll take a look!

Issue #740, opened 2017-04-03
serialize null should be a boolean in the context


goetas:
related to https://github.com/schmittjoh/serializer/pull/630

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/743/commits/387ea6c390b17007c892bef8539f0106836b3353

Issue #710, opened 2017-01-31
Serializer::handleDeserializeResult() changes the result after using Visitor::visitArray()


BartoszBartniczak:
Hi ;)

I'm trying to build handler for ArrayObject and I think I found the problem with (de-)serialization.

As the example, to create my own handler, I used the [ArrayCollectionHandler](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Handler/ArrayCollectionHandler.php). I assume that this code works.

So, here is the code of mine handler:
[link](https://github.com/BartoszBartniczak/JMSSerializerArrayObjectHandler/blob/master/src/ArrayObjectHandler.php)

```php
<?php
/**
 * Created by PhpStorm.
 * User: Bartosz Bartniczak <kontakt@bartoszbartniczak.pl>
 */

namespace BartoszBartniczak\JMSSerializer\ArrayObjectHandler;

use BartoszBartniczak\JMSSerializer\ArrayObjectHandler\Fixtures\User;
use JMS\Serializer\GraphNavigator;
use JMS\Serializer\Handler\SubscribingHandlerInterface;
use JMS\Serializer\VisitorInterface;

class ArrayObjectHandler implements SubscribingHandlerInterface
{
    public static function getSubscribingMethods()
    {
        $methods = array();
        $formats = array('json', 'xml', 'yml');
        $collectionTypes = array(
            \ArrayObject::class
        );

        foreach ($collectionTypes as $type) {
            foreach ($formats as $format) {
                $methods[] = array(
                    'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
                    'type' => $type,
                    'format' => $format,
                    'method' => 'serializeArrayObject'
                );

                $methods[] = array(
                    'direction' => GraphNavigator::DIRECTION_DESERIALIZATION,
                    'type' => $type,
                    'format' => $format,
                    'method' => 'deserializeArrayObject'
                );
            }
        }

        return $methods;
    }

    public function serializeArrayObject(VisitorInterface $visitor, \ArrayObject $arrayObject, array $type, \JMS\Serializer\Context $context)
    {
//        $type['name'] = 'array';
        return $visitor->visitArray($arrayObject->getArrayCopy(), $type, $context);
    }

    public function deserializeArrayObject(\JMS\Serializer\VisitorInterface $visitor, $data, array $type, \JMS\Serializer\Context $context)
    {
//        $type['name'] = 'array<'.User::class.'>';
[...]

Issue #691, opened 2016-12-30
Deprecate set*ContextFactory in the Serializer to keep it immutable


goetas:

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/743/commits/d5feab8af9bdd737b3b230bc1e660934bc239e2e

Issue #687, opened 2016-12-05
Remove Symfony Validator <2.6 support


goetas:

scaytrase:
Symfony 2.3 has ends its life this month, so support could be dropped

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/743/commits/8a08538335e5807cf30a5699ddd9fe27f2c3d541

PR #743, opened 2017-04-07
[WIP] 2.0

goetas:
The intent is to cleanup the core, keeping it compatible with the public api for the majority of cases.

There will be no killing feature, but mainly closing a big list of annoying bugs and unexpected behaviors that are present in the 1.x versions.
After many year, some bugs become features and many developers found workarounds that were relying on those bugs... the 2.0 release will be mainly a cleaning release.

**The goal is to allow upgrading transparently from 1.x to 2.x serializer in "standard" use-cases, including customizations as custom *subscribing handlers, event listeners, exclusion strategies*.**

## Features and bugfixes

[Here](https://github.com/schmittjoh/serializer/milestone/3) a list of features, enhancements and bugfixes that will be introduced with the 2.0 release  (not all are yet confirmed)

## Breaking changes
- Removed the abstract classes `GenericSerializationVisitor` and `GenericDeserializationVisitor`, extend the specific visitor instead.
- Added `GraphNavigatorInterface` interface, changed method signatures to use `GraphNavigatorInterface` instead of `GraphNavigator` class
- `GraphNavigator` is an abstract class now and is ony for internal proposes
- Removed deprecated method `VisitorInterface::getNavigator()` from all the visitors, use `Context::getNavigator()` instead
- Removed deprecated method `JsonSerializationVisitor::addData`, use `JsonSerializationVisitor::setData` instead
- Removed Propel and PhpCollection support
- Changed default date format from `ISO8601` to `RFC3339`
- Event listeners/handlers class names are case sensitive now
- Changed default "serializeNull" value from `NULL` to `false`
- Removed `Context::accept()` method, use `Context::getNavigator()->accept()` 
- Removed `AbstractVisitor::getNamingStrategy()` method from all the visitors
- Removed Symfony 2.x support
- Removed PHP Driver metadata support
- Removed in-object handler callbacks, use event listeners instead
- Deprecated `VisitorInterface`, use `SerializationVisitorInterface` and `DeserializationVisitorInterface` instead
- Changed `SerializerInterface::serialize` signature to `serialize($data, $format, SerializationContext $context = null, $type = null)`
- Changed `ArrayTransformerInterface::toArray` signature to `toArray($data, SerializationContext $context = null, $type = null);`
- Removed `Context::initialize`, use `DeserializationContext::initialize` and `SerializationContext::initialize`
- Removed `Serializer::setSerializationContextFactory` and `Serializer::setDeserializationContextFactory`, context factories are now constructor parameters
- Marked as `final` many classes, use composition over inheritance
- PHP 7.2 is the minimum PHP version
 
 
### In which situations transparent upgrade will not work?

Transparent upgrades probably are not possible if:
- you have extended in your application a "visitor" or implemented a custom one
- you have extended the main class `Serializer` re-implementing the `serialize` or `toArray` method
- you have re-implemented  `SerializerInterface` or `ArrayTransformerInterface`
- implemented `ObjectConstructorInterface` 
- extended one of the internal classes, now marked as final (always prefer composition over extension) (list will be available soon)

## Github issues

Doing a quick analysis of the currently  listed github "issues" for the jms serializer and jms serializer bundle i found out that: 

- 40% are support requests  (many duplicates)
- 30% are bugs or unexpected behaviors (many duplicates)
- 20% are feature requests  (some duplicates)
- Mixed stuff

The 2.0 release should close at least 40 tickets

# Todo

Issues: 

- [x] #740	serialize null should be a boolean in the context
- [x] #714	Use ISO 8601 compliant datetime format as default
- [x] #710	Serializer::handleDeserializeResult() changes the result after using Visitor::visitArray()
- [x] #691	Deprecate set*ContextFactory in the Serializer to keep it immutable
- [x] #687	Remove Symfony Validator <2.6 support
- [x] #686	Remove PHP metadata driver
- [x] #685	Prevent doctrine proxy loading for virtual types by default
- [x] #632	MaxDepth is handled incorrectly with Collections
- [x] #628	Replace deprecated getMock with createMock
- [x] #624	Event Listeners are "lowercasing" class names for event match
- [x] #623	Removed unused imports
- [x] #616	Remove PhpCollection and PhpOption and use arrays instead
- [x] #615	Remove handler callback 
- [ ] #614	Use a value object for the "type"
- [x] #613	Do not instantiate visitors in the serialization builder
- [x] #612	Change type hint for TypeParser on Serializer
- [x] #475	serializer fix
- [x] #473	GraphNavigator using \LogicException without serializer namespace
- [ ] #401	better json serialized of a formtype
- [x] #338	GraphNavigatorInterface
- [x] #295	Fixed handling of runs of capital letters in CamelCaseNamingStrategy


TNAJanssen:
What is the ETA of version 2.0?

goetas:
there is not really. 
I'm not sure that the benefits are worth the BC break.

goetas:
Wohooo! Rebased

Tobion:
thanks for working on this. On my wishlist performance improvements would come first. We use jms serializer heavily for an api. and the jms serializer is taking a huge chunk of the response time for big object graphs due to its graph navigator pattern. so if there was a way to make the serializer faster for 2.0, even if it means removing some uncommon features, that would be cool. we mostly use versioning, groups and virtual properties. i don't know if there is potential for improvement when you don't need event listeners for example. do you have any plans in this regard?

goetas:
Thanks for the feedback! Actually https://github.com/schmittjoh/serializer/pull/875 and https://github.com/schmittjoh/serializer/pull/861 bring already a ~10% of improvements already in the 1.x serie.

goetas:
Regarding your setup, to get the best performance make sure to have a look to https://github.com/schmittjoh/serializer/pull/760 and https://github.com/schmittjoh/serializer/pull/684.

They default values gives you the worst performance, this because the aim is guarantee 100% of backward compatibility, but setting:

```php
$skipVirtualTypeInit = true;
$initializeExcluded = false;
```
In some cases i got even 300% of performance improvements without doing anything else...

If you are using the symfony-bundle v2.x, then you are already benefiting the optimized values.

goetas:
blackfire
Current status :tada:

PR #744, opened 2017-04-10
Do not output the XML node when the object will be emtpy

goetas:
this will be part of 1.6.1


tests/JMS/Serializer/Tests/Fixtures/ObjectWithEmptyHash.php
```diff
@@ -24,6 +24,7 @@ class ObjectWithEmptyHash
 {
     /**
      * @Serializer\Type("array<string,string>")
+     * @Serializer\XmlList(skipWhenEmpty=false)
```
goetas: this was a buggy test. 

When arrays are empty, they are skipped by default

PR #745, opened 2017-04-13
Prevent fatal error when trying to deserialize object

stloyd:
Fatal error may occur if passed into deserialization value was object, i.e.:
```
Fatal error: Cannot use object of type Guzzle\Http\EntityBody as array in vendor/jms/serializer/src/JMS/Serializer/XmlDeserializationVisitor.php on line 378
```

### DIFF
--- a/src/JMS/Serializer/XmlDeserializationVisitor.php
+++ b/src/JMS/Serializer/XmlDeserializationVisitor.php
@@ -362,12 +362,16 @@ class XmlDeserializationVisitor extends AbstractVisitor
     /**
      * Retrieves internalSubset even in bugfixed php versions
      *
-     * @param \DOMDocumentType $child
-     * @param string $data
+     * @param mixed $data
+     *
      * @return string
      */
     private function getDomDocumentTypeEntitySubset($data)
     {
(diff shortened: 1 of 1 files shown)


goetas:
You should call cast the guzzle body do string before passing it to the `deserialize` method

stloyd:
@goetas We cannot do this as this cause this is done via guzzle services & auto serialization/deserialization, so that change is hard BC break, cause if we downgrade to version `1.0 - 1.1` everything works fine...

goetas:
Guzzle auto deserialization? Can you point me where is this implemented?

stloyd:
In fact this was hidden a little bit deeper in one of dependencies, but this still is an BC break, cause before it was allowed to put as data something different than just string (no validation), now it will just fail with deserialization if you use i.e. object with `__toString()` method.

goetas:
@stloyd  I see, probably in the `prepare()` we can just cast the input to `string`, should fix your usecase... what do you think?

goetas:
can you try https://github.com/schmittjoh/serializer/pull/755 ?

PR #748, opened 2017-04-19
#292 added option to decide to make the deserialisation strict or not

tspycher:
Serializer requires an type attribute to be able to deserialize data when a discriminator map is used. this makes not always sense if the filed only contains a referenced id to an existing object. This option enables to set the strict type checking to true/false on the discriminator map.

### DIFF
--- a/src/JMS/Serializer/Annotation/Discriminator.php
+++ b/src/JMS/Serializer/Annotation/Discriminator.php
@@ -35,4 +35,7 @@ class Discriminator
 
     /** @var string[] */
     public $groups = array();
+
+    /** @var boolean */
+    public $strictdeserialize = true;
 }
--- a/src/JMS/Serializer/GraphNavigator.php
+++ b/src/JMS/Serializer/GraphNavigator.php
@@ -210,7 +210,13 @@ final class GraphNavigator
                 }
 
                 if ($context instanceof DeserializationContext && ! empty($metadata->discriminatorMap) && $type['name'] === $metadata->discriminatorBaseClass) {
-                    $metadata = $this->resolveMetadata($data, $metadata);
+                    try {
+                        $metadata = $this->resolveMetadata($data, $metadata);
+                    } catch(\LogicException $e) {
+                        if($metadata->discriminatorStrictDeserialize) {
+                            throw $e;
+                        }
+                    }
                 }
 
                 if (null !== $exclusionStrategy && $exclusionStrategy->shouldSkipClass($metadata, $context)) {
--- a/src/JMS/Serializer/Metadata/ClassMetadata.php
+++ b/src/JMS/Serializer/Metadata/ClassMetadata.php
@@ -53,6 +53,7 @@ class ClassMetadata extends MergeableClassMetadata
     public $handlerCallbacks = array();
 
     public $discriminatorDisabled = false;
+    public $discriminatorStrictDeserialize = true;
     public $discriminatorBaseClass;
     public $discriminatorFieldName;
     public $discriminatorValue;
@@ -62,7 +63,7 @@ class ClassMetadata extends MergeableClassMetadata
     public $xmlDiscriminatorAttribute = false;
     public $xmlDiscriminatorCData = true;
 
-    public function setDiscriminator($fieldName, array $map, array $groups = array())
+    public function setDiscriminator($fieldName, array $map, array $groups = array(), $strictdeserialize = true)
     {
         if (empty($fieldName)) {
             throw new \InvalidArgumentException('The $fieldName cannot be empty.');
@@ -76,6 +77,7 @@ class ClassMetadata extends MergeableClassMetadata
         $this->discriminatorFieldName = $fieldName;
         $this->discriminatorMap = $map;
         $this->discriminatorGroups = $groups;
+        $this->discriminatorStrictDeserialize = $strictdeserialize;
     }
 
     /**
@@ -176,6 +178,10 @@ class ClassMetadata extends MergeableClassMetadata
             $this->discriminatorDisabled = $object->discriminatorDisabled;
         }
 
+        if ($object->discriminatorStrictDeserialize !== null) {
+            $this->discriminatorStrictDeserialize = $object->discriminatorStrictDeserialize;
+        }
+
         if ($object->discriminatorMap) {
             $this->discriminatorFieldName = $object->discriminatorFieldName;
             $this->discriminatorMap = $object->discriminatorMap;
@@ -251,6 +257,7 @@ class ClassMetadata extends MergeableClassMetadata
             $this->customOrder,
             $this->handlerCallbacks,
             $this->discriminatorDisabled,
+            $this->discriminatorStrictDeserialize,
             $this->discriminatorBaseClass,
             $this->discriminatorFieldName,
             $this->discriminatorValue,
@@ -279,6 +286,7 @@ class ClassMetadata extends MergeableClassMetadata
             $this->customOrder,
             $this->handlerCallbacks,
             $this->discriminatorDisabled,
+            $this->discriminatorStrictDeserialize,
             $this->discriminatorBaseClass,
             $this->discriminatorFieldName,
             $this->discriminatorValue,
--- a/src/JMS/Serializer/Metadata/Driver/XmlDriver.php
+++ b/src/JMS/Serializer/Metadata/Driver/XmlDriver.php
@@ -82,6 +82,10 @@ class XmlDriver extends AbstractFileDriver
             $discriminatorMap[(string) $entry->attributes()->value] = (string) $entry;
         }
 
+        if ('false' === (string) $elem->attributes()->{'discriminator-strict-deserialize'}) {
+            $metadata->discriminatorStrictDeserialize = false;
+        }
+
         if ('true' === (string) $elem->attributes()->{'discriminator-disabled'}) {
             $metadata->discriminatorDisabled = true;
         } elseif ( ! empty($discriminatorFieldName) || ! empty($discriminatorMap)) {
--- a/src/JMS/Serializer/Metadata/Driver/YamlDriver.php
+++ b/src/JMS/Serializer/Metadata/Driver/YamlDriver.php
@@ -284,6 +284,10 @@ class YamlDriver extends AbstractFileDriver
             if (isset($config['discriminator']['disabled']) && true === $config['discriminator']['disabled']) {
                 $metadata->discriminatorDisabled = true;
             } else {
+                if (isset($config['discriminator']['strict_deserialize'])) {
+                    $metadata->discriminatorStrictDeserialize = $config['discriminator']['strict_deserialize'];
+                }
+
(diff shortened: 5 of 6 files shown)


goetas:
without tests this will not be merged.

tspycher:
roger that. delivering tests asap

src/JMS/Serializer/GraphNavigator.php
```diff
@@ -210,7 +210,13 @@ public function accept($data, array $type = null, Context $context)
                 }
 
                 if ($context instanceof DeserializationContext && ! empty($metadata->discriminatorMap) && $type['name'] === $metadata->discriminatorBaseClass) {
-                    $metadata = $this->resolveMetadata($data, $metadata);
+                    try {
+                        $metadata = $this->resolveMetadata($data, $metadata);
+                    } catch(\LogicException $e) {
+                        if($metadata->discriminatorStrictDeserialize) {
```
goetas: what will be the content of `$metadata` if the `LogicException` is thrown? the abstract class defined on `@Type`?

goetas: Instead of catching the exception, why not improve the `resolveMetadata()` method to support your feature ?

goetas:
Generally I'm sceptic  about this PR, can you try to explain your use case?

src/JMS/Serializer/GraphNavigator.php
tspycher: $metadata is based on the definition on the root class where the descriminator is defined. So ->resolveMetadata() does simply enrich the Metadata.

Yes, your right. It would make more sense to improve ->resolveMetadata()

tspycher:
Assume i have Users and contacts and both are based on an abstract user:
```
/**
 * @Serializer\Discriminator(field = "type", map = {
 *      "user": "xxx\AppBundle\Document\User\User",
 *      "contact": "xxx\AppBundle\Document\Contact\Contact",
 * })
 */
abstract class AbstractUser {}

class User extends AbstractUser {}
class Contact extends AbstractUser {}
```

Then i have an Object like a technical Service which has technical contacts assigned of the type abstractuser. I can not simply add a new contact or user to this service as technical contact by sending the following json:

```
{
	'name': 'serviceA',
	'tech_contacts': [
		{'id': 'aabbcc11'},
		{'id': 'aabbcc22'}
	]
}
```

I have to provide the type as well which is not necessary, because the type is already given by the referenced object which is stored in the database (mongodb). 

```
{
	'name': 'serviceA',
	'tech_contacts': [
		{'id': 'aabbcc11', 'type':'user'},
		{'id': 'aabbcc22', 'type':'user}
	]
}

```
My provided change allows to change this behaviour and being able again to just provide the id of referenced data again.

tspycher:
This is the issue i get:

https://github.com/schmittjoh/JMSSerializerBundle/issues/292

goetas:
do you have a custom object constructor? how your application knows if the ID `aabbcc11` is an "user" or an "advancedUser" instance?

tspycher:
No i dont have a custom object constructor. Doctrine ODM loads the data from the database based on the AbstractUser and gets either an user oder contact object.

goetas:
i see.


have you tried just to remove `@Serializer\Discriminator` and just to typehint your `tech_contacts` as `AbstractUser` ?

tspycher:
If you would remove the Discriminator i won't longer be able to create either User or Contacts aren't I? 

I would like to send simply:

`{'type':'user', 'email':"blubb@domain.tld", 'password': "iamverysecure"}
`
or 

`{'type':'contact', 'email':"blubb@domain.tld"}`

to create new user/contact objects.

goetas:
ah, you want to have `type`  when "serializing", but you want to exclude it when deserializing... am i understanding correctly?

tspycher:
Correct! 
During serialisation it is necessary for the frontend (angular) how tondisplay the object. On deserialisation it is mandatory for object creation to know what to create but not necessary on setting an existing object as reference cause the type information ist stored in the database.

src/JMS/Serializer/GraphNavigator.php
tspycher: Change done

src/JMS/Serializer/Annotation/Discriminator.php
```diff
@@ -35,4 +35,7 @@ class Discriminator
 
     /** @var string[] */
     public $groups = array();
+
+    /** @var boolean */
+    public $strictdeserialize = true;
```
goetas: `strictDeserialize`

src/JMS/Serializer/GraphNavigator.php
```diff
@@ -292,14 +292,18 @@ private function resolveMetadata($data, ClassMetadata $metadata)
                 break;
 
             default:
-                throw new \LogicException(sprintf(
-                    'The discriminator field name "%s" for base-class "%s" was not found in input data.',
-                    $metadata->discriminatorFieldName,
-                    $metadata->name
-                ));
+                if($metadata->discriminatorStrictDeserialize) {
```
goetas: PSR

src/JMS/Serializer/GraphNavigator.php
```diff
@@ -292,14 +292,18 @@ private function resolveMetadata($data, ClassMetadata $metadata)
...
+                    throw new \LogicException( sprintf(
+                        'The discriminator field name "%s" for base-class "%s" was not found in input data.',
+                        $metadata->discriminatorFieldName,
+                        $metadata->name
+                    ) );
+                } else {
+                    $typeValue = null;
```
goetas: instead of this you can just `return $metadata;` and you will not need the changes on line 306

src/JMS/Serializer/Metadata/ClassMetadata.php
```diff
@@ -62,7 +63,7 @@ class ClassMetadata extends MergeableClassMetadata
     public $xmlDiscriminatorAttribute = false;
     public $xmlDiscriminatorCData = true;
 
-    public function setDiscriminator($fieldName, array $map, array $groups = array())
+    public function setDiscriminator($fieldName, array $map, array $groups = array(), $strictdeserialize = true)
```
goetas: `strictDeserialize`

src/JMS/Serializer/Metadata/Driver/XmlDriver.php
```diff
@@ -82,6 +82,8 @@ protected function loadMetadataFromFile(\ReflectionClass $class, $path)
             $discriminatorMap[(string) $entry->attributes()->value] = (string) $entry;
         }
 
+        $strict_deserialize = 'false' == (string)$elem->attributes()->{'discriminator-strict-deserialize'} ? false : true;
```
goetas: here is enough `'false' !== (string)$elem->attributes()->{'discriminator-strict-deserialize'}`, no `? false : true` is needed

src/JMS/Serializer/Metadata/Driver/YamlDriver.php
```diff
@@ -292,7 +292,9 @@ private function addClassProperties(ClassMetadata $metadata, array $config)
                     throw new RuntimeException('The "map" attribute must be set, and be an array for discriminators.');
                 }
                 $groups = isset($config['discriminator']['groups']) ? $config['discriminator']['groups'] : array();
-                $metadata->setDiscriminator($config['discriminator']['field_name'], $config['discriminator']['map'], $groups);
+                $strict_deserialize = isset($config['discriminator']['strict_deserialize']) ? $config['discriminator']['strict_deserialize'] : true;
```
goetas: `$config['discriminator']['strict_deserialize']` is a string, make it a boolean here

src/JMS/Serializer/Metadata/ClassMetadata.php
```diff
@@ -76,6 +77,7 @@ public function setDiscriminator($fieldName, array $map, array $groups = array()
         $this->discriminatorFieldName = $fieldName;
         $this->discriminatorMap = $map;
         $this->discriminatorGroups = $groups;
+        $this->discriminatorStrictDeserialize = $strictdeserialize;
```
goetas: cast to boolean here please

src/JMS/Serializer/Metadata/Driver/XmlDriver.php
goetas: `$strictDeserialize` per PSR rule

tests/JMS/Serializer/Tests/Fixtures/Discriminator/User.php
```diff
@@ -0,0 +1,14 @@
...
+ * Created by PhpStorm.
+ * User: tspycher
+ * Date: 20.04.17
+ * Time: 11:03
+ */
+
+namespace JMS\Serializer\Tests\Fixtures\Discriminator;
+
+
+class User  extends AbstractUser {
```
goetas: psr

tests/JMS/Serializer/Tests/Fixtures/Discriminator/User.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: tspycher
```
goetas: put a proper docblock here please

src/JMS/Serializer/GraphNavigator.php
```diff
@@ -292,14 +292,18 @@ private function resolveMetadata($data, ClassMetadata $metadata)
...
+                }
         }
 
-        if ( ! isset($metadata->discriminatorMap[$typeValue])) {
+        if ( $typeValue && ! isset($metadata->discriminatorMap[$typeValue])) {
```
goetas: PSR, remove whitespace

tspycher:
@goetas all changes finished. You need something else for this PR?

tspycher:
@goetas all changes done

goetas:
@tspycher as I can understad, this PR was about handing the case when the "type" column/property was missing... shouldn't `{"name":"john doe", "type":"user"}` be `{"name":"john doe"}` ?

tspycher:
yes and no... Yes if the object still exists somewhere in the database. I don't know how to test this usecase without having a persistence of data? To test this, i would need to have persisted data with Single Collection Inheritance.

goetas:
> yes and no... Yes if the object still exists somewhere in the database. 


this was the main propose of your PR

https://github.com/schmittjoh/serializer/blob/master/tests/JMS/Serializer/Tests/Serializer/BaseSerializationTest.php#L1272 is a test example with custom object constructor that can simulate an existing object in a storage

tspycher:
Oukay, try to add an other test

tspycher:
@goetas unittest added

goetas:
How should be handled the case when the object whit an not existent ID needs to be deserialized?

```
{"id":"some-not-existent-id"}
```

What should happen here?

goetas:
BTW, Since you do not have a `type` field... isn't simpler just use a "custom type" and decide inside the "type" of the object based on the logic you prefer?

Issue #670, opened 2016-11-03
From old issue #290


mmeloni:
We have same problem described in #290 :

ReflectionProperty::setValue() expects parameter 1 to be object, null given in .../vendor/jms/metadata/src/Metadata/PropertyMetadata.php:60

Solution proposed by @nuxwin still working

goetas:
if somebody can implement a proper test case, im open for merging it

nuxwin:
@goetas 

A test which ensure that a **EntityNotFoundException** is throw when the entity IDENTIFIER is inconsistent? There is no test covering such a case already? If so, you could just extend it by add the fallback code part as shown in #290, right?

goetas:
as said in https://github.com/schmittjoh/serializer/pull/290 you can create a new pr with a testcase for it

nuxwin:
@goetas 

Did you read my question at least? The question is: Does a testcase already exists for ensuring that **EntityNotFoundException** is raised ? I would be happy to provide a PR, including a testcase but I would be also happy to known the purpose of that testcase. You known your code better than us.

BTW: We could also answer you: There is a bug regarding this. Please, fix it. You're the maintainer after all ;)

Thank you for your understanding.

goetas:
@nuxwin please read the [License](https://github.com/schmittjoh/serializer/blob/master/LICENSE) provided with this software and its terms and conditions  (especially sections 7,8,9)

nuxwin:
@goetas 

Funny answer (ironic)....

To be more serious, If all open-source community acted like you, nobody would use an open-source software... Your answer here is really an asshole' answer. If you have not more interest for that library, just go ahead but don't provide such answer which only discredits the the project.

@schmittjoh  

I don't even know how such a guy can be a collaborator on this project.

Issue #492, opened 2015-09-10
Deserialization of sub entites with non existing id


feridmovsumov:
When I deserialize a json to entity and there is sub entity with id that doesn't exist on database. serializer set sub entity as null. 

Are there any way to throw exception when serializer can't find sub object instead of giving null?

Issue #216, opened 2014-01-16
Warning when deserializing removed entity


ddeboer:
When a Doctrine entity is serialized, removed from the database and then  deserialized, a warning is given:

```
Warning: ReflectionProperty::setValue() expects parameter 1 
to be object, null given in /vendor/jms/serializer/src/JMS/Serializer/GenericDeserializationVisitor.php on line 176
```

The entity is then deserialized as an integer (its original id) instead of an object. The relevant code is [DoctrineObjectConstructor.php](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Construction/DoctrineObjectConstructor.php#L86): if the entity is not found, `null` is returned. 

Simply adding the following helps, but means that you can end up with entities that are no longer in the database:

``` php
// Entity update, load it from database
$object = $objectManager->find($metadata->name, $identifierList);
if (!$object) {
    return $this->fallbackConstructor->construct($visitor, $metadata, $data, $type, $context);
}
```

What would be the intended behaviour in this case?

magnetik:
Facing the same problem

brettminnie:
@magnetik have a proposed fix sitting here #290 Hopefully it will help you until it's addressed

brettminnie:
@magnetik have a look here #290 this is how we solved it

ignace:
I used the serializer component for an API. The client sends an entity with an identifier, which makes the doctrineobjectconstructor confuse it with an existing entity. Doing the above solves this problem and the new entity is constructed as expected as are any associated entities (in my case a parent object "parent": {"id": 1}}).

nuxwin:
See my answer here for alternative solution: https://github.com/schmittjoh/serializer/pull/290#issuecomment-159778071

mmeloni:
Same problem

PR #751, opened 2017-04-25
Allow to configure the doctrine object constructor

goetas:
Allows to decide what should happen if an entity is not found in the database. Possible options:

- `ON_MISSING_NULL`: return null (current behavior)
- `ON_MISSING_EXCEPTION`: throw an excpetion
- `ON_MISSING_FALLBACK`: fallback to the standard object constructor

Closes https://github.com/schmittjoh/serializer/issues/216
Closes https://github.com/schmittjoh/serializer/issues/492
Closes https://github.com/schmittjoh/serializer/issues/670


courtney-miles:
I would like to take advantage of these options but I can find any documentation on how to change these options via a dcm file.

I would appreciate if someone can describe how might I apply one of these options?

PR #757, opened 2017-04-27
Allow to skip "empty serialization result" when serializing

goetas:
Fixes BC break introduced with #744 (and released with 1.6.1) by allowing to configure the serializer behavior when dealing with "empty" serialization result

[here](https://travis-ci.org/goetas-webservices/xsd2php/builds/221891989) some test suite failing with 1.6.1 & 1.6.2 while working with <=1.6.0 & this branch

"empty" means in this context: excluded by some exclusion-strategy, by nulls or empty objects with no properties


Webonaute:
in JsonSerializationVisitor.php
line 162. It check for \ArrayObject. but if $v is a normal array, it doesnt skip it.

goetas:
@Webonaute thanks! Fixed with https://github.com/schmittjoh/serializer/commit/34f5b67714c3925fba96e3d18ae2ed134f58cf57

Tobion:
@goetas doesn't this fix #124?

goetas:
good catch! thanks!

src/JMS/Serializer/JsonSerializationVisitor.php
```diff
@@ -158,7 +158,9 @@ public function visitProperty(PropertyMetadata $metadata, $data, Context $contex
         $v = $this->accessor->getValue($data, $metadata);
 
         $v = $this->navigator->accept($v, $metadata->type, $context);
-        if (null === $v && $context->shouldSerializeNull() !== true) {
+        if ((null === $v && $context->shouldSerializeNull() !== true)
+            || ($v instanceof \ArrayObject && 0 === count($v) && $metadata->skipWhenEmpty)
```
Tobion: So a plain empty array is not skipped as it checks for \ArrayObject? That's what I hoped for with #124

goetas:
It works for arraya too

See https://github.com/schmittjoh/serializer/commit/34f5b67714c3925fba96e3d18ae2ed134f58cf57

PR #770, opened 2017-05-15
[RFC] Move to PSR 4

goetas:
Is this a BC break?


goetas:
Considering some twitter feedback https://twitter.com/goetas_asmir/status/864084687618277376

Issue #772, opened 2017-05-21
Allow access to PropertyMetadata in Dynamic Exclusion strategies


bblue:
Is this possible to access the propertyMetada in via `$navigatorContext`, or can it be added as a variable in `ExpressionLanguageExclusionStrategy`. 

The `metadata` can be added as a variable in the same was as the `context `is.
https://github.com/schmittjoh/serializer/blob/master/src/Exclusion/ExpressionLanguageExclusionStrategy.php#L56

goetas:
the navigator is already available via `context.getNavigator()`, regarding the metadata, PR are welcome

PR #776, opened 2017-05-22
Add propertyMetdata to dynamic expression variables

bblue:
Fix for https://github.com/schmittjoh/serializer/issues/772 

This makes  `PropertyMetatadata $property`  available for dynamic exclusion strategies, similar to what would be expected for standard exclusion strategies, as defined by the [ExclusionStrategyInterface](https://github.com/schmittjoh/serializer/blob/master/src/Exclusion/ExclusionStrategyInterface.php#L48).

Example:

```
class MyClass
{
    /**
     * @Serializer\Expose(if="someMethod(property_metadata)")
     */
    protected $myVariable;
}
```

`SomeMethod` would now recieve the metadata for the property.

### DIFF
--- a/src/JMS/Serializer/Annotation/ExclusionPolicy.php
+++ b/src/JMS/Serializer/Annotation/ExclusionPolicy.php
@@ -27,13 +27,13 @@ use JMS\Serializer\Exception\RuntimeException;
 final class ExclusionPolicy
 {
     const NONE = 'NONE';
-    const ALL  = 'ALL';
+    const ALL = 'ALL';
 
     public $policy;
 
     public function __construct(array $values)
     {
-        if ( ! is_string($values['value'])) {
+        if (!is_string($values['value'])) {
             throw new RuntimeException('"value" must be a string.');
         }
 
--- a/src/JMS/Serializer/Annotation/XmlElement.php
+++ b/src/JMS/Serializer/Annotation/XmlElement.php
@@ -28,7 +28,7 @@ final class XmlElement
      * @var boolean
      */
     public $cdata = true;
-    
+
     /**
      * @var string
      */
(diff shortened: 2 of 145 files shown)


goetas:
Hi . Thanks for the PR.

Can you add some tests? Without, this can't be merged

src/Exclusion/ExpressionLanguageExclusionStrategy.php
```diff
@@ -54,6 +54,7 @@ public function shouldSkipProperty(PropertyMetadata $property, Context $navigato
 
         $variables = [
             'context' => $navigatorContext,
+            'property_metadata'=> $property
```
goetas: Please add semicolon

mvrhov: you probably meant comma **,** , semicolon is **;** :)

goetas: Yes... My Bad english :-/

bblue:
I was looking for an existing test for the `context` variable, hoping to piggy-back on that one, but could not find anything. Are there any such tests, or will I have to create some from scratch?

goetas:
there is not yet a dedicated test for it, but it is a good idea to add it. you can add it to https://github.com/schmittjoh/serializer/tree/master/tests/Exclusion calling it maybe `ExpressionExclusionStrategyTest`... whdyt?

bblue:
How's that? I had to use the `DeserializationContext` to avoid the context blowing up when the strategy queries for the `object` variable [here](https://github.com/bblue/serializer/blob/cd9ce57bfed407921676e1759bbff29bcbfe10a0/src/Exclusion/ExpressionLanguageExclusionStrategy.php#L60). As such, the  `object` is not tested against, but the `context` and `property_metadata` is, as well as if the exclusion strategy actually works (simple true/false test). 

From what I could tell the `context` would need to be initialized of the `object` were to be tested, and a whole lot of other dependencies would need to be created. It felt somewhat out of scope for this PR.

goetas:
@bblue 
For a reason that I still do not understand, while pushing my commit to your branch, github removed commits and closed the pr.... I have created a copy of it in #778

I have also rebased it, and changed your unit test with a simpler one... (does not require the deserialization context, and properly tests the presence of `property_metadata` variable in the context [your tests was not doing it, event if the main propose of the test was it])

moving the conversation there

PR #778, opened 2017-05-24
Add propertyMetdata to dynamic expression variables

goetas:
copy of https://github.com/schmittjoh/serializer/pull/776


bblue:
Your tests looks a lot better. I don't have much experience writing tests, so happy for the learning. 

However, I am reading the new tests on my phone now, and need to have a more thorough review later today. At first sight I don't see how your tests confirms the presence of the variables either, but let me have a look again.

bblue:
Took me a minute to figure out what was going on there, but that was indeed a lot cleaner than the hack I tried at. Thanks for supporting.

I suppose that's it then.

Issue #79, opened 2013-04-14
Allow Constructed Object to be Passed to Deserialize


cmorelli:
It doesn't appear that there's a way to give an already constructed object to the deserialize method (in which case the serializer would just skip the object construction phase and begin mapping properties).

My use case is simple:
In a REST API, I want to allow a user to create or update objects. Each type of operation (create, update, delete, read) has a different set of security rules. However, with the current implementation of the deserializer (in which we are using the Doctrine object constructor), the user can issue a create request, but specify an "id" as part of the payload. The Doctrine object constructor will then return a mapped entity before deserialization begins, allowing the end user to perform an update (while the server still thinks it's doing a create).

Also, on this point:
I was considering making a separate object constructor - but I need to give some attributes to the object constructor on each use (such as the ID which it should construct an object for). Right now, the only way to get those attributes to the deserializer is to include them in the payload string that gets passed to it - which is precisely what I'd like to avoid.

There are ways to handle this in the actual REST endpoint, but none of them are really ideal.What I'd like to be able to do is pass an already-constructed object (which my REST endpoint will handle fetching) to the deserializer and just have it do it's property mapping onto the object. This way, I can always pass a blank object in create calls, while I can pass a constructed object in update calls.

Would this be possible? I could submit a PR if this is something you'd be willing to implement.

terwey:
I followed @eschwartz recommendation and just did a

```bash
cp vendor/jms/serializer/tests/JMS/Serializer/Tests/Fixtures/InitializedObjectConstructor.php src/myApp/
```
Modified the namespace in the ```InitializedObjectConstructor.php``` and it all works like a charm after this.

Can this class be moved into a more proper and permanent location so I don't have to maintain it myself?

RedEdgeKatelyn:
For anyone trying to do this in Symfony, I copied `vendor/jms/serializer/tests/JMS/Serializer/Tests/Fixtures/InitializedObjectConstructor.php` to a Services folder in my Bundle and added this to config.yml:

    services:
        jms_serializer.object_constructor:
            alias: jms_serializer.initialized_object_constructor
            public: false
        jms_serializer.initialized_object_constructor:
             class:        VendorName\Bundle\ApiBundle\Services\InitializedObjectConstructor
             arguments:    ["@jms_serializer.doctrine_object_constructor"]

faheemhameed:
Note about the Symfony helped a lot. Thanks!

tom10271:
[...]

Issue #775, opened 2017-05-22
[2.0] [Feature-request] Provide InitializedObjectConstructor as default


scaytrase:
It is really hard to understand that you cannot to deserialize into existing object without additional configuration or using doctrine

Also this is weird that such useful thing lives within fixtures.

If it is applicable for 1.x branch (since it is somehow backward compatible), it would be nice

goetas:
`InitializedObjectConstructor` is just a specific usecase of the of the `ObjectConstructorInterface`.

`InitializedObjectConstructor` works only on the first element of the deserialization object graph, and in this case is more convenient just for test proposes.

Since `InitializedObjectConstructor` works only for a specific usecase, and has a lot fo drawbacks, I do not plan to integrate it into the core. If your usecase is similar to https://github.com/schmittjoh/JMSSerializerBundle/issues/575, i suggest you to have a look at the symfony/form component and the approach suggested in https://github.com/schmittjoh/JMSSerializerBundle/issues/575#issuecomment-303058694

scaytrase:
@goetas the problem of using forms is that you have to write form classes instead of describing the serialized presentation of entities. It's like suggestion to manually map DB data to entities but not using the Doctrine2 hydration metadata

goetas:
I know than it is not an optimal solution... at the moment the serializer is not really powerful to update objects (as example, can not  easily append items  to a collection or deal with different data-formats where normalization is needed).

On the other hand, symfony form is really good at it, but does not offer serialization, that is why I like to combine the two of them.

The provided object constructors are domain independent...
 - `UnserializeObjectConstructor` creates an object instance for any class,  using PHP reflection
 - `DoctrineObjectConstructor` creates/retrieves an object instance from the doctrine registry of managed entities 

-----
`InitializedObjectConstructor` is really domain-specific and hacky (it works only for the first object of the deserialization process...). Requires you to pass the instance via context, making your code coupled to the construction logic.
If you are really interested in working on a similar solution, something more interesting can be a registry of constructed objects.... as:

```php
class InstantiatedObjectsObjectConstructor implements ObjectConstructorInterface
{

    private $registry = array();
    private $fallbackConstructor;

    public function __construct(ObjectConstructorInterface $fallbackConstructor)
    {
        $this->fallbackConstructor = $fallbackConstructor;
    }

    public function add($id, $object)
    {
        $this->registry[$id] = $object;
    }

[...]

PR #786, opened 2017-06-16
move InitializedObjectConstructor to src directory

mimol91:
This change allows to use `InitializedObjectConstructor`  which could be used to serialize into existing object (https://github.com/schmittjoh/serializer/issues/79).

For production setup test directory is mostly excluded from composer class-map which disallow use this class.


goetas:
see https://github.com/schmittjoh/serializer/issues/775

PR #787, opened 2017-06-17
Added deserialize type

bobvandevijver:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/JMSSerializerBundle/issues/582
| License       | Apache-2.0

This PR adds the possibility to have a different deserialization type than the original type. This can be useful for DateTime serialization, when you might expect an all zero time as result (see linked issue).

PR #788, opened 2017-06-17
Added support for a third deserialize parameter for the DateTime type

bobvandevijver:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/JMSSerializerBundle/issues/582
| License       | Apache-2.0

This PR adds an third parameter to the DateTime format for the Type annotation, allowing to use a different deserialization type. This can be useful when only serializing dates in a custom format and you need to have the other fields all zeroed.

### DIFF
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -357,7 +357,9 @@ Available Types:
 | DateTime<'format', 'zone'>                               | PHP's DateTime object (custom format/timezone)   |
 +----------------------------------------------------------+--------------------------------------------------+
 | DateTime<'format', 'zone', 'deserializeFormat'>          | PHP's DateTime object (custom format/timezone,   |
-|                                                          | deserialize format)                              |
+|                                                          | deserialize format). If you do not want to       |
+|                                                          | specify a specific timezone, use an empty        |
+|                                                          | string ('').                                     |
 +----------------------------------------------------------+--------------------------------------------------+
 | DateTimeImmutable                                        | PHP's DateTimeImmutable object (default format/  |
 |                                                          | timezone)                                        |
@@ -369,7 +371,9 @@ Available Types:
 |                                                          | timezone)                                        |
 +----------------------------------------------------------+--------------------------------------------------+
 | DateTimeImmutable<'format', 'zone', 'deserializeFormat'> | PHP's DateTimeImmutable object (custom format/   |
-|                                                          | timezone/deserialize format)                     |
+|                                                          | timezone/deserialize format). If you do not want |
+|                                                          | to specify a specific timezone, use an empty     |
+|                                                          | string ('').                                     |
 +----------------------------------------------------------+--------------------------------------------------+
 | DateInterval                                             | PHP's DateInterval object using ISO 8601 format  |
 +----------------------------------------------------------+--------------------------------------------------+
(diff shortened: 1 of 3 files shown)


bobvandevijver:
Relates to #787, both will fix the problem I experience in https://github.com/schmittjoh/JMSSerializerBundle/issues/582.

src/Handler/DateHandler.php
```diff
@@ -175,8 +175,8 @@ public function deserializeDateIntervalFromJson(JsonDeserializationVisitor $visi
 
     private function parseDateTime($data, array $type, $immutable = false)
     {
-        $timezone = isset($type['params'][1]) ? new \DateTimeZone($type['params'][1]) : $this->defaultTimezone;
-        $format = $this->getFormat($type);
+        $timezone = isset($type['params'][1]) && $type['params'][1] != "" ? new \DateTimeZone($type['params'][1]) : $this->defaultTimezone;
```
goetas: you can just use `!empty` instead of `isset`

doc/reference/annotations.rst
```diff
@@ -324,57 +324,63 @@ to the least super type::
...
+|                                                          | array<string, MyNamespace\MyObject>, etc.        |
++----------------------------------------------------------+--------------------------------------------------+
+| DateTime                                                 | PHP's DateTime object (default format/timezone)  |
++----------------------------------------------------------+--------------------------------------------------+
+| DateTime<'format'>                                       | PHP's DateTime object (custom format/default     |
+|                                                          | timezone)                                        |
++----------------------------------------------------------+--------------------------------------------------+
+| DateTime<'format', 'zone'>                               | PHP's DateTime object (custom format/timezone)   |
++----------------------------------------------------------+--------------------------------------------------+
+| DateTime<'format', 'zone', 'deserializeFormat'>          | PHP's DateTime object (custom format/timezone,   |
```
goetas: can you specify that is optional? (same for zone since we are editing it)

src/Handler/DateHandler.php
```diff
@@ -206,9 +206,13 @@ private function parseDateInterval($data)
     /**
      * @return string
      * @param array $type
+     * @param bool $deserialize
      */
-    private function getFormat(array $type)
+    private function getFormat(array $type, $deserialize = false)
     {
+        if ($deserialize && isset($type['params'][2])){
```
goetas: space before `{` (PSR standards)

src/Handler/DateHandler.php
```diff
@@ -206,9 +206,13 @@ private function parseDateInterval($data)
...
+          return $type['params'][2];
```
goetas: 4 spaces indent

goetas:
Very appreciated this PR :smile: , I also had to use some workarounds to solve a similar issue

bobvandevijver:
Solved your review notes, and added a simple test. I'm not really experienced with tests, so when it's not sufficient, maybe you can take a look?

goetas:
@bobvandevijver thanks for your PR. (If you are interested have a look at my version of the testcase https://github.com/schmittjoh/serializer/blob/b92cf25987d0f2bae0817476a117f8b2811a54ef/tests/Handler/DateHandlerTest.php)

PR #790, opened 2017-06-23
Detect XML xsi:nil="true" to null when deserializing

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | -
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/serializer/pull/771
| License       | Apache-2.0

This is an alternative implementation of #771 

The main difference is that the "null aware" interface is a visitor.

### DIFF
--- a/src/XmlDeserializationVisitor.php
+++ b/src/XmlDeserializationVisitor.php
@@ -25,7 +25,7 @@ use JMS\Serializer\Exception\XmlErrorException;
 use JMS\Serializer\Metadata\ClassMetadata;
 use JMS\Serializer\Metadata\PropertyMetadata;
 
-class XmlDeserializationVisitor extends AbstractVisitor implements NullEvaluatorInterface
+class XmlDeserializationVisitor extends AbstractVisitor implements NullAwareVisitorInterface
 {
     private $objectStack;
     private $metadataStack;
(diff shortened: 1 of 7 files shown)


goetas:
@Bouwdie  what do you think about this? 

the nullevaluator interface make no sense as it was, it is an extension of the visitor behavior, so I extended it.

Bouwdie:
@goetas I like the changes you made! 👍  It does make more sense to hook the evaluation in the already existing visitor structure; making it more transparent. From my side it is a go.

src/XmlDeserializationVisitor.php
```diff
@@ -387,4 +387,23 @@ private function getDomDocumentTypeEntitySubset($data)
...
+     *
+     * @return bool
+     */
+    public function isNull($value)
+    {
+        if ($value instanceof \SimpleXMLElement) {
+            $xsiAttributes = $value->attributes('http://www.w3.org/2001/XMLSchema-instance');
+
+            //We have to keep the isset quiet, some tests give error: `Node no longer exists`; even though it evaluates to false
+            if (@isset($xsiAttributes['nil']) && (string) $xsiAttributes['nil'] === 'true') {
```
Tobion: Booleans in xml can also be represented as 1 and 0. See https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/datatypes.html#boolean So this check doesn't work in this case.

goetas: hmmm! true! PR are welcome!

Bouwdie: Completely true; PR incoming.. 👍

PR #794, opened 2017-07-04
Use AbstractVisitor for type hinting DateHandler

paulinevos:
instead of the concrete implementations (XmlDeserializationVisitor,
JsonDeserializationVisitor). This offers more flexibility, as using
custom deserialization visitors that extend the abstract class causes
problems now.
Also updated test to use a mock of the AbstractVisitor class instead
of the conrete implementations.

| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  |no
| Doc updated   |no
| BC breaks?    |no
| Deprecations? |no
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | Apache-2.0

### DIFF
--- a/src/Handler/DateHandler.php
+++ b/src/Handler/DateHandler.php
@@ -18,12 +18,11 @@
 
 namespace JMS\Serializer\Handler;
 
+use JMS\Serializer\AbstractVisitor;
 use JMS\Serializer\Context;
 use JMS\Serializer\Exception\RuntimeException;
 use JMS\Serializer\GraphNavigator;
-use JMS\Serializer\JsonDeserializationVisitor;
 use JMS\Serializer\VisitorInterface;
-use JMS\Serializer\XmlDeserializationVisitor;
 use JMS\Serializer\XmlSerializationVisitor;
 
 class DateHandler implements SubscribingHandlerInterface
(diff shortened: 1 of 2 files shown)


paulinevos:
Sidenote: please note the `$visitor` arguments aren't even used in any of the methods touched, but still left them in as to not break BC for users.

goetas:
Hi! thanks for the pull request!

changing the signature of a public method can be considered a BC break (see https://3v4l.org/KKnOq), so this PR can't be accepted unless it solves some bigger issue to justify a possible BC break for users.


> Side note: please note the $visitor arguments aren't even used in any of the methods touched, but still left them in as to not break BC for users.

This can't be done anyway. Since parameters are positional, and the method is called by the event manager, parameter order is fundamental, and can't be changed without BC breaks.

goetas:
closing

paulinevos:
Hi @goetas, can you tell me how it would be a BC break when both classes typehinted were replaced by the abstract class they extend and they're not used in the method implementations?

goetas:
if somebody extended any of the classes you have modified, will get the error.

I'm pretty sure that somebody did it. (not the datetime, but in this case FOSRestBundle extended the form handler  https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Serializer/Normalizer/FormErrorHandler.php) that is the same case

paulinevos:
@goetas Ah right, I see what you mean now. Thanks.

PR #802, opened 2017-08-02
Added mapping for guid type

develth:
I recognized that my Entites ids, that have alle the type 'guid' are not serialized and shown in the, from NelmioApiDocBundle generated, documentation, but it always serialized in e.g. json output. 

So i dived into the parsers and came up to this file, where guid was not in the fieldMapping 

This little commit gives the possibility to map the guid type, too.


goetas:
hi, thanks for the PR! can you add some tests?

develth:
Hi @goetas, 
for sure - its done.

goetas:
thanks!

PR #808, opened 2017-08-13
added @required annotation as well as tests

michaelrios:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | Apache-2.0

### DIFF
--- a/src/JMS/Serializer/Annotation/Required.php
+++ b/src/JMS/Serializer/Annotation/Required.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * Copyright 2013 Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


mjgorman:
+1

goetas:
Hi. Thanks for the PR! 
Quick question: why do not use a validator as symfony/validator?

michaelrios:
This just fits what I need, and without adding any overhead to existing processes. I also do not need any other features in the symfony/validator package, so that would add a bit more bloat to my application.

goetas:
Hi @michaelrios thanks again for the PR. 
Unfortunately can not accept the PR (even from the code point of view looks good).

This serializer library focuses on serialization/deserialization; normalization and validation are sometimes overlapping tasks, but IMO fundamentally different (and already well covered by way better alternatives as symfony form or symfony validator).

PR #810, opened 2017-08-16
Use fallback in DoctrineObjectConstructor when identifier is null

matthias-z:
In the environment I'm currently working in, the incoming data to be deserialized always includes the identifier field, even if the value is `null` and thus no database record exists.

In such cases the `DoctrineObjectConstructor` is calling the `find` method from the repository regardless, even though the identifiers (primary key) can never be null (at least in all DBMS I've worked with and according to standard SQL).

This will inevitably cause an exception to be raised in Doctrine (see [EntityManager.php](https://github.com/doctrine/doctrine2/blob/a0c0d3bf2a7f49020ad7e54d7d3319a251574230/lib/Doctrine/ORM/EntityManager.php#L404)).

This pull request extends the current behaviour to treat identifiers with `null` values the same way as if they were not given in the data array at all.

### DIFF
--- a/src/JMS/Serializer/Construction/DoctrineObjectConstructor.php
+++ b/src/JMS/Serializer/Construction/DoctrineObjectConstructor.php
@@ -87,7 +87,7 @@ class DoctrineObjectConstructor implements ObjectConstructorInterface
         $identifierList = array();
 
         foreach ($classMetadata->getIdentifierFieldNames() as $name) {
-            if (!array_key_exists($name, $data)) {
+            if (!isset($data[$name])) {
                 return $this->fallbackConstructor->construct($visitor, $metadata, $data, $type, $context);
             }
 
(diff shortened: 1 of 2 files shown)


goetas:
Hi, thanks for the PR.

This is a BC break and can't be merged as it is. 
A configurable behavior would be much better. See https://github.com/schmittjoh/serializer/pull/751 for an example implementation.

matthias-z:
Thanks for the quick feedback!.

However, I fail to see how this is a BC break when the current behaviour is going to cause an exception anway.

Am I missing any use cases, where this would not be the case?

goetas:
> However, I fail to see how this is a BC break when the current behaviour is going to cause an exception anway.

the current behavior (the exception) may be handled by some user to achieve some goal, removing this behavior, their code will break.

matthias-z:
Personally, I expected the **Doctrine**ObjectConstructor to comply with Doctrine in this regard, so I'd rather consider this a bug fix.

I'm happy to accept your policies though and will work around it in my project instead.

Thank you for your time.

PR #812, opened 2017-08-19
Groups doctrine deserialization

bertuz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | Apache-2.0

Doctrine's ObjectConstructor ignored the Context groups that can be specified when deserializing. An identifier could be sent and it was used to find an entity, irrespective of its group.  This is not a problem if groups are not used while deserializing, but can cause big security concerns if groups are used to not let the user send data that could affect persisted data.
The pull request adds some tests to show it and it tries to fix it without altering the other functionalities.

### DIFF
--- a/src/JMS/Serializer/Construction/DoctrineObjectConstructor.php
+++ b/src/JMS/Serializer/Construction/DoctrineObjectConstructor.php
@@ -135,20 +135,23 @@ class DoctrineObjectConstructor implements ObjectConstructorInterface
         /** @var array $deserializingGroups */
         $deserializingGroups = $context->getGroups()->getOrElse(array());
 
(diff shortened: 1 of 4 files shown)


vwatel:
Hi @goetas 

Do you have any news on this pull request?
According to Veracode, this is the possible security patch for an existing vulnerability within the library.

Thank you.

goetas:
Hi, Unfortunately this pr is not implemented on a way that makes it ready to be merged. It works only for annotations (not for yaml or xml metadata). Do you want to resume the work on it?

Issue #826, opened 2017-10-11
Dynamic exclusion strategy, Variable "object" is not valid


InfopactMLoos:
| Q                | A
| ---------------- | -----
| Bug report?      | not sure?
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Steps required to reproduce the problem

1. Enable Expression Language exclusion strategy as per: [https://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#dynamic-exclusion-strategy](https://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#dynamic-exclusion-strategy)
2. Use the object variable in a expression within the Expose annotation on a class property, like so:
```
@Serializer\Expose(if="object.isRedacted() === false")
```
3. Call deserialize on a string using this class, like so:
```
$serializer->deserialize($requestBody, 'MyClass', 'json');
```
4. Receive exception:
```
Variable \"object\" is not valid around position 3 for expression `!(object.isRedacted() === false)`.
```

## Expected Result

* I expect a successful deserialization.

## Actual Result

* Serialization failed with above exception message. As I looked closer at the code it seems that the object variable is only set when there is a SerializationContext. Meaning ofcourse that this variable is not available during deserialization. You can find this in class JMS\Serializer\Exclusion\ExpressionLanguageExclusionStrategy line 60. 
Although this might be quite logical since the object is still being created (and I actually don't need this expression to be evaluated during deserialization), however how do I only enable it on serialization? Or actually what would be the best way to do this? I can ofcourse extend the expression but since I have this expression on like 106 properties I would like to keep these expressions as simple and clean as possible. Or maybe I should write a custom exclusion strategy...

goetas:
You are right. This is an unexpected behavior of the current implementation.

A possible workaround for this can be 

`@Serializer\Expose(if="context.getDirection() === 1 && object.isRedacted() === false")`.

Actually this solution will require anyway to update the `ExpressionLanguageExclusionStrategy.php`  with

```php
L59
if ($navigatorContext instanceof SerializationContext) {
        $variables['object'] = $navigatorContext->getObject();
} else {
        $variables['object'] = null;
}
```


This solution might work but I guess there are better solutions... ideas?

InfopactMLoos:
Yes I indeed already tried the workaround (forgot to mention) but the variable is always evaluated. Your commit seems promising though I will have a look at it. 
[...]

PR #827, opened 2017-10-16
Alow to use "object" var in expressions when deserializing

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/serializer/issues/826
| License       | Apache-2.0

### DIFF
--- a/src/JMS/Serializer/Exclusion/ExpressionLanguageExclusionStrategy.php
+++ b/src/JMS/Serializer/Exclusion/ExpressionLanguageExclusionStrategy.php
@@ -58,6 +58,8 @@ class ExpressionLanguageExclusionStrategy
         ];
         if ($navigatorContext instanceof SerializationContext) {
             $variables['object'] = $navigatorContext->getObject();
+        } else {
+            $variables['object'] = null;
         }
 
         return $this->expressionEvaluator->evaluate($property->excludeIf, $variables);
--- a/tests/Fixtures/PersonSecretWithVariables.php
+++ b/tests/Fixtures/PersonSecretWithVariables.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * Copyright 2016 Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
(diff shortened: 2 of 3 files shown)


mpoiriert:
@goetas Why does the object is not available and must be set to null in a Deserialization context ?

Looking at the DeserializationGraphNavigator the object is available.

This cause some issue because if we want to do some security to allow edition of some attribute base on the currently connected user we can't. 

Since the visitor is accessible from the context and that the current object could bet access from getCurrentObject we can have a work arround but the getCurrentObject is not part of the DeserializationVisitorInterface or VisitorInterface, both XMl and Json deserialization implements the method maybe it should be part of it ?

goetas:
Actually the object sometimes is available. 
- When doing the per-class exclusion, the object is not available as it has not yet been created (see https://github.com/schmittjoh/serializer/blob/8727285749a23ca3b3154b1932f07e69289583a4/src/GraphNavigator/DeserializationGraphNavigator.php#L176, the object gets created few lines later)
- When doing the per-property exclusion the object is available, but is in a non consistent state, some props are deserialized and some not. 

Given this contradictory situations, was much simpler to not make it available. 
Can you please describe better your usecase?

mpoiriert:
I understand for the shouldSkip class, it make sens since it's about the class...

For the state of the object I thought about it and there is 2 way that I came up with:
- **Not great** Clone the object and work on the clone itself for the test
- **Best way** Loop trough all the property first just to check which property should be updated and then loop again to do the visit.

The use case itself is to have ACL on a **object** property level and not on a **class** property level.

I could submit a PR of the second solution if you want.

mpoiriert:
We will also need to come up with a none BC way to make sure that if some developer have done a check if object is null to apply some logic it will still work. 

I thought of a few way but I don't want to make configuration too much complicated either...

mpoiriert:
Any interest in me making a PR of the second solution ? (I still need to implement it)

PR #829, opened 2017-10-29
Fixed issue where timezone is lost when creating DateTime from unix timestamp

veloxy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | /
| License       | Apache-2.0

Timezone is lost when creating a `DateTime` from a unix timestamp, quote from the [PHP docs](http://php.net/manual/en/datetime.construct.php)

> The $timezone parameter and the current timezone are ignored when the $time parameter either is a UNIX timestamp (e.g. @946684800) or specifies a timezone (e.g. 2010-01-28T15:00:00+02:00).

To fix this, I re-set the timezone after creating the `DateTime` object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,20 @@
 # Change Log
 
-## [1.9.0](https://github.com/schmittjoh/serializer/tree/1.9.0)
+## [1.9.1](https://github.com/schmittjoh/serializer/tree/1.9.1)
 
+**Fixed bugs:**
+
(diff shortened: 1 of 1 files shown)


goetas:
hmm, here github did it wrong. here is the PR https://github.com/schmittjoh/serializer/pull/835

Issue #832, opened 2017-11-02
Allow generators as return type for SubscribingHandlerInterface::getSubscribingMethods


ruudk:
In my project I use a `SubscribingHandlerInterface`. The `getSubscribingMethods` should return an array, but since my project has lots of serializers (of the same type, just other class) it becomes a very large array. 

My idea was to use a [Generator](http://php.net/manual/en/language.generators.overview.php) for this and simply `yield` array elements when I generate them.

This works great.

Only thing is that the phpdoc on `SubscribingHandlerInterface::getSubscribingMethods` says that it should return an array.

Can we extend this to be a Generator as well? We could add a test that would allow it, so that we won't break it in future releases.

goetas:
Generators are nice, from what I know also symfony is using generators in some special cases. 
We can introduce something as `interface AdvancedSubscribingHandlerInterface extends SubscribingHandlerInterface` so it will backward compatible...

I'm just curios what will be the performance benefit... since on the first "serialize" call, all the elements will have to be initialized anyway.

ruudk:
Well, it's not really for performance. It's just that I can write cleaner code. I don't have to first create an array, foreach and store everything in it, and then return it.
I can just foreach and yield as I go.

I saw a `2.0` branch. Maybe good to widen the return type in that release?

goetas:
> I saw a 2.0 branch. Maybe good to widen the return type in that release?

I do not have a plan for a 2.0 release yet... it might take years :-/
Having `AdvancedSubscribingHandlerInterface` will allow generators from the next release... and we can deprecate the old interface... just a bit worried for the name...

ruudk:
I'm thinking. Is it really a BC break when you just change the PHPDoc to **also allow Generators** ?

Because the project doesn't use return types, nothing will break.

Anybody that will return an array, or use PHP 7 ` : array` return type, will just work.

The only thing that we need to do, is add an extra test, that uses a Generator with yields. This way, we will be warned in the future when somebody changes the return type of the method.

ruudk:
Created a PR to show it: https://github.com/schmittjoh/serializer/pull/833 Waht do you think?

goetas:
The BC problem here is that someone could have used the interface by itself and was expecting to have always an array. while now can be also a generator... this can lead to bugs in third-party code.

goetas:
closing, there are not special advantages by using generators here. in all the cases all the velues returned by it will have to be iterated when creating the event dispatcher

PR #833, opened 2017-11-02
Allow `getSubscribedEvents` to return a Generator

ruudk:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| Doc updated   |no
| BC breaks?    |no 
| Deprecations? |no 
| Tests pass?   | yes
| Fixed tickets | #832 
| License       | Apache-2.0

### DIFF
--- a/src/JMS/Serializer/EventDispatcher/EventSubscriberInterface.php
+++ b/src/JMS/Serializer/EventDispatcher/EventSubscriberInterface.php
@@ -32,7 +32,7 @@ interface EventSubscriberInterface
      * The class may be omitted if the class wants to subscribe to events of all classes.
      * Same goes for the format key.
      *
-     * @return array
+     * @return array|\Generator
      */
     public static function getSubscribedEvents();
 }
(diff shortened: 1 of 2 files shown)


docteurklein:
when would it make sense to use generators in this method ?

tests/Serializer/EventDispatcher/EventDispatcherTest.php
```diff
@@ -156,6 +156,25 @@ public function testAddSubscriber()
...
+
+        $this->dispatcher->addSubscriber($subscriber);
+        $this->assertAttributeEquals(array(
+            'foo.bar_baz' => array(
+                array(array($subscriber, 'onfoobarbaz'), null, 'foo'),
+            ),
+            'bar' => array(
+                array(array($subscriber, 'bar'), 'foo', null),
+            ),
+        ), 'listeners', $this->dispatcher);
```
docteurklein: we shouldn't assert the internal state of the SUT I guess.

src/JMS/Serializer/EventDispatcher/EventSubscriberInterface.php
```diff
@@ -32,7 +32,7 @@
      * The class may be omitted if the class wants to subscribe to events of all classes.
      * Same goes for the format key.
      *
-     * @return array
+     * @return array|\Generator
```
docteurklein: wouldn't the pseudotype `iterable` work here?

goetas:
in reality with symfony, this gets compiled, so there is no really a reason to use generators here...

goetas:
closing

Issue #836, opened 2017-11-14
DateTime format documentation


enumag:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | no

The [documentation](https://jmsyst.com/libs/serializer/master/reference/annotations#type) of the `@Type` annotation is unclear what the default format for DateTime / DateTimeImmutable actually is.

Also is there a way to use predefined format like `\DateTime::ATOM`? I mean without actually writing the format with `@Type("DateTimeImmutable<'Y-m-d\TH:i:sP'>")`.

goetas:
> The documentation of the @Type annotation is unclear what the default format for DateTime / DateTimeImmutable actually is.

the default format is decided [here](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Handler/DateHandler.php#L64) or [here](https://jmsyst.com/bundles/JMSSerializerBundle/master/configuration#configuration-block-2-0) if you are using symfony

> Also is there a way to use predefined format like \DateTime::ATOM? 

currently no, there is no way to specify constants

PR on the documentation are welcome, the sources are[here](https://github.com/schmittjoh/serializer/tree/master/doc)

enumag:
Hmm this seems a bit odd. The configuration page you linked claims it has the default values - and there is `\DateTime::ATOM` there. Yet the source code you also linked has `\DateTime::ISO8601` as default.

I realize that the constructor's default value is probably not actually used and I also found out in the meantime that the default configuration actually was `\DateTime::ISO8601` in previous versions. So the default value in `DateHandler` is a leftover and maybe it should be changed to ATOM as well? Should I send a PR?

enumag:
For the documentation I thought I would change the "default format" string to a link to `https://jmsyst.com/bundles/JMSSerializerBundle/master/configuration#extension-reference`. However I'm not sure how to write this kind of link since it would go from Serializer docs to JMSSerializerBundle docs.

goetas:
> I realize that the constructor's default value is probably not actually used and I also found out in the meantime that the default configuration actually was \DateTime::ISO8601 in previous versions. So the default value in DateHandler is a leftover and maybe it should be changed to ATOM as well? 

Not fully true, it depends if you are using the standalone library or the symfony bundle. Thats why I sent you the two separate links.

> Should I send a PR?
For documenting it OK, if you want ti change defaults, NO as it will result in a BC break.

[...]

PR #838, opened 2017-11-14
Improve documentation about default DateTime format

enumag:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | https://github.com/schmittjoh/serializer/issues/836
| License       | Apache-2.0

### DIFF
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -351,7 +351,7 @@ Available Types:
 |                                                          | Examples: array<string, string>,                 |
 |                                                          | array<string, MyNamespace\MyObject>, etc.        |
 +----------------------------------------------------------+--------------------------------------------------+
-| DateTime                                                 | PHP's DateTime object (default format/timezone)  |
+| DateTime                                                 | PHP's DateTime object (default format*/timezone)  |
 +----------------------------------------------------------+--------------------------------------------------+
 | DateTime<'format'>                                       | PHP's DateTime object (custom format/default     |
 |                                                          | timezone)                                        |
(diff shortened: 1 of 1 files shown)


doc/reference/annotations.rst
```diff
@@ -388,6 +388,8 @@ Available Types:
 |                                                          | into Doctrine's ArrayCollection class.           |
 +----------------------------------------------------------+--------------------------------------------------+
 
+*) If the standalone jms/serializer is used then default format is `\DateTime::ISO8601` (which is not compatible with ISO-8601 despite the name). For jms/serializer-bundle the default format is `\DateTime::ATOM` (the real ISO-8601 format) but it can be changed in [configuration](https://jmsyst.com/bundles/JMSSerializerBundle/master/configuration#configuration-block-2-0).
```
enumag: I'm not sure how to make the link correctly. This is probably not what you want here.

goetas: > (which is not compatible with ISO-8601 despite the name).

can you elaborate? why is not compatible?

enumag: See PHP documentation: http://php.net/manual/en/class.datetime.php#datetime.constants.atom

```
DateTime::ISO8601
DATE_ISO8601
    ISO-8601 (example: 2005-08-15T15:52:01+0000)

        Note: This format is not compatible with ISO-8601, but is left this way for backward compatibility reasons. Use DateTime::ATOM or DATE_ATOM for compatibility with ISO-8601 instead. 
```

goetas: ok!

enumag: The difference between `DateTime::ISO8601` and `DateTime::ATOM` is the `:` in timezone. According to ISO-8601 it should be there. PHP does not have it there because of a mistake when the constant was added. It probably won't be fixed because of BC.

goetas:
thanks!

Issue #820, opened 2017-09-19
Mapping request payload works for JSON but not for XML


BentCoder:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

I have a model which is used to map JSON and XML payloads. JSON mapping works fine but XML doesn't. I asked similar question [here](https://github.com/schmittjoh/JMSSerializerBundle/issues/470) but it was ignored with a passion!

Property `random` contains random/unknown/dynamic properties so I can not hardcode them. How do you map such properties?

Thanks

**MODEL**
```
/**
 * @Serializer\XmlRoot("data")
 */
class Data
{
    /**
     * @Serializer\Type("array<string>")
     * @Serializer\XmlValue
     */
    public $random;
}
```
**JSON**
```
{
  "random": {
    "hello": "world",
    "colour": "blue",
    "age": "19"
  }
}
```

**RESULT**
```
Data Object
(
      [random] => Array
          (
              [hello] => world
              [colour] => blue
              [age] => 19
          )
)
```

**XML**
```
<?xml version="1.0" encoding="UTF-8"?>
<data>
    <random>
        <hello>world</hello>
        <colour>blue</colour>
        <age>19</age>
    </random>
</data>
```

**RESULT**
```
Data Object
(
      [random] => Array
          (
          )
)
```

This won't map either:
```
@Serializer\Type("array<string, string>")
@Serializer\XmlKeyValuePairs
```

goetas:
Random has to be a type. you should create a class that contains the fields hello, color, age. can't be just a plain array.

An alternative solutioni is custom type handler.

```php
/**
 * @Serializer\XmlRoot("data")
 */
class Data
{
    /**
     * @Serializer\Type("MyType")
     */
    public $random;
}
```

```php

class MyHandler implements SubscribingHandlerInterface
{
    public static function getSubscribingMethods()
    {
        return array(
            array(
                'direction' => GraphNavigator::DIRECTION_DESERIALIZATION,
                'format' => 'xml',
                'type' => 'MyType',
                'method' => 'deserializeMyType'
            ),
            array(
                'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
                'format' => 'xml',
                'type' => 'MyType',
                'method' => 'serializeMyType'
            )
        );
    }

    public function serializeAnyType(XmlSerializationVisitor $visitor, $data, array $type, Context $context)
    {
        // serialize your object here
[...]

PR #840, opened 2017-11-22
Deserialize xmlKeyValuePairs

fdyckhoff:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #820 https://github.com/schmittjoh/JMSSerializerBundle/issues/521 https://github.com/schmittjoh/JMSSerializerBundle/issues/470
| License       | Apache-2.0

### DIFF
--- a/src/JMS/Serializer/XmlDeserializationVisitor.php
+++ b/src/JMS/Serializer/XmlDeserializationVisitor.php
@@ -151,6 +151,27 @@ class XmlDeserializationVisitor extends AbstractVisitor implements NullAwareVisi
 
     public function visitArray($data, array $type, Context $context)
     {
+        if (null === $this->currentMetadata && $context->getMetadataStack()->count()) {
+            $this->setCurrentMetadata($context->getMetadataStack()->top());
+        }
+
+        // handle key-value-pairs
+        if (null !== $this->currentMetadata && $this->currentMetadata->xmlKeyValuePairs) {
+            if (2 !== count($type['params'])) {
+                throw new RuntimeException('The array type must be specified as "array<K,V>" for Key-Value-Pairs.');
+            }
+
+            list($keyType, $entryType) = $type['params'];
+
+            $result = [];
+            foreach ($data as $key => $v) {
+                $k = $this->navigator->accept($key, $keyType, $context);
+                $result[$k] = $this->navigator->accept($v, $entryType, $context);
+            }
+
+            return $result;
+        }
+
         $entryName = null !== $this->currentMetadata && $this->currentMetadata->xmlEntryName ? $this->currentMetadata->xmlEntryName : 'entry';
         $namespace = null !== $this->currentMetadata && $this->currentMetadata->xmlEntryNamespace ? $this->currentMetadata->xmlEntryNamespace : null;
 
--- a/src/JMS/Serializer/XmlDeserializationVisitor.php
+++ b/src/JMS/Serializer/XmlDeserializationVisitor.php
@@ -151,13 +151,6 @@ class XmlDeserializationVisitor extends AbstractVisitor implements NullAwareVisi
 
     public function visitArray($data, array $type, Context $context)
     {
-        $metadata = null;
-        if (null === $this->currentMetadata && $context->getMetadataStack()->count()) {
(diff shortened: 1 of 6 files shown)


src/JMS/Serializer/XmlDeserializationVisitor.php
```diff
@@ -151,6 +151,28 @@ public function visitDouble($data, array $type, Context $context)
...
     {
+        $metadata = $this->currentMetadata;
+        if (null === $metadata && $context->getMetadataStack()->count()) {
+            $metadata = $context->getMetadataStack()->top();
+        }
+
+        // handle key-value-pairs
+        if (null !== $metadata && $metadata->xmlKeyValuePairs) {
+            if (2 !== count($type['params'])) {
+                throw new RuntimeException(sprintf('The array type must be specified as "array<K,V>" for Key-Value-Pairs.'));
```
goetas: `sprintf` here is not neccessary

tests/Serializer/XmlSerializationTest.php
```diff
@@ -249,6 +250,13 @@ public function testArrayKeyValues()
         $this->assertEquals($this->getContent('array_key_values'), $this->serializer->serialize(new ObjectWithXmlKeyValuePairs(), 'xml'));
     }
 
+    public function testDeserializeArrayKeyValues()
+    {
+        $xml = $this->getContent('array_key_values_with_same_value_types');
+        $result = $this->serializer->deserialize($xml, ObjectWithXmlKeyValuePairsWithSameValueTypes::class, 'xml');
+        $this->assertEquals(new ObjectWithXmlKeyValuePairsWithSameValueTypes(), $result);
```
goetas: can you put in the xml different values than the defaults for the property `$list` and assert that those values are set?
the thing is that this test will be true even if the deserializer simply ignores the `$list`  property

src/JMS/Serializer/XmlDeserializationVisitor.php
```diff
@@ -151,6 +151,28 @@ public function visitDouble($data, array $type, Context $context)
 
     public function visitArray($data, array $type, Context $context)
     {
+        $metadata = $this->currentMetadata;
+        if (null === $metadata && $context->getMetadataStack()->count()) {
```
goetas: instead of this "dirty" check, what about adding here a `setCurrentMetadata` call as [here](https://github.com/schmittjoh/serializer/blob/f1d9f1662c4a9b169fd0c81f674187d0032c442c/src/JMS/Serializer/XmlDeserializationVisitor.php#L265). Should be faster and "cleaner"

fdyckhoff:
@goetas Something is still not perfect... In the project, where I integrated this feature, the exception that the array<K,V> type needs to be set is now thrown for other attributes (which do not have the key-value pair annotation set).

I added a testcase for this bug but I'm not sure how to fix it. Should I add the `revertCurrentMetadata` call before each return statement?
The previous version without setting the current metadata variable worked... Should I revert back to that?

fdyckhoff:
I added the revert calls... Works again - but the solution is not pretty... Do you have any improvement ideas?

goetas:
I was more thinking of something as https://github.com/dyvelop/serializer/compare/deserialize-xml-key-value-pairs...goetas:dyvelop-deserialize-xml-key-value-pairs-example?expand=1 

(my example should ne not working but notice the `setCurrentMetadata` inside the `visitProperty` method

fdyckhoff:
Like this? https://github.com/schmittjoh/serializer/pull/840/commits/805286b095ab642b89f607256afcf6426be6c280

goetas:
> Like this? 805286b

Not really. There should be no "visitor-specific" code in the graph navigator.

Did you check my comment (and the code there)?
> I was more thinking of something as https://github.com/dyvelop/serializer/compare/deserialize-xml-key-value-pairs...goetas:dyvelop-deserialize-xml-key-value-pairs-example?expand=1
(my example should ne not working but notice the setCurrentMetadata inside the visitProperty method

BentCoder:
Happy that this will address my issue tickets. Thanks.

PR #844, opened 2017-11-24
support PSR-11 compatible DI containers

xabbuh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | Apache-2.0

To be able to use lazily-loaded event dispatchers even with private services in JMSSerializerBundle we can leverage the `ServiceLocator` feature introduced in Symfony 3.3. However, that class does not depend on the `ContainerInterface` from the DI component, but just needs an arbitrary PSR-11 compatible `ContainerInterface` implementation (which the Symfony containers implements too since Symfony 3.3). The same applies to the `LazyHandlerRegistry`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,6 +40,8 @@
         "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "doctrine/phpcr-odm": "^1.3|^2.0",
         "propel/propel1": "~1.7",
+        "psr/container": "^1.0",
+        "symfony/dependency-injection": "^2.7|^3.3|^4.0",
         "symfony/yaml": "^2.1|^3.0",
         "symfony/translation": "^2.1|^3.0",
         "symfony/validator": "^2.2|^3.0",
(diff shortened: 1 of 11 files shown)


goetas:
Do you mind add `psr/container` in the `require-dev` section and add some basic test? [this](https://github.com/schmittjoh/serializer/tree/master/tests/Serializer/EventDispatcher) should be the right place?

xabbuh:
@goetas Tests added, do I need to update the changelog file too? Oh and I also updated the branch alias as I think that this change qualifies as a new feature.

goetas:
to me looks just great! :man_dancing: 
will check it soon and merge it!
no worries for the changelog, is automatically generated

goetas:
thanks!

PR #856, opened 2018-01-17
Added Object handler + test

carnage:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no
| Tests pass?   | yes
| License       | Apache-2.0

This adds a dynamic Object type for (de)serialization. When serializing, it adds an additional property to the result: '::class' which records the class of the object that is in the property to allow for correct deserialization. 

At the moment this handler isn't registered by default. Should it be?

Docs need updating based on answer to above question.

Future scope for this might be to include a paramter eg Object<Interface> to perform a check that the object is an instance of an interface or extends a base class.

### DIFF
--- a/src/JMS/Serializer/Handler/ObjectHandler.php
+++ b/src/JMS/Serializer/Handler/ObjectHandler.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace JMS\Serializer\Handler;
+
+use JMS\Serializer\Context;
+use JMS\Serializer\Exception\RuntimeException;
+use JMS\Serializer\GraphNavigator;
+use JMS\Serializer\JsonDeserializationVisitor;
+use JMS\Serializer\JsonSerializationVisitor;
+
+class ObjectHandler implements SubscribingHandlerInterface
+{
+    public static function getSubscribingMethods()
+    {
+        $methods = [];
+
+        foreach (array('json', 'xml', 'yml') as $format) {
+            $methods[] = [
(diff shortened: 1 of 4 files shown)


goetas:
Thanks for your PR and for contributing! :tada: ....but not planning to add "generic" objects support in the core. As most of the use cases are too app-specific.

Currently there is https://github.com/schmittjoh/serializer/blob/master/src/Handler/StdClassHandler.php that might be useful but not what you might need...

carnage:
Clearly, as I made this PR, I disagree :) 

I made this PR after copy/pasting or rewriting this handler for a fourth project. So I think it does have a wider use case. My two use cases were:

To remove the need for a discriminator map and aid extensibility. One of my applications allows users to supply their own custom Discount Types as the object which contains the discount type for an order gets serialized a custom one would fail as it wouldn't be in the discriminator map. Using this handler gets around that issue by allowing any object in there; the future scope idea would allow me to garantee that an object provided matches an interface but this wasn't an up front requirement.

For storing arbitrary objects in a rdbms eg newish MySQL json column. This has multiple use cases but I mostly use it for storing domain events. This is an area where using a discriminator map becomes unweildy due to the sheer number of different classes which could be in the property.

After writing all of this out I've realized both use cases could be covered by an instance of handler (eg implement the future scope and make it mandatory) this gives a better level of control while still allowing some flexibility. It also has the advantage of not giving the user a potential security foot gun to use.

Would you be willing to accept an "instance of" handler instead?

Issue #862, opened 2018-01-25
Exception thrown for non-existant accessor to an excluded property


0mars:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Steps required to reproduce the problem

1. Declare a class definition with access type = public_method and an excluded property with no accessor 
2. serialize

## Expected Result

* successful serialization

## Actual Result

* Runtime Exception:
```
JMS\Serializer\Exception\RuntimeException: There is neither a public getIShallNotBeAccessed method, nor a public isIShallNotBeAccessed method, nor a public hasIShallNotBeAccessed method in class JMS\Serializer\Tests\Fixtures\ExcludePublicAccessor. Please specify which public method should be used for retrieving the value of the property iShallNotBeAccessed.
```

goetas:
Fixed in https://github.com/schmittjoh/serializer/pull/863

PR #863, opened 2018-01-25
Allow excluded private properties to not have a getter acc…

0mars:
…essor

Exception is thrown, if the class accessor is set to public_method, and there is an excluded property with no accessor

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #862
| License       | Apache-2.0


goetas:
Do you know if the other mapping strategies have the same issue?

0mars:
@goetas yes it turned out that all the drivers with class accessor definition have the same issue

goetas:
Looks good! Thanks!

Issue #817, opened 2017-09-05
 Node no longer exists - Deserialize Error


mikaelplouhinecvetup:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Steps required to reproduce the problem

1.  Initiate the deserializer
2. Deserialize the data

## Expected Result

No Warning 

## Actual Result

<table border="1">
	<tr><td>str : </td><td>JMS\Serializer\XmlDeserializationVisitor::isNull(): Node no longer exists</td></tr>
	<tr><td>line : </td><td>402</td></tr>
	<tr><td>file : </td><td>/var/www/myApp/vendor/jms/serializer/src/JMS/Serializer/XmlDeserializationVisitor.php</td></tr>
	<tr><td>Warning on : </td><td>localhost</td></tr>
	<tr><td>no : </td><td>2</td></tr>
	<tr><td>callStack : </td><td><br>ln: 402 on /var/www/myApp/vendor/jms/serializer/src/JMS/Serializer/XmlDeserializationVisitor.php, function errorHandler<br>ln: 130 on /var/www/myApp/vendor/jms/serializer/src/JMS/Serializer/GraphNavigator.php, function isNull<br>ln: 266 on /var/www/myApp/vendor/jms/serializer/src/JMS/Serializer/XmlDeserializationVisitor.php, function accept<br>ln: 267 on /var/www/myApp/vendor/jms/serializer/src/JMS/Serializer/GraphNavigator.php, function visitProperty<br>ln: 193 on /var/www/myApp/vendor/jms/serializer/src/JMS/Serializer/Serializer.php, function accept<br>ln: 128 on /var/www/myApp/vendor/jms/serializer/src/JMS/Serializer/Serializer.php, function visit<br>ln: ? on ?, function JMS\Serializer\{closure}<br>ln: 89 on /var/www/myApp/vendor/phpoption/phpoption/src/PhpOption/Some.php, function call_user_func<br>ln: 131 on /var/www/myApp/vendor/jms/serializer/src/JMS/Serializer/Serializer.php, function map<br>ln: 355 on /var/www/myApp/modules/abaxis/abaxisControllerAjax.php, function deserialize<br></td></tr>
[...]

PR #871, opened 2018-02-07
Workaround for Node no longer exists

Deltachaos:
The `isset($attributes[$name])` sometimes also throws a notice like described in https://bugs.php.net/bug.php?id=75168

This introduces the workaround form #817 also in this place. We cannot use the `isNull` method here because of the todo mark in this method.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #817
| License       | Apache-2.0

### DIFF
--- a/src/JMS/Serializer/XmlDeserializationVisitor.php
+++ b/src/JMS/Serializer/XmlDeserializationVisitor.php
@@ -268,11 +268,12 @@ class XmlDeserializationVisitor extends AbstractVisitor implements NullAwareVisi
         }
 
         if ($metadata->xmlAttribute) {
-
-            $attributes = $data->attributes($metadata->xmlNamespace);
-            if (isset($attributes[$name])) {
-                $v = $this->navigator->accept($attributes[$name], $metadata->type, $context);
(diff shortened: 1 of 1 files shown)


goetas:
any way to test this?

goetas:
Can you help here to reproduce the issue?

Deltachaos:
I do not have the original case that caused this problem anymore

Kalyse:
I can't provide a test case, but after applying the patch manually to my code, this resolves my issue also. 

For some unknown reason to me, the issue was the visitProperty was receiving an empty SimpleXMLElement without a name. I have no idea how that is possible, but for obviousness, 

`if (isset($attributes[$name])) {` fails, because of no name. That's why the warning is thrown. I can't provide a test case, no more than the author in 2018 could, but can confirm this does fix the underlying issue for me.

goetas:
I understand that for some cases might be a problem, i'm wondering if this is a bug or an issue with the configurations in your project, and IMO the only way to figure out it is to find a way to reproduce the issue

PR #872, opened 2018-02-08
make xml serializer a bit more configurable

brosenberger:
| Q             | A
| ------------- | ---
| Bug fix?      |/no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | /no 
| Tests pass?   | yes/no
| Fixed tickets | #... 
| License       | Apache-2.0

following scenario was not possible to be achieved with generated yml-metadata (which was not touchable):
* generate serialized xml data where the default namespace elements are prefixed, e.g. with n1
* additionally all attributes needed to have there representative namespace prefix

proposed solution:
* modification of the setDefaultRootName method of XmlSerializationVisitor to add a 3rd parameter prefix, which is added if any rootnamespace is given on dom document creation
* modification of the visibility of the method setAttributeOnNode in order to override the default behavior of the namespace prefixing, e.g.:

`
protected function setAttributeOnNode(\DOMElement $node, $name, $value, $namespace = null)
    {
        if ($namespace===null && $node->namespaceURI!==null) {
            $namespace = $node->namespaceURI;
        }
        parent::setAttributeOnNode($node, $name, $value, $namespace);
    }
`

### DIFF
--- a/src/JMS/Serializer/XmlSerializationVisitor.php
+++ b/src/JMS/Serializer/XmlSerializationVisitor.php
@@ -37,6 +37,7 @@ class XmlSerializationVisitor extends AbstractVisitor
     private $navigator;
     private $defaultRootName = 'result';
     private $defaultRootNamespace;
+    private $defaultRootPrefix = '';
     private $defaultVersion = '1.0';
     private $defaultEncoding = 'UTF-8';
     private $stack;
@@ -57,10 +58,11 @@ class XmlSerializationVisitor extends AbstractVisitor
         $this->formatOutput = true;
     }
 
-    public function setDefaultRootName($name, $namespace = null)
+    public function setDefaultRootName($name, $namespace = null, $prefix = '')
     {
         $this->defaultRootName = $name;
         $this->defaultRootNamespace = $namespace;
+        $this->defaultRootPrefix = !empty($prefix)?$prefix.':':'';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/JMS/Serializer/XmlSerializationVisitor.php
```diff
@@ -57,10 +58,11 @@ public function __construct($namingStrategy, AccessorStrategyInterface $accessor
         $this->formatOutput = true;
     }
 
-    public function setDefaultRootName($name, $namespace = null)
+    public function setDefaultRootName($name, $namespace = null, $prefix = '')
```
goetas: can you make `$prefix=null` instead of `''`?

goetas: can you make `$prefix=null` instead of `$prefix = ''`?

goetas: can you add a test for this?

src/JMS/Serializer/XmlSerializationVisitor.php
```diff
@@ -498,7 +500,7 @@ private function createElement($tagName, $namespace = null)
         return $this->document->createElementNS($namespace, $prefix . ':' . $tagName);
     }
 
-    private function setAttributeOnNode(\DOMElement $node, $name, $value, $namespace = null)
+    protected function setAttributeOnNode(\DOMElement $node, $name, $value, $namespace = null)
```
goetas: :+1:

goetas:
closed by c30cd008cd875ac6463e4f469f09d11ae560d094

Issue #334, opened 2014-09-10
SerializedName ignored when using IdenticalPropertyNamingStrategy


eschwartz:
It appears as though the `@SerializedName` annotation is being ignored, when used with the `IdenticalPropertyNamingStrategy`.

My user class

``` php
// MyApp\Model\User

namespace MyApp\Model;

use JMS\Serializer\Annotation as JMS;

/**
 * User
 */
class User {
  /**
   * @var integer
   *
   * @JMS\SerializedName("foo")
   */
  private $someProp;

  // ...
}
```

Instantiating the serializer (I'm using ZF2, so this is a service factory)

``` php
return array(
  // ...
  'service_manager' => array(
    'factories' => array(
      'jms_serializer' => function() {
        \Doctrine\Common\Annotations\AnnotationRegistry::registerAutoloadNamespace(
          'JMS\Serializer\Annotation',
          __DIR__ . '/../../../vendor/jms/serializer/src'
        );

        $namingStrategy = new \JMS\Serializer\Naming\IdenticalPropertyNamingStrategy();
        return \JMS\Serializer\SerializerBuilder::create()
          ->setCacheDir(__DIR__ . '/../../../data/cache')
          ->setPropertyNamingStrategy($namingStrategy)
          ->build();
      }
    )
  )
);
```

With this setup, the `User` object is being serialized a `{ "someProp": "someValue" }`. I would expect it to be serialized as `{  "foo": "someValue" }`

I believe the issue is within the `JMS\Serializer\Naming\IdenticalPropertyNamingStrategy`:

``` php
class IdenticalPropertyNamingStrategy implements PropertyNamingStrategyInterface
{
    public function translateName(PropertyMetadata $property)
    {
        return $property->name;
    }
}
```

If I override the `translateName` method to check for a `serializedName` property, my issue is fixed.

``` php
       public function translateName(PropertyMetadata $property) {
        if ($property->serializedName) {
            return $property->serializedName;
        }
        return $property->name;
    }
```

I can submit a pull request to fix this -- I just want to check that I'm going about this the right way, first.

SchizoDuckie:
This should be fixed indeed. It's a nasty one to stumble upon and figure out.

Steps to fix for those coming after me:

- added @m14t's IdenticalUnlessSpecified strategy to my project
- change config.yml parameters: 
[...]

PR #874, opened 2018-02-15
Added note on SerializedName annotation valididity

bobvandevijver:
As discussed in https://github.com/schmittjoh/serializer/issues/334, when you're using anything else than the default configuration of the serializer with respect to the naming strategy, you will need to explicitly enable the SerializedNameAnnotationStrategy which wraps your custom naming strategy.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #334 
| License       | Apache-2.0


goetas:
Thanks!

PR #879, opened 2018-02-27
Discriminator property serialization when parent is in discriminator map

supersmile2009:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | maybe (see description below)
| Deprecations? | no
| Tests pass?   | yes (fixed)
| Fixed tickets | n/a
| License       | Apache-2.0

Discriminator property was added only when merging class metadata. It covered the cases when discriminated parent class didn't belong to discriminator map, so it kind of assumed that parent isn't going to belong to discriminator map and serialized.

So if parent class is a normal instantiable class and when it belongs to discriminator map, logic adding discriminator property was never executed and it was serialized without discriminator field.
I moved this logic to another method, now it's called when setting discriminator initially and when merging metadata.

Example of an issue:
```
/**
 * @ORM\Entity()
 * @ORM\Table(name="foo_bar_table")
 * @ORM\InheritanceType("SINGLE_TABLE")
 * @ORM\DiscriminatorColumn(name="type", type="string", length=30, fieldName="type")
 * @ORM\DiscriminatorMap({
 *     "foo" = "Foo",
 *     "bar" = "Bar",
 * })
 */
class Foo
{

}
```

```
/**
 * @ORM\Entity()
 */
class Bar extends Foo
{
}
```
### Why this is potentially a BC?
I had to fix a few unit tests by making parent classes abstract. It would be logical for parent class to be abstract since it's not in discriminator map. *This is potentially a BC.*
If parent is not in the map and is not abstract or interface it will throw an exception 'The sub-class "%s" is not listed in the discriminator of the base class "%s".'
It didn't happen before, because this check was never executed for parent class. It shouldn't affect serialization itself, but it will start throwing an exception now if parent isn't abstract and isn't in the map, which is weird from design point of view anyway.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -32,6 +32,8 @@ in case you have heavily used internal-api here are the most important the chang
 - Removed `Serializer::setSerializationContextFactory` and `Serializer::setDeserializationContextFactory`
 - Removed `Serializer::getMetadataFactory` 
 - As default now JSON preserve trailing zeros when serializing a float
+- When using a discriminator map, parent class should either be declared abstract, or included into the discriminator
+  map
 
 **Deprecations** (will be removed in 3.0)
 
--- a/src/Metadata/ClassMetadata.php
+++ b/src/Metadata/ClassMetadata.php
@@ -137,6 +137,8 @@ class ClassMetadata extends MergeableClassMetadata
         $this->discriminatorFieldName = $fieldName;
         $this->discriminatorMap = $map;
         $this->discriminatorGroups = $groups;
+
+        $this->setDiscriminatorProperty();
     }
 
     private function getReflection(): \ReflectionClass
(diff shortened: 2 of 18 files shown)


goetas:
@supersmile2009 if you are interested (and rebase if), this can land in 2.x but not in 1.x

supersmile2009:
@goetas 
Rebased and added some metadata/serialization/deserialization tests for this particular case.

src/Metadata/ClassMetadata.php
```diff
@@ -384,6 +353,48 @@ public function unserialize($str)
         parent::unserialize($parentStr);
     }
 
+    private function setDiscriminatorProperty()
```
goetas: can you please rename this method using a different name? this is not a setter. something as "handleDiscriminatorMap" ?

UPGRADING.md
```diff
@@ -32,6 +32,8 @@ in case you have heavily used internal-api here are the most important the chang
 - Removed `Serializer::setSerializationContextFactory` and `Serializer::setDeserializationContextFactory`
 - Removed `Serializer::getMetadataFactory` 
 - As default now JSON preserve trailing zeros when serializing a float
+- When using a discriminator map, parent class should either be declared abstract, or included into the discriminator
```
goetas: Looks a pretty common case extending a class that is not abstract... why enforcing that kind of constraint?

classical usecase: Abstract vehicle > concrete car > concrete electrical car

src/Metadata/ClassMetadata.php
supersmile2009: done

UPGRADING.md
supersmile2009: But why would someone want to leave ConcreteCar not included into discriminator map then? If you can instantiate it, it means that you will potentially serialize/deserialize it, so you need to include it into the map.

goetas: makes sese :+1:

goetas:
Thank you!

PR #884, opened 2018-03-08
Add AccessorFinder.

zored:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | Apache-2.0


zored:
TODO: add global accessor finder.

src/JMS/Serializer/Annotation/AccessorFinder.php
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Defines accessor finder.
+ *
+ * @Annotation
+ * @Target("CLASS")
+ */
+
+final class AccessorFinder
+{
+    public $class = DefaultAccessorFinder::class;
```
goetas: What about having also a "service" ? (and some kind of accessor registry where is possible to add accessors via dependency injection?) to avoid instantiating the class in the metadata reader.... in this way 
accessor finders can have parameters and their dependencies too.

src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
```diff
@@ -81,9 +84,16 @@ public function loadMetadataForClass(\ReflectionClass $class)
         $excludeAll = false;
         $classAccessType = PropertyMetadata::ACCESS_TYPE_PROPERTY;
         $readOnlyClass = false;
+        $accessorFinder = null;
         foreach ($this->reader->getClassAnnotations($class) as $annot) {
             if ($annot instanceof ExclusionPolicy) {
                 $exclusionPolicy = $annot->policy;
+            } elseif ($annot instanceof AccessorFinder) {
+                $accessorFinder = $annot->class;
+                $accessorFinder = new $accessorFinder();
```
goetas: see my comment on `src/JMS/Serializer/Annotation/AccessorFinder.php`

tests/Fixtures/AccessorFinder.php
```diff
@@ -0,0 +1,23 @@
...
+namespace JMS\Serializer\Tests\Fixtures;
+use JMS\Serializer\Annotation as Serializer;
+
+/** @Serializer\AccessorFinder() */
+final class AccessorFinder
+{
+    /**
+     * @Serializer\Type("integer")
+     */
+    private $a1_b = 1;
```
goetas: can you also add one property that does not have getters/setters to see how it behaves if is missing

goetas: can you also add one property that is marked as read-only to see how it behaves

tests/Serializer/XmlSerializationTest.php
```diff
@@ -87,6 +88,23 @@ public function getXMLBooleans()
         return array(array('true', true), array('false', false), array('1', true), array('0', false));
     }
 
+    public function testAccessorFinderDeserialization()
+    {
+        $value = 2018;
+        $finder = $this->deserialize(<<<XML
```
goetas: can you move this xml to a file?

zored:
@goetas thanks for review! Good notes, I'll fix them. 👍 
This PR is intended to add general naming rules for accessors. So we can add it once instead of writing for each property.

zored:
@goetas I found `@AccessType` configuration and therefore created another PR: #888
Took into account all your notes there. 👍

Issue #852, opened 2018-01-11
setSerialiseNull(true) + exclusion strategies still include data


carnage:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | maybe (It worked as expected on a very old version, possibly prior to 1.0.0)
| RFC?             | no

## Steps required to reproduce the problem

1. Use an exclusion strategy to exclude an item from an array/hash
2. Set serialise null => true
3. Get extra null values in the array (eg item wasn't fully excluded)

## Expected Result

*  Only the items the exclusion strategy allows through are in the serialised result

## Actual Result

*  Extra nulls are included in the array/hash
* (untested) I suspect the same will happen for excluded properties. eg their value will be set null instead of the key being excluded entirely.

## Details
The issue occurs here:
https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/GenericSerializationVisitor.php#L111

When the navigator determines an item should be excluded, it returns null for that value. This works as expected if setSerialiseNull is false, however when it's true, the item gets included with a null value.

## Solution

The solution here is to have the navigator throw an UnacceptableException (https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/GraphNavigator.php#L225) when it encounters a field which should not be included in the result. That can then be caught in the code linked above and handled eg excluded from the result. 

## BC preservation
As this will now change behaviour, it may be desirable to include a configuration parameter (eg setIncludeExcludedAsNull()) to enable this new behaviour, defaulting to on. However I consider this behaviour to be a bug and is inconsistent with the behaviour when setSerialiseNull is set to false so I don't think this is required

fsevestre:
I'm having the exact same issue after upgrading from `1.1.0` to the latest version. Do you know any workaround ?

fsevestre:
The issue was introduces by #660 :|

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/895

PR #886, opened 2018-03-09
852 - setSerialiseNull(true) + exclusion strategies still include data

fsevestre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | wip
| Fixed tickets | #852
| License       | Apache-2.0

In this PR, I supposed that the only way to transform an object to null was using an exclusion strategy (I hope that this is the only way).

----

First commit: the failing test.
Second commit: part of the fix (only `GenericSerializationVisitor` and `JsonSerializationVisitor` atm).

However, I have an issue with `XmlSerializationVisitor` and `YamlSerializationVisitor` because they call `$this->navigator->accept` after the check for `null` (instead of before like the other 2 serialization visitor).
Is this the wanted behavior ?


fsevestre:
Juste updated the first (better example and same set of tests for `setSerializeNull(true)` and `setSerializeNull(false)` and second commits.

The tests pass now, but I'm not sure that the way I check if the exclusion rule has skipped the class is good.

Any help with this will be well appreciated.

goetas:
Thanks a lot for the PR. This was a long-standing feature request and I'm interested in merging it.

Im just afraid this could be considered a BC break. Is it? Should we have a "flag" to disable the serialization of null values inside a collection?

fsevestre:
Personally, I consider this as a bug fix since the BC break was added in 1.4 and it broke my code when I tried to upgrade an old app.
But the bug was introduced in October 2016 so it may indeed break newer applications (I don't know in which case but it can be possible I think).

If you think it's safer to add a way to enable this change I will add it.

src/JMS/Serializer/GenericSerializationVisitor.php
```diff
@@ -109,16 +109,18 @@ public function visitArray($data, array $type, Context $context)
         $isList = isset($type['params'][0]) && !isset($type['params'][1]);
 
         foreach ($data as $k => $v) {
-            $v = $this->navigator->accept($v, $this->getElementType($type), $context);
+            $val = $this->navigator->accept($v, $this->getElementType($type), $context);
 
-            if (null === $v && $context->shouldSerializeNull() !== true) {
+            if (null === $val && ($context->shouldSerializeNull() !== true
+                || (is_object($v) && null !== $context->getExclusionStrategy() && $context->getMetadataStack()->isEmpty()))
```
goetas: can you explain me why `(is_object($v) && null !== $context->getExclusionStrategy() && $context->getMetadataStack()->isEmpty())` is necessary? isnt `(null === $v && $context->shouldSerializeNull() !== true)` enough?

fsevestre: In fact, it's the `$context->shouldSerializeNull() !== true` condition the problem because I want `null` values to be serialized (`->setSerializeNull(true)`) but not objects serialized to `null` using an exclusion strategy.

fsevestre: But as I said in my comment on the PR, I'm not sure that `is_object($v) && null !== $context->getExclusionStrategy() && $context->getMetadataStack()->isEmpty()` is the right check to do. I came to this condition because I went in the `GraphNavigator` class and saw this https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/GraphNavigator.php#L222-L228
But in case of some refactoring, the condition might be wrong :/

goetas: To me makes much more sense to not serialize nulls, independently from the reason why they are null (simple nulls or exclusion strategies). 

What do you think

fsevestre: In my case it will be complicated because it will break my clients' fronts because they don't check for the key's existence but ony the value :/

--

Maybe it's more a "returned value" issue, which shouldn't be considered as the same `null` value ?

I may be wrong (I don't work a lot with serializers) but in my mind `array(new Object(toSkip), new Object(doNotSkip))` (with an exclusion strategy removing the first object) and `array(new Object(doNotSkip))` should return the same result. FYI it's not totally the case anyway because of the array key: in the first example, the second object will have the key `1` and in the second the key `0` in the json/xml/... result.

--

Just to add my use case, we generate a big tree (no limit of level, depends on the configuration) of objects, do lots of things on it, and update a `display (bool)` property which is used in an exclusion strategy to skip the class (`shouldSkipClass`) with `display` at false and remove the property (`shouldSkipProperty`).
All the objects are very important during the process (even the ones that are supposed to be hidden) but we don't want to return some of them (`display` false) through the API because it's a business requirement.

PR #888, opened 2018-03-11
AccessType naming

zored:
# New features:
- Added access type naming configuration: with new `naming="camel_case"` for property `$foo_bar` we need `getFooBar()` accessor (instead of `getFoo_Bar()`).
- Now we can require accessors for each class: by modifying builder (default behavour preserved).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | Apache-2.0

### DIFF
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -98,6 +98,36 @@ set the value via reflection, but you may change this to use a public method ins
         }
     }
 
+    class Shop
+    {
+        /**
+         * @AccessType("public_method", naming="camel_case")
+         * @ReadOnly
+         */
+        private $full_address;
+
+        public function getFullAddress()
+        {
+            return $this->full_address;
+        }
+    }
+
+
+To use accessors instead of properties by default, modify builder:
+
+.. code-block :: php
+
+    <?php
+    use JMS\Serializer\Accessor\Updater\ClassAccessorUpdater;
+    use JMS\Serializer\SerializerBuilder;
+
+    $serializerBuilder = SerializerBuilder::create()->build();
+    $serializerBuilder->setPropertyUpdater(new ClassAccessorUpdater(
+        PropertyMetadata::ACCESS_TYPE_PUBLIC_METHOD,
+        PropertyMetadata::ACCESS_TYPE_NAMING_CAMEL_CASE
+    ));
+    $serializer = $serializerBuilder->build();
+
 @Accessor
 ~~~~~~~~~
 This annotation can be defined on a property to specify which public method should
@@ -126,7 +156,7 @@ be called to retrieve, or set the value of the given property:
             $this->name = $name;
         }
     }
-    
+
 .. note ::
 
     If you need only to serialize your data, you can avoid providing a setter by
--- a/src/JMS/Serializer/Accessor/Updater/ClassAccessorUpdater.php
+++ b/src/JMS/Serializer/Accessor/Updater/ClassAccessorUpdater.php
@@ -1,0 +1,173 @@
+<?php
+
+namespace JMS\Serializer\Accessor\Updater;
+
+use JMS\Serializer\Exception\RuntimeException;
+use JMS\Serializer\Metadata\ClassMetadata;
+use JMS\Serializer\Metadata\ClassMetadataUpdaterInterface;
+use JMS\Serializer\Metadata\PropertyMetadata;
+
+/**
+ * Default behaviour is required for propper accessors work.
+ *
+ * For `$person_name` field expected accessors are:
+ * - `getPerson_name()`
+ * - `setPerson_name()`
+ */
+class ClassAccessorUpdater implements ClassMetadataUpdaterInterface
+{
+    /**
+     * @var string
+     */
+    private $defaultType;
+
+    /**
+     * @var string
+     */
+    private $defaultNaming;
+
+    /**
+     * @var array
+     */
+    private $getterPrefixes;
+
+    /**
+     * @var array
+     */
+    private $setterPrefixes;
+
+    public function __construct(
+        $defaultType = PropertyMetadata::ACCESS_TYPE_PROPERTY,
+        $defaultNaming = PropertyMetadata::ACCESS_TYPE_NAMING_EXACT,
+        array $getterPrefixes = ['get', 'is', 'has'],
+        array $setterPrefixes = ['set', 'update']
+    )
+    {
+        $this->defaultType = $defaultType;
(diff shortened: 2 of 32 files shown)


src/JMS/Serializer/Metadata/PropertyMetadata.php
```diff
@@ -197,10 +225,17 @@ public function unserialize($str)
             $this->skipWhenEmpty = $unserialized['skipWhenEmpty'];
         }
 
-        parent::unserialize($parentStr);
+        list($this->class, $this->name) = unserialize($parentStr);
+        $this->prepareLazyReflection();
```
scaytrase: looks like `__wakeup` method

src/JMS/Serializer/Metadata/PropertyMetadata.php
```diff
@@ -27,6 +26,9 @@ class PropertyMetadata extends BasePropertyMetadata
     const ACCESS_TYPE_PROPERTY = 'property';
     const ACCESS_TYPE_PUBLIC_METHOD = 'public_method';
 
+    const ACCESS_TYPE_NAMING_EXACT = 'exact';
```
scaytrase: I think this should be something like `ACCESSOR_NAMING_STRATEGY_EXACT`

src/JMS/Serializer/Accessor/Updater/ClassAccessorUpdater.php
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        $class = $this->getDeclaringClass($metadata);
+        $name = $metadata->name;
+
+        if (null !== $method = $this->findAccessorName($this->setterPrefixes, $name, $class, $naming)) {
+            return $method;
+        }
+
+        throw new RuntimeException("Specify public setter for {$class->getName()}::\${$name} (using $naming naming)");
```
scaytrase: This should be metadata exceptions, not raw runtime

src/JMS/Serializer/Accessor/Updater/ClassAccessorUpdater.php
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * @param string $accessorPrefix
+     * @param string $propertyName
+     * @param string $naming
+     *
+     * @return string|null
+     */
+    protected function getAccessorName($accessorPrefix, $propertyName, $naming)
```
scaytrase: This method is totally decoupled from the class logic, so it could be refactored out making easy extending list of supported naming strategies

src/JMS/Serializer/Accessor/Updater/ClassAccessorUpdater.php
```diff
@@ -0,0 +1,166 @@
...
+            $propertyMetadata->accessType ?:
+                $classMetadata->accessType ?:
+                    $this->defaultType;
+
+        if ($type !== PropertyMetadata::ACCESS_TYPE_PUBLIC_METHOD) {
+            return;
+        }
+
+        $naming =
+            $propertyMetadata->accessTypeNaming ?:
```
scaytrase: I'm not sure we should event take `propertyMetadata->accessTypeNaming` into account, since it's easier to configure accessors there

src/JMS/Serializer/Annotation/AccessType.php
```diff
@@ -27,8 +27,12 @@
...
     /**
-     * @Required
      * @var string
      */
     public $type;
+
+    /**
+     * @var string
+     */
+    public $naming;
```
scaytrase: maybe `namingStrategy` ?

src/JMS/Serializer/Metadata/ClassMetadata.php
```diff
@@ -63,6 +63,9 @@ class ClassMetadata extends MergeableClassMetadata
     public $xmlDiscriminatorCData = true;
     public $xmlDiscriminatorNamespace;
 
+    public $accessType;
+    public $accessTypeNaming;
```
scaytrase: I suggest `accessorNamingStrategy`

src/JMS/Serializer/Metadata/Driver/AnnotationDriver.php
```diff
@@ -240,12 +251,19 @@ public function loadMetadataForClass(\ReflectionClass $class)
...
+                        $accessor[0],
+                        $accessor[1],
+                        $accessTypeNaming
+                    );
                     $classMetadata->addPropertyMetadata($propertyMetadata);
                 }
             }
         }
 
+        $this->classUpdater->update($classMetadata);
```
scaytrase: I'm not really sure that `classUpdater`, `ClassMetadataUpdater` and `update` method really reflects current responsibilities

src/JMS/Serializer/Metadata/Driver/XmlDriver.php
```diff
@@ -351,6 +367,8 @@ protected function loadMetadataFromFile(\ReflectionClass $class, $path)
             }
         }
 
+        $this->classMetadataUpdater->update($metadata);
```
scaytrase: All calls are made at the end of `loadMetadataFromFile`. Shouldn't this be called somewhere in `AbstractFileDriver`? This would simplify the logic flow

doc/reference/annotations.rst
```diff
@@ -98,6 +98,36 @@ set the value via reflection, but you may change this to use a public method ins
         }
     }
 
+    class Shop
+    {
+        /**
+         * @AccessType("public_method", naming="camel_case")
```
goetas: In the example, I think  would be more useful use this anotation at class-level.

doc/reference/annotations.rst
```diff
@@ -98,6 +98,36 @@ set the value via reflection, but you may change this to use a public method ins
...
+To use accessors instead of properties by default, modify builder:
+
+.. code-block :: php
+
+    <?php
+    use JMS\Serializer\Accessor\Updater\ClassAccessorUpdater;
+    use JMS\Serializer\SerializerBuilder;
+
+    $serializerBuilder = SerializerBuilder::create()->build();
+    $serializerBuilder->setPropertyUpdater(new ClassAccessorUpdater(
```
goetas: for the same reasons https://github.com/schmittjoh/JMSSerializerBundle/pull/427 was not accepted, even here will prefer not allowing to change the access type at runtime but asking explicit configurations via metadata

PR #916, opened 2018-04-26
Enable strict mode, fix type juggling and drop some runtime coercion broken by design

Majkl578:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes

This breaks some type juggling which is incompatible with strict mode by design (like deserializing object with `__toString()`).

### DIFF
--- a/src/SerializationGraphNavigator.php
+++ b/src/SerializationGraphNavigator.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * Copyright 2016 Johannes M. Schmitt <schmittjoh@gmail.com>
  *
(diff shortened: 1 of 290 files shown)


src/SerializationGraphNavigator.php
```diff
@@ -108,22 +110,22 @@ public function accept($data, array $type = null, Context $context)
                 return $visitor->visitNull($data, $type);
 
             case 'string':
-                return $visitor->visitString($data, $type);
+                return $visitor->visitString((string)$data, $type);
```
Majkl578: Are these changes are potentially BC breaking, but not by this PR, by already existing type hints. This only eliminates runtime type juggling.

tests/Serializer/ContextTest.php
```diff
@@ -191,11 +193,5 @@ public function testSerializeNullOption()
 
         $context->setSerializeNull(true);
         $this->assertTrue($context->shouldSerializeNull());
-
-        $context->setSerializeNull("foo");
-        $this->assertTrue($context->shouldSerializeNull());
-
-        $context->setSerializeNull("0");
-        $this->assertFalse($context->shouldSerializeNull());
```
Majkl578: This is BC break-ish, incompatible with strict typing by design.

tests/Serializer/JsonSerializationTest.php
```diff
@@ -203,34 +205,18 @@ public function getPrimitiveTypes()
                 'type' => 'boolean',
                 'data' => true,
             ),
-            array(
```
Majkl578: All these are removed, type coertion of this sort is incompatible with strict types by design.

tests/Serializer/XmlSerializationTest.php
```diff
@@ -385,15 +387,6 @@ public function testDeserializingNull()
         $this->markTestSkipped('Not supported in XML.');
     }
 
-    public function testDeserializeWithObjectWithToStringMethod()
```
Majkl578: Objects with `__toString()` are not `string`, therefore removed as incompatible with strict types.

goetas: ok, this "feature" can be removed

tests/Serializer/ContextTest.php
goetas: agree with getting rid of it

tests/Serializer/XmlSerializationTest.php
Majkl578: It has simple user-land workaround: `(string) $stringishObject`, but imho that doesn't make much sense anyway. :grinning:

goetas:
Checked with https://github.com/schmittjoh/serializer/pull/917, works!

Issue #910, opened 2018-04-26
Risky tests in master


Majkl578:
There are currently 9 risky tests that don't assert anything:
```
There were 9 risky tests:

1) JMS\Serializer\Tests\Handler\ArrayCollectionHandlerTest::testSerializeArray
This test did not perform any assertions

2) JMS\Serializer\Tests\Handler\ArrayCollectionHandlerTest::testSerializeArraySkipByExclusionStrategy
This test did not perform any assertions

3) JMS\Serializer\Tests\Handler\DateHandlerTest::testSerializeDate with data set #0 (array('Y-m-d'))
This test did not perform any assertions

4) JMS\Serializer\Tests\Handler\DateHandlerTest::testSerializeDate with data set #1 (array('Y-m-d', '', 'Y-m-d|'))
This test did not perform any assertions

5) JMS\Serializer\Tests\Handler\DateHandlerTest::testSerializeDate with data set #2 (array('Y-m-d', '', 'Y'))
This test did not perform any assertions

6) JMS\Serializer\Tests\Serializer\GraphNavigatorTest::testNavigatorChangeTypeOnSerialization
This test did not perform any assertions

7) JMS\Serializer\Tests\Serializer\JsonSerializationTest::testFormErrorsWithNonFormComponents
This test did not perform any assertions

8) JMS\Serializer\Tests\Serializer\XmlSerializationTest::testWhitelistedDocumentTypesAreAllowed
This test did not perform any assertions

9) JMS\Serializer\Tests\Serializer\XmlSerializationTest::testFormErrorsWithNonFormComponents
This test did not perform any assertions
```

goetas:
True... test should be improved

goetas:
most of the risky tests are because they are testing that exceptions are not thrown... ideas here?

Majkl578:
https://github.com/sebastianbergmann/phpunit/issues/2484

You can use `@doesNotPerformAssertions`, but it would be better to rework the test to actually test some behavior.

goetas:
Do not know what is better... `@doesNotPerformAssertions` or https://github.com/schmittjoh/serializer/pull/918 :thinking:

Majkl578:
I'd say `@doesNotPerformAssertions`. Asserting that true is true is nonsense. :D

goetas:
fixed in #918

simPod:
It doesn't count towards coverage tho

PR #919, opened 2018-04-26
Make PHPUnit stricter, drop redundant defaults

Majkl578:
* add PHPUnit schema
* most of the options are now only duplicating defaults
* be strict about test behavior
* fail on risky tests and warnings

Should prevent #910 to re-appear.


goetas:
:+1:

PR #929, opened 2018-04-29
Added graph navigator factories

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | Apache-2.0

This reduces from 11  to 7 the arguments necessary for the serializer class.

### DIFF
--- a/src/Serializer.php
+++ b/src/Serializer.php
@@ -124,6 +124,15 @@ final class Serializer implements SerializerInterface, ArrayTransformerInterface
 
     private function getNavigator(int $direction) : GraphNavigatorInterface
     {
+        if (!isset($this->graphNavigators[$direction])) {
+            throw new RuntimeException(
+                sprintf(
+                    'Can not find a graph navigator for the direction "%s".',
+                    $direction === GraphNavigatorInterface::DIRECTION_SERIALIZATION ? 'serialization' : 'deserialization'
+                )
+            );
+        }
+
         return $this->graphNavigators[$direction]->getGraphNavigator();
     }
 
--- a/src/Serializer.php
+++ b/src/Serializer.php
@@ -78,8 +78,7 @@ final class Serializer implements SerializerInterface, ArrayTransformerInterface
         SerializationContextFactoryInterface $serializationContextFactory = null,
         DeserializationContextFactoryInterface $deserializationContextFactory = null,
         AbstractParser $typeParser = null
-    )
-    {
+    ) {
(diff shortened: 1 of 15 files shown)


src/Serializer.php
```diff
@@ -45,13 +40,24 @@
...
-    private $objectConstructor;
-    private $dispatcher;
+
+    /**
+     * @var TypeParser
+     */
     private $typeParser;
 
+    /**
+     * @var array|SerializationVisitorFactory[]
```
Majkl578: "array" is useless and implied by "[]"

src/SerializerBuilder.php
```diff
@@ -96,18 +102,25 @@ class SerializerBuilder
...
 
-    public static function create()
+    public static function create(...$args)
     {
-        return new static();
+        return new static(...$args);
     }
 
-    public function __construct()
+    public function __construct(HandlerRegistryInterface $handlerRegistry = null, EventDispatcherInterface $eventDispatcher = null)
```
Majkl578: Missing nullable `?`.

src/SerializerBuilder.php
```diff
@@ -96,18 +102,25 @@ class SerializerBuilder
...
     {
         $this->typeParser = new TypeParser();
-        $this->handlerRegistry = new HandlerRegistry();
-        $this->eventDispatcher = new EventDispatcher();
+        $this->handlerRegistry = $handlerRegistry ?: new HandlerRegistry();
```
Majkl578: should be ??

goetas: should be no difference, since can be only `object|null`, right?

src/Serializer.php
goetas: fixed

src/SerializerBuilder.php
goetas: isnt optional when is an object?

goetas: I mean, with the current definition, 
```php
new SerializerBuilder();
new SerializerBuilder(null, null);
new SerializerBuilder(null, new EventDispatcher());
```

all of this should be valid, right?

Majkl578: Yes, but the signature is still missing the `?`. 😆

Issue #904, opened 2018-04-11
[2.0 proposal] Consider generated hydrator to construct objects/graph


Majkl578:
| Q                | A
| ---------------- | -----
| Feature request? | yes
| BC Break? | yes
| RFC?             | yes

https://github.com/Ocramius/GeneratedHydrator

goetas:
Not for 2.0.. too much different from the current processing model

Majkl578:
Why did you close it?

goetas:
Ocramius/GeneratedHydrator has a completely different approach to serialization/deserialization compared to this library.
In my opinion is a great solution but implementing it into the jms serializer will make it a completely different serializer. most probably the resulting library will have a completely different set of features. It will be just a different serializer.

goetas:
After some discussion with @Majkl578 , this approach could work.
Setting all the props in an array and then using the hydrator could work (and be very fast!)

PR #933, opened 2018-05-01
Generated hydrators

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #904
| License       | Apache-2.0

tried to implement https://github.com/schmittjoh/serializer/issues/904, but for now looks slower...must be doing something wrong

### DIFF
--- a/src/Accessor/DefaultAccessorStrategy.php
+++ b/src/Accessor/DefaultAccessorStrategy.php
@@ -20,6 +20,8 @@ declare(strict_types=1);
 
 namespace JMS\Serializer\Accessor;
 
+use GeneratedHydrator\Configuration;
+use JMS\Serializer\Metadata\ClassMetadata;
 use JMS\Serializer\Metadata\PropertyMetadata;
 
 /**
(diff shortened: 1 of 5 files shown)


goetas:
@Majkl578 What do you think?

goetas:
I have checked the approach used internally by ocraminus, in reality used `Closure::bind` ....  so it is not faster that what we have now... and is exactly `Closure::bind` :smile:  
Not sure if we makes sense to continue this experiment.

Majkl578:
> in reality used Closure::bind

Yes, that is true. But he does that at once for whole object, not per-property. Calling closure per property (imagine 30 properties per object) will be slower than calling it once for all :)

goetas:
the version in this PR is 10% slower that the current master... :confused: .... still to figureout why.

Majkl578:
As noted on Slack:

> ad your generated hydrator: you don't use production ready configuration (always generate PHP code + eval() it instead),

src/Accessor/DefaultAccessorStrategy.php
```diff
@@ -20,21 +20,47 @@
...
+    {
+        $class = get_class($object);
+
+        if (!isset($this->gen[$class])) {
+            $config = new Configuration($class);
+            $hydratorClass = $config->createFactory()->getHydratorClass();
+            $this->gen[$class] = new $hydratorClass();
+        }
+
+        return $this->gen[$class]->extract($object);
```
goetas: @Majkl578 generators are generated and cached, so is "production" ready after the warmup.

goetas: The code inside the "if" is never executed during the benchmark, so should not make it slower

Issue #932, opened 2018-05-01
Property access should not rely on PropertyMetadata but only on AccessorStrategyInterface


goetas:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | yes/no
| RFC?             | yes

Currently to access the value of a property, is done via `PropertyMetadata::getValue` but should be done only via `AccessorStrategyInterface` implementation and use `PropertyMetadata` as configuration object.

This could allow to avoid loading `ReflectionProperty` for each property and saving some memory.

PR #934, opened 2018-05-02
Do not use property metadata to get/set object values

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/serializer/issues/932
| License       | Apache-2.0

first draft of what could be a solution for  #932. To get some  extra memory improvements, even `jms/metadata` should be updated (removing the reflection stuff from property metadata)

### DIFF
--- a/src/Accessor/DefaultAccessorStrategy.php
+++ b/src/Accessor/DefaultAccessorStrategy.php
@@ -75,9 +75,6 @@ final class DefaultAccessorStrategy implements AccessorStrategyInterface
 
     public function setValue(object $object, $value, PropertyMetadata $metadata): void
     {
-        if ($metadata instanceof StaticPropertyMetadata || $metadata instanceof VirtualPropertyMetadata) {
-            throw new LogicException(sprintf('%s on %s is immutable.'), $metadata->name, $metadata->class);
-        }
         if ($metadata->readOnly) {
             throw new LogicException(sprintf('%s on %s is read only.'), $metadata->name, $metadata->class);
         }


src/Accessor/DefaultAccessorStrategy.php
```diff
@@ -20,21 +20,76 @@
...
+
+    public function __construct(ExpressionEvaluatorInterface $evaluator = null)
+    {
+        $this->evaluator = $evaluator;
+    }
 
     public function getValue(object $object, PropertyMetadata $metadata)
     {
-        return $metadata->getValue($object);
+        if ($metadata instanceof StaticPropertyMetadata) {
```
Majkl578: Nice. This bunch of IFs calls for a strategy pattern imho, this is a bit too coupled.

src/Accessor/DefaultAccessorStrategy.php
```diff
@@ -20,21 +20,76 @@
...
+use JMS\Serializer\Metadata\StaticPropertyMetadata;
+use JMS\Serializer\Metadata\VirtualPropertyMetadata;
 
 /**
  * @author Asmir Mustafic <goetas@gmail.com>
  */
 final class DefaultAccessorStrategy implements AccessorStrategyInterface
 {
+    private $readAccessors = array();
+    private $writeAccessors = array();
```
Majkl578: Inho one shared accessor is enough, no? They are bound to type only.

Majkl578:
As an experiment (perf), you could also try closures bound to specific object instances and/or properties.

src/Accessor/DefaultAccessorStrategy.php
goetas: this has to be fast! very fast! :thinking: 
not sure if having extra classes to load and call will influence on it

src/Accessor/DefaultAccessorStrategy.php
goetas: The body of the closure is differnet, 

```php
function ($o, $name, $value) {
   $o->$name = $value;
}
```
vs
```php
function ($o, $name {
   return $o->$name;
}
```
so cant be done

goetas:
> As an experiment (perf), you could also try closures bound to specific object instances and/or properties.

Can you explain?

Majkl578:
Right now you are using unbound accessors (2nd argument to Closure::bind() is null) so you have to pass around `$o` as an object instance.
If you bind the closure to specific object, you can simply use `$this->$propertyName` instead.

Majkl578:
Btw this needs some test cases for inheritance with private properties: with different names, with same names (but different serialized names) and so.

goetas:
> Right now you are using unbound accessors (2nd argument to Closure::bind() is null) so you have to pass around $o as an object instance.
If you bind the closure to specific object, you can simply use $this->$propertyName instead.

But if i do it per object is much slower... creating the closure takes a lot apparently.

goetas:
https://github.com/schmittjoh/serializer/pull/935 proves that does not work :)

PR #938, opened 2018-05-02
Move property ordering strategy out of Metadata

Majkl578:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | -
| BC breaks?    | no-ish?

This moves accessor ordering logic out of Metadata.
It has little value on its own currently, it just cleans up metadata responsibilities and opens further refactoring opportunities in Metadata.

No impact on runtime performance, it's only executed during Metadata load.

### DIFF
--- a/src/Ordering/AlphabeticalPropertyOrderingStrategy.php
+++ b/src/Ordering/AlphabeticalPropertyOrderingStrategy.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * Copyright 2016 Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


src/Ordering/AlphabeticalPropertyOrderingStrategy.php
```diff
@@ -0,0 +1,41 @@
...
+namespace JMS\Serializer\Ordering;
+
+use JMS\Serializer\Metadata\PropertyMetadata;
+
+final class AlphabeticalPropertyOrderingStrategy implements PropertyOrderingInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function order(array $properties) : array
```
Majkl578: Did not use reference because there is no reason to do so. :grinning:

goetas:
Nice!

Issue #942, opened 2018-05-04
Revert fix for #43 (JSON representation of empty objects)


Majkl578:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| RFC?             | yes

https://github.com/schmittjoh/serializer/blob/e75cdaf4a89ca7cd9f473c25ba0bf54c60889b5f/src/JsonSerializationVisitor.php#L112-L115

This piece of code is wrong. It breaks output for valid collections (especially first class collections with `@Inline` attribute).
It should be reverted or fixed properly, i.e. when not serializing inline collection.
This has caused troubles in past to me and consumers of my API since it violates API interface (which _returns collection of objects_ - `{}` is not an empty collection).

This was done in 66b26e97e9267ceb4cd8b47cbb8ce248d8b77b64 in order to fix #43. Unfortunately it has broken valid collections. According to the comments, it didn't even fix the root cause, only symptom for root objects.

goetas:
Did you check https://github.com/schmittjoh/serializer/blob/master/doc/cookbook/arrays.rst?

goetas:
Currently this is necessary to avoid empty objects being serialized as arrays. 
Can you send a failing test case?

Majkl578:
Here you go: #943 

This is the 2nd case from you example called _typed as strict array_.

goetas:
Closing as it is not a BC break or a but but a feature request as commented on https://github.com/schmittjoh/serializer/pull/943

Majkl578:
Whatever you say, this behavior is broken and violates JSON specification.

goetas:
The code you are suggesting to remove is used mainly to support https://github.com/schmittjoh/serializer/blob/e75cdaf4a89ca7cd9f473c25ba0bf54c60889b5f/tests/Serializer/JsonSerializationTest.php#L247 this test case. 
Do you know a better solution? 
I was thinking about having the `data` variable always instance of arrayobject. It could work but requires some extra fixes when merging inline properties

PR #943, opened 2018-05-05
Failing test for first class collections

Majkl578:
Failing test for #942 - empty first class collections serialized as object.

### DIFF
--- a/tests/Fixtures/FirstClassCollection.php
+++ b/tests/Fixtures/FirstClassCollection.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Tests\Fixtures;
+
+use JMS\Serializer\Annotation as Serializer;
+
+class FirstClassCollection implements \IteratorAggregate
+{
+    /**
+     * @Serializer\Type("array<int>")
+     * @Serializer\Inline
+     * @var int[]
+     */
+    public $items = [];
+
+    public function __construct(int ...$items)
+    {
+        $this->items = $items;
+    }
+
+    public function getIterator() : iterable
+    {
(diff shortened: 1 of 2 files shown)


tests/Fixtures/FirstClassCollection.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace JMS\Serializer\Tests\Fixtures;
+
+use JMS\Serializer\Annotation as Serializer;
+
+class FirstClassCollection implements \IteratorAggregate
+{
+    /**
+     * @Serializer\Type("array<int>")
+     * @Serializer\Inline
```
goetas: Currently, your first test case works just by coincidence. `@Inline` was never tested on collections.
If will also fail if your you unset one item from the collection (making keys not consequent).

And on top of it what it the expected result if you add one more property to the  `FirstClassCollection` ?

Majkl578: > what it the expected result if you add one more property

Depends. Either an error as there could only be one property with ``@Inline, or, if you consider things like exclusion strategy, error of there is more than one property after applying exclusion rules.

goetas:
However could be a nice feature to have. So PR are welcome

tests/Fixtures/FirstClassCollection.php
goetas: An error (doable at "metadata" discovery time) should be fine

goetas: Consider that the collection could be also an associative array, so it can be both `[]` and `{}`

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/944

PR #944, opened 2018-05-06
Allow inline lists and maps

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | https://github.com/schmittjoh/serializer/pull/943 
| License       | Apache-2.0

~This removes the "object" special case, making the codebase much more predictable.~

~@Majkl578 However does not solve your feature request since your `@Inline` trick will not work anymore and collections will be consistently serialized to json objects (unless special manipulation is done inside `endVisitingObject` for `@Inline`).~

### DIFF
--- a/src/JsonSerializationVisitor.php
+++ b/src/JsonSerializationVisitor.php
@@ -112,7 +112,7 @@ final class JsonSerializationVisitor extends AbstractVisitor implements Serializ
         $rs = $this->data;
         $this->data = $this->dataStack->pop();
 
-        if ($metadata->isList) {
+        if ($metadata->isList === true) {
             return array_values((array)$rs);
         }
 
--- a/src/JsonSerializationVisitor.php
+++ b/src/JsonSerializationVisitor.php
@@ -104,7 +104,7 @@ final class JsonSerializationVisitor extends AbstractVisitor implements Serializ
     public function startVisitingObject(ClassMetadata $metadata, object $data, array $type): void
     {
         $this->dataStack->push($this->data);
-        $this->data = new \ArrayObject();
+        $this->data = $metadata->isMap === true ? new \ArrayObject() : [];
     }
 
     public function endVisitingObject(ClassMetadata $metadata, object $data, array $type)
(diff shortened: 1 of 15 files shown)


goetas:
What about a syntax as this?

```php
/**
* @Serializer\Inline("array<int>")
*/
class FirstClassCollection implements \IteratorAggregate
{
    /**
     * @var int[]
     */
    public $items = [];

    public function __construct(int ...$items)
    {
        $this->items = $items;
    }

    public function getIterator() : iterable
    {
        yield from $this->items;
    }
}
```

goetas:
> since your @Inline trick will not work anymore and collections will be consistently serialized to json objects
> -1 That is completely wrong and serious BC break. Serializing collection as object is pure non-sense

The implementation proposed in this PR is better that the current inconsistent behavior.
Once is made consistent it will be possible to introduce a way to make it be `[]` or `{}` by using some configuration.

Majkl578:
I don't know how to express it clearer: there is nothing like _either `[]` or `{}` to configure_, arrays (as in lists) and objects (as in dictionaries/maps) are two different things and are not interchangeable.

To quote JSON specification from [RFC 8259](https://tools.ietf.org/html/rfc8259#section-5):

> 5.  Arrays
> 
>    An array structure is represented as square brackets surrounding zero
>    or more values (or elements).  Elements are separated by commas.

This is a list.


> 4.  Objects
> 
>    An object structure is represented as a pair of curly brackets
>    surrounding zero or more name/value pairs (or members).  A name is a
>    string.  A single colon comes after each name, separating the name
>    from the value.  A single comma separates a value from a following
>    name.  The names within an object SHOULD be unique.

This is a dictionary/map.

As a result, your statement "_collections will be consistently serialized to json objects_" is incompatible with how JSON works.

src/JsonSerializationVisitor.php
```diff
@@ -101,17 +104,16 @@ public function visitArray(array $data, array $type)
...
 
     public function endVisitingObject(ClassMetadata $metadata, object $data, array $type)
     {
         $rs = $this->data;
         $this->data = $this->dataStack->pop();
 
-        // Force JSON output to "{}" instead of "[]" if it contains either no properties or all properties are null.
-        if (empty($rs)) {
-            $rs = new \ArrayObject();
+        if ($metadata->isList) {
```
goetas: this is the whole trick

goetas:
but this makes the serializer 3-5% slower

goetas:
The latest implementation has no memory and performance impact when the inline list/map is not used

PR #949, opened 2018-05-12
Revert https://github.com/schmittjoh/serializer/pull/113

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | Apache-2.0

Removes extra complexity from the library by reverting https://github.com/schmittjoh/serializer/pull/113

Use dependency injection instead.

### DIFF
--- a/src/Serializer.php
+++ b/src/Serializer.php
@@ -273,12 +273,4 @@ final class Serializer implements SerializerInterface, ArrayTransformerInterface
 
         return $data;
     }
-
-    /**
-     * @return MetadataFactoryInterface
-     */
-    public function getMetadataFactory(): MetadataFactoryInterface
-    {
-        return $this->factory;
-    }
 }


gnutix:
@goetas Would you have any instructions on how to "use dependency injection instead" ? As far as I understand, this MetadataFactory class is instantiated "manually" in the build() method, and it's passed directly to the Serializer and then only used internally. It's never accessible from a dependency injection container. Or did I missed something ?

goetas:
If you are using symfony, the Metadata factory is available via dedicated service. If you are not using symfony, you can build your own metadata factory and pass that to the serializer builder.

(maybe could be nice to have a getMetadataFactory method in the serializer builder)

Issue #806, opened 2017-08-09
DoctrineObjectConstructor and deserialize not work


sashaaro:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Steps required to reproduce the problem
My serializer naming strategy with underscore 
I have entity "Category" with identifier filed "categoryId"
Json -`{categoryId: 1, discription_text: 'new discription'}`
When I do deserialize I recive full object from db and changed field discriptionText
It work fine.
But same json with `{ category_id: 2..... }`
after deserialize I got empty entity (but categoryId field = 2..and other ones I sent)
DoctrineObjectConstructor not work in this case

goetas:
i'm aware of that bug. cant find it now, but should be already reported by someone else time ago...

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/951

Issue #734, opened 2017-03-22
[Symfony] DoctrineObjectorConstructor always creates new entity because of camel case to snake case conversion


asgardia7:
Hello,

I am deserializing using 
`    jms_serializer.object_constructor:
         alias: jms_serializer.doctrine_object_constructor`.

Always new entity is generated instead of updated entity initaly loaded from database because this test fails [Identifiers test](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Construction/DoctrineObjectConstructor.php#L78) . Debugging it, I've seen the identifier is not recognized because it was originally `idPlanification` while after serialization and coming back from AJAX call, it becomes `id_planification`.

It's ok to convert camel case to snake case through serialization, but shouldn't this handled automatically through deserialization if so?
Or how can I automatically transform back my JSON string to inital camel case before it gets handled by the `DoctrineObjectorConstructor `

Thanks,

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/951

PR #951, opened 2018-05-17
Fix DoctrineObjectConstructor deserialization with naming strategies

re2bit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #806 https://github.com/schmittjoh/serializer/issues/734
| License       | MIT

Use NamingStrategy to determine which array key in data should be used.
This should fix Error #806.

Problem Analysis:
While Object construction $data is used as Source for Doctrine Identifier Fields.
Since $data is either a raw array the names are not converted yet.

### DIFF
--- a/src/JMS/Serializer/Construction/DoctrineObjectConstructor.php
+++ b/src/JMS/Serializer/Construction/DoctrineObjectConstructor.php
@@ -87,11 +87,12 @@ class DoctrineObjectConstructor implements ObjectConstructorInterface
         $identifierList = array();
 
         foreach ($classMetadata->getIdentifierFieldNames() as $name) {
-            if (!array_key_exists($name, $data)) {
+            $dataName = $metadata->propertyMetadata[$name]->serializedName ?: $name;
+            if (!array_key_exists($dataName, $data)) {
                 return $this->fallbackConstructor->construct($visitor, $metadata, $data, $type, $context);
             }
 
-            $identifierList[$name] = $data[$name];
+            $identifierList[$name] = $data[$dataName];
         }
 
         // Entity update, load it from database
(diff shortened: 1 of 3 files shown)


src/JMS/Serializer/Construction/DoctrineObjectConstructor.php
```diff
@@ -87,11 +88,19 @@ public function construct(VisitorInterface $visitor, ClassMetadata $metadata, $d
         $identifierList = array();
 
         foreach ($classMetadata->getIdentifierFieldNames() as $name) {
-            if (!array_key_exists($name, $data)) {
+            if (method_exists($visitor, 'getNamingStrategy')) {
```
goetas: cant you check if `$visitor instanceof AbstractVisitor`?

goetas:
Thanks a lot for the contribution. 
In case this gets merged are you ok with releasing it also as MIT license? (as we are migrating this project to a more permissive MIT license)

re2bit:
Sure MIT licence is fine.

src/JMS/Serializer/Construction/DoctrineObjectConstructor.php
re2bit: i`m going to change this. but i think it will break the current master since the method "getNamingStrategy" was removed.
See: https://github.com/schmittjoh/serializer/commit/9e3fd4d83a0e81d97da2f2cb12568a9858952007#diff-04059dbc4f97e37925ef18e828ce4f5b

i`m going to rework this merge request this Sunday.
i hope this pull request helps you to pinpoints the error from #806

goetas:
@re2bit do you think https://github.com/schmittjoh/JMSSerializerBundle/issues/687 could be connected to your work done here?

PR #954, opened 2018-05-21
Fix bug deserialising inline key-value pairs

CyberiaResurrection:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | NIL
| License       | Apache-2.0

We made this change to enable the POData project to work properly.  
We needed to deserialise inline KVPs and as we found it, the JMS serialiser didn't do that.


CyberiaResurrection:
Further investigation reveals even more work needed

PR #903, opened 2018-04-11
[2.0] Compiled navigator

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   |no
| License       | Apache-2.0

PR #962, opened 2018-06-07
Make access strategies aware of the context

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | not yet
| Deprecations? | no
| Tests pass?   | yes,  but need new tests
| Fixed tickets | -
| License       | MIT

### DIFF
--- a/src/Accessor/AccessorStrategyInterface.php
+++ b/src/Accessor/AccessorStrategyInterface.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace JMS\Serializer\Accessor;
 
+use JMS\Serializer\DeserializationContext;
 use JMS\Serializer\Metadata\PropertyMetadata;
+use JMS\Serializer\SerializationContext;
 
 /**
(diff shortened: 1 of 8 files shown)


goetas:
@Majkl578 What do you think about this? 

Has the drawback that making access strategies compiled in the future will be more difficult... but I had many cases where I needed some extra info when using custom access (as expression language as example)

goetas:
Another way would be introducing and `AccessorStrategyFactory` that is invoked in the serialization step (instead of using the current single object instance) to produce the `AccessorStrategyInterface` object

goetas:
as explored in https://github.com/schmittjoh/serializer/pull/964 https://github.com/schmittjoh/serializer/pull/903 https://github.com/schmittjoh/serializer/pull/933 , it is not easy to have compiled graph navigators, exclusion strategies or access strategies... and even more weird, performance improvements are not really visible... 

Becase of that, I think can be a fine.

PR #968, opened 2018-06-19
[WIP] Apply Doctrine CS

Majkl578:
Initial import of Doctrine CS.

@goetas Excluded some rules you mentioned before that you don't like.

Also excluded BC-breaking rules since the code base is currently a bit messy, a lot of inconsistent/missing parameter/return types. :(

Only applied to src/ to give you some look at the result, and to have something to discuss/improve later on.

### DIFF
--- a/src/EventDispatcher/EventDispatcherInterface.php
+++ b/src/EventDispatcher/EventDispatcherInterface.php
@@ -9,11 +9,7 @@ interface EventDispatcherInterface
     /**
      * Returns whether there are listeners.
      *
-     * @param string $eventName
-     * @param string $class
-     * @param string $format
      *
-     * @return boolean
      */
     public function hasListeners(string $eventName, string $class, string $format): bool;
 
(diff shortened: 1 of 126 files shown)


goetas:
Is there a way to disable the `=` alignment?

src/EventDispatcher/EventDispatcherInterface.php
```diff
@@ -23,31 +19,18 @@ public function hasListeners(string $eventName, string $class, string $format):
...
      * were added to the dispatcher.
      *
-     * @param string $eventName
-     * @param string $class
-     * @param string $format
-     * @param Event $event
-     * @return void
      */
-    public function dispatch($eventName, string $class, string $format, Event $event): void;
+    public function dispatch(string $eventName, string $class, string $format, Event $event): void;
```
goetas: good catch, this is a bug! :)

Majkl578:
> Is there a way to disable the `=` alignment?

yes

> Why remove the `@author` annotation?

It's totally useless and very often outdated, Git already provides more accurate information.
See https://github.com/doctrine/coding-standard/pull/9#discussion_r137123497 + https://github.com/doctrine/coding-standard/pull/9#issuecomment-353438394

src/EventDispatcher/EventDispatcherInterface.php
Majkl578: There is more of missing types, some break more things so I would probably not touch types in first iteration and do them separately, then add proper docblocks for properties etc. (this is especially in extremely bad shape in Serializer).
Then we can also consider PHPStan.

Majkl578:
Also please decide the following:
* yoda conditions: enforce or forbid
* assignments in conditions: allow or not

goetas:
How to run the comment to apply the rules? is it documented somewhere?

Majkl578:
https://www.doctrine-project.org/projects/doctrine-coding-standard/en/latest/reference/index.html#composer-project-dependency

Since the project has `phpcs.xml.dist`, it's enough to simply run:
* `vendor/bin/phpcs` to validate CS
* `vendor/bin/phpcbf` to fix fixable errors according to CS

goetas:
will try, tnx!

goetas:
implemented in https://github.com/schmittjoh/serializer/pull/971

PR #969, opened 2018-06-21
Bugfix/metadata serialization

supersmile2009:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no (N/A)
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes (fixed/added tests)
| Fixed tickets | N/A
| License       | MIT

Serialize/unserialize methods in each PropertyMetadata class was completely independent, which is difficult to maintain and keep track of. It resulted in a fact that some of these methods were not updated and some properties were ommited during serialization.

I encountered the issue when serializing a VirtualProperty with SkipWhenEmpty. Initially extracted metadata was OK, but after caching it and retrieving back from cache metadata was corrupted - `skipWhenEmpty` property was `false` again.

In my fix I tried to reuse property serialization code as much as possible without reorganizing class inheritance to preserve backwards compatibility.
I also added unit tests to verify property metadata serialization and deserialization.

The final run of the test suite also revealed a bug in YamlSerializationVisitor - SkipWhenEmpty was not respected in it, so I fixed that as well.

### DIFF
--- a/src/JMS/Serializer/Metadata/ExpressionPropertyMetadata.php
+++ b/src/JMS/Serializer/Metadata/ExpressionPropertyMetadata.php
@@ -46,66 +46,23 @@ class ExpressionPropertyMetadata extends PropertyMetadata
     public function serialize()
     {
         return serialize(array(
-            $this->sinceVersion,
-            $this->untilVersion,
-            $this->groups,
-            $this->serializedName,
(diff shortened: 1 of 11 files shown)


goetas:
Thanks!

PR #970, opened 2018-06-25
[WIP] Add XML support for DoctrineObjectConstructor and fix some tests.

re2bit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

Tests fail on Master and Doctrine Object Constructor does not support deserialize XML documents since $data is not an array but \SimpleXml Object.

@goetas i would love to receive your feedback.


tests/Serializer/Doctrine/ObjectConstructorTest.php
```diff
@@ -295,6 +337,25 @@ private function createSerializerWithDoctrineObjectConstructor()
...
     }
+
+    /**
+     * @param string $className
+     * @return ClassMetadata
+     * @throws \Doctrine\Common\Annotations\AnnotationException
+     * @throws \JMS\Serializer\Exception\InvalidMetadataException
+     * @throws \ReflectionException
+     */
+    private function createClassMetadataFor($className): ClassMetadata
```
goetas: you sure this is necessary? the default serializer metadata loader should fallback to annotations anyway

tests/Serializer/Doctrine/ObjectConstructorTest.php
```diff
@@ -209,6 +213,44 @@ public function testNamingForIdentifierColumnIsConsidered()
...
+        $serializer = $this->createSerializerWithDoctrineObjectConstructor();
+
+        /** @var EntityManager $em */
+        $em = $this->registry->getManager();
+        $server = new Server('Linux', '127.0.0.1', 'home');
+        $em->persist($server);
+        $em->flush();
+        $em->clear();
+
+        $jsonData = '<?xml version="1.0" encoding="UTF-8"?>
```
goetas: `$xmlData`

src/Construction/DoctrineObjectConstructor.php
```diff
@@ -63,7 +68,16 @@ public function construct(DeserializationVisitorInterface $visitor, ClassMetadat
         }
 
         // Managed entity, check for proxy load
-        if (!\is_array($data)) {
+
+        if ($data instanceof \SimpleXMLElement) {
```
goetas: instead of creating a new array and later trying to match between $attributes and $properties, why not transform the $data (simplexml object) into a plain $data array ?

goetas: this should make not necessary alll the changes you had to do inside the foreach loop.

src/Construction/DoctrineObjectConstructor.php
```diff
@@ -72,18 +86,50 @@ public function construct(DeserializationVisitorInterface $visitor, ClassMetadat
...
+                'No cache driver implementation for the mapping metadata is given',
+                1526462942588
+            );
+        }
+        $reader = $driver->getReader();
+
         foreach ($classMetadata->getIdentifierFieldNames() as $name) {
+            $searchArray = $properties ?? $data;
+            $reflectionProperty = $classMetadata->getReflectionClass()->getProperty($name);
+            $isAttribute = (bool)$reader->getPropertyAnnotation(
```
goetas: can not rely on this, people might be using yaml or xml metadatata, while this considers only annotations.

if you really want to look into the metadata, you have the `$metadata` for it

re2bit: Your right. I was blind since we only use Annotations as Metadata Source.

tests/Serializer/Doctrine/ObjectConstructorTest.php
re2bit: As far as i can tell yes. the Constructor for ClassMetadata does not initialize anything by himself.
The old Code
```
$class = new ClassMetadata(Author::class);
```
Will result in an "empty" Class Metadata. For example the propertyMetadata array will simply be empty.
We need to initialize the Metadata since it is used for the extraction for the Identifier Field.

Is there a better way to Create a fully initialized ClassMetadata?

src/Construction/DoctrineObjectConstructor.php
re2bit: Since an Attribute and a Property could have the same Name, I decided to distinguish between them.
```
<body id="body">
  <id>12223</id>
</body>
```

re2bit:
Since I did not find the time to finish this pull request, I will close it here.
I will try to provide my gained experiences from the last weeks into a new pull request.

Issue #914, opened 2018-04-26
Adopt some coding standard


Majkl578:
| Q                | A
| ---------------- | -----
| RFC?             | yes

Would be nice to have some deterministic CS used in this library. My humble suggestion is [Doctrine CS](doctrine/coding-standard), but if you have any other specific in mind, let me know.
If you agree, I'll provide a PR (or ask someone to do so). :blush:

goetas:
hmm sounds a good thing! (would be possible to add also some kind of build step to make it fail if the code is not well formatted?)

I guess something based on https://github.com/FriendsOfPHP/PHP-CS-Fixer or you had a different idea?

Will have a look to DoctrineCS... I'm using for private work something closer to SymfonyCS to evaluate which one should require less changes

goetas:
https://github.com/schmittjoh/serializer/pull/971 is merged

PR #971, opened 2018-07-01
Code style

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/serializer/issues/914
| License       | MIT

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+# Contribute
+
+Thank you for contributing!
+
+Before we can merge your Pull-Request here are some guidelines that you need to follow. 
+These guidelines exist not to annoy you, but to keep the code base clean, unified and future proof.
+
+## Coding Standard
+
+This project uses [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) to enforce coding standards.
+The coding standard rules are defined in the **phpcs.xml.dist** file (part of this repository).
+The project follows a relaxed version of the Doctrine Coding standards v4.
+
+Your Pull-Request must be compliant with the said standard.
+To check your code you can run `vendor/bin/phpcs`. This command will give you a list of violations in your code (if any).
+The most common errors can be automatically fixed just by running `vendor/bin/phpcbf`.
+
+## Unit-Tests
+
+Please try to add a test for your pull-request. This project uses [PHPUnit](https://phpunit.de/) as testing framework.
+
+You can run the unit-tests by calling `vendor/bin/phpunit`.
+
+New features without tests can't be merged.
+
+## CI
+
+We automatically run your pull request through [Travis CI](https://www.travis-ci.org) 
+and [Scrutinizer CI](https://scrutinizer-ci.com/).
+If you break the tests, we cannot merge your code,
+so please make sure that your code is working before opening up a Pull-Request.
+
+## Issues and Bugs
+
+To create a new issue, you can use the GitHub issue tracking system.
+Please try to avoid opening support-related tickets. For support related questions please use more appropriate
+channels as Q&A platforms (such as Stackoverflow), Forums, Local PHP user groups.
+
+If you are a Symfony user, please try to distinguish between issues related to the Bundle and issues related to this 
+library.  
+
+## Getting merged
+
+Please allow us time to review your pull requests.
+We will give our best to review everything as fast as possible, but cannot always live up to our own expectations.
+
+Thank you very much again for your contribution!
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -26,7 +26,6 @@
         <exclude name="Generic.Formatting.MultipleStatementAlignment.NotSame"/>
         <exclude name="Generic.Formatting.MultipleStatementAlignment"/>
 
-        <exclude name="Squiz.Strings.DoubleQuoteUsage.ContainsVar"/>
         <exclude name="Squiz.Arrays.ArrayDeclaration.KeySpecified"/>
 
         <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -21,9 +21,8 @@
         <exclude name="SlevomatCodingStandard.ControlStructures.DisallowYodaComparison.DisallowedYodaComparison"/>
         <exclude name="SlevomatCodingStandard.ControlStructures.AssignmentInCondition.AssignmentInCondition"/>
 
-        <exclude name="SlevomatCodingStandard.Commenting.RequireOneLinePropertyDocComment.MultiLinePropertyComment"/>
+        <exclude name="SlevomatCodingStandard.Commenting.RequireOneLinePropertyDocComment"/>
 
-        <exclude name="Generic.Formatting.MultipleStatementAlignment.NotSame"/>
         <exclude name="Generic.Formatting.MultipleStatementAlignment"/>
 
         <exclude name="Squiz.Arrays.ArrayDeclaration.KeySpecified"/>
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -30,7 +30,6 @@
         <exclude name="Squiz.Commenting.FunctionComment.SpacingAfterParamType"/>
 
         <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
-        <exclude name="SlevomatCodingStandard.TypeHints.NullableTypeForNullDefaultValue"/>
 
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableParameterTypeHintSpecification"/>
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableReturnTypeHintSpecification"/>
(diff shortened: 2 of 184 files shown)


phpcs.xml.dist
```diff
@@ -0,0 +1,97 @@
...
+
+
+        <exclude name="Squiz.Strings.DoubleQuoteUsage.ContainsVar"/>
+        <exclude name="Squiz.Arrays.ArrayDeclaration.KeySpecified"/>
+
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableParameterTypeHintSpecification"/>
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableReturnTypeHintSpecification"/>
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversablePropertyTypeHintSpecification"/>
+
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint"/>
```
goetas: @Majkl578 I would like to use this but the `serialize` and `unserialize` methods should be ignored bu this rule... is it possible?

Majkl578: Yes, ideally by adding this into method's annotation:
```
@phpcsSuppress SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint
```
else you would need to exclude whole file here.

phpcs.xml.dist
```diff
@@ -0,0 +1,97 @@
...
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming"/>
+        <exclude name="SlevomatCodingStandard.ControlStructures.DisallowYodaComparison.DisallowedYodaComparison"/>
+        <exclude name="Generic.Formatting.MultipleStatementAlignment.NotSame"/>
+        <exclude name="SlevomatCodingStandard.ControlStructures.AssignmentInCondition.AssignmentInCondition"/>
+
+        <exclude name="SlevomatCodingStandard.Commenting.RequireOneLinePropertyDocComment.MultiLinePropertyComment"/>
+
+        <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
+        <exclude ref="Generic.Formatting.MultipleStatementAlignment"/>
+        <exclude ref="Generic.Formatting.MultipleStatementAlignment"/>
```
Majkl578: duplicated exclude

phpcs.xml.dist
goetas: My hope was that was possible to exclude it by some class/method combo. :( 
Will try

Majkl578: I don't think it's currently possible, but it would be good idea for CodeSniffer 4. :)

goetas:
@Majkl578 to you see some other "easy to implment" rule that should be added (https://github.com/schmittjoh/serializer/pull/971/files#diff-6758ef1f416bd888bd1a5cad8f97cc2b)?

goetas:
@schmittjoh I'm planning to merge this soon... any comments on it from your side?

phpcs.xml.dist
```diff
@@ -0,0 +1,88 @@
...
+
+    <file>src/</file>
+
+    <rule ref="Doctrine">
+        <exclude name="Generic.Formatting.SpaceAfterNot"/>
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming"/>
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousInterfaceNaming"/>
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming"/>
+
+        <exclude name="SlevomatCodingStandard.ControlStructures.DisallowYodaComparison.DisallowedYodaComparison"/>
```
Majkl578: Since you disable `DisallowYodaComparison`, you should enable the inverse sniff: `RequireYodaComparison` so the code is consistent.

phpcs.xml.dist
```diff
@@ -0,0 +1,88 @@
...
+        <exclude name="SlevomatCodingStandard.ControlStructures.AssignmentInCondition.AssignmentInCondition"/>
+
+        <exclude name="SlevomatCodingStandard.Commenting.RequireOneLinePropertyDocComment.MultiLinePropertyComment"/>
+
+        <exclude name="Generic.Formatting.MultipleStatementAlignment.NotSame"/>
```
Majkl578: As you exclude whole sniff below, this exclude is redundant.

phpcs.xml.dist
```diff
@@ -0,0 +1,88 @@
...
+    <rule ref="Doctrine">
+        <exclude name="Generic.Formatting.SpaceAfterNot"/>
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming"/>
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousInterfaceNaming"/>
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming"/>
+
+        <exclude name="SlevomatCodingStandard.ControlStructures.DisallowYodaComparison.DisallowedYodaComparison"/>
+        <exclude name="SlevomatCodingStandard.ControlStructures.AssignmentInCondition.AssignmentInCondition"/>
+
+        <exclude name="SlevomatCodingStandard.Commenting.RequireOneLinePropertyDocComment.MultiLinePropertyComment"/>
```
Majkl578: It's preferable to exclude whole sniff unless you are excluding only a specific error (and this sniff only has this single error):
```xml
<exclude name="SlevomatCodingStandard.Commenting.RequireOneLinePropertyDocComment"/>
```

phpcs.xml.dist
```diff
@@ -0,0 +1,88 @@
...
+
+        <exclude name="Generic.Formatting.MultipleStatementAlignment.NotSame"/>
+        <exclude name="Generic.Formatting.MultipleStatementAlignment"/>
+
+        <exclude name="Squiz.Arrays.ArrayDeclaration.KeySpecified"/>
+
+        <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
+        <exclude name="SlevomatCodingStandard.TypeHints.NullableTypeForNullDefaultValue"/>
```
Majkl578: This one **should** be kept enabled and enforced.

Non-nullable types without default value null are a legacy behavior that should not be relied upon.
See discussion here for reasoning: https://github.com/doctrine/common/pull/805#discussion_r128721368

Also to quote [Typed Properties RFC](https://wiki.php.net/rfc/typed_properties_v2) (targeting 7.3):
> Typed properties **cannot have a null default value, unless the type is explicitly nullable (?Type)**. This is in contrast to parameter types, where a null default value automatically implies a nullable type. **We consider this to be a legacy behavior**, which we do not wish to support for newly introduced syntax.

phpcs.xml.dist
```diff
@@ -0,0 +1,88 @@
...
+
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableParameterTypeHintSpecification"/>
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableReturnTypeHintSpecification"/>
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversablePropertyTypeHintSpecification"/>
```
Majkl578: You should enable all these three and properly document the methods. Otherwise the code is untyped mess and breaks static analysis for example (not just here, but in consumer projects.)

goetas: I would really like to have this but the array for the type `['name' => 'foo', 'params' => [...]]` makes it impossible to apply.. and putting `mixed[]` defeats the propose of it.

goetas: I have not better ideas...

phpcs.xml.dist
goetas: Personally prefer to not bother here... too much code to change...

goetas:
@Majkl578 thanks for the review

phpcs.xml.dist
Majkl578: True, `mixed[]` is not much better. It usually indicated bad design (we have some similar issues in Doctrine) - should eventually be addressed by collections/bags/configurations.
Also the fact that PHP doesn't has _tuples_ doesn't help the cause. 🤔

phpcs.xml.dist
goetas: done

CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
...
+
+Please try to add a test for your pull-request. This project uses [PHPUnit](https://phpunit.de/) as testing framework.
+
+You can run the unit-tests by calling `vendor/bin/phpunit`.
+
+New features without tests can't be merged.
+
+## CI
+
+We automatically run your pull request through [Travis CI](http://www.travis-ci.org)
```
Majkl578: HTTPS please

CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
...
+These guidelines exist not to annoy you, but to keep the code base clean, unified and future proof.
+
+## Coding Standard
+
+This project uses [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) to enforce coding standards.
+The coding standard rules are defined in the **phpcs.xml.dist** file (part of this repository).
+The project follows a relaxed version of the Doctrine Coding standards v4.
+
+Your Pull-Request must be compliant with the said standard.
+To check your code you can run `vendor/bin/phcs`. This command will give you a list of violations in your code (if any).
```
Majkl578: typo: `phpcs`

goetas: super fast! tnx!

goetas:
I thinks is good to be merged!

PR #977, opened 2018-07-13
Add type "mixed" to configure arrays of mixed values

supersmile2009:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes (added tests)
| Fixed tickets | n/a
| License       | MIT

This PR adds support for type `mixed` in serializer type annotations.
It is particularly useful when configuring arrays, when you want to explicitly configure associative array of mixed values - `array<string, mixed>`, or when you want to serialize an associative array as simple indexed array - `array<mixed>.
Setting `mixed` type to a property is the same as no type at all.

### DIFF
--- a/doc/cookbook/arrays.rst
+++ b/doc/cookbook/arrays.rst
@@ -34,12 +34,14 @@ In case of a JSON serialization:
     $serializer->serialize([1, 2], SerializationContext::create()->setInitialType('array<integer>')); //  [1, 2]
     $serializer->serialize(['a', 'b'], SerializationContext::create()->setInitialType('array<integer>')); //  ['a', 'b']
     $serializer->serialize(['c' => 'd'], SerializationContext::create()->setInitialType('array<string>')); //  ["d"]
+    $serializer->serialize(['c' => 'd', 'i' => 1], SerializationContext::create()->setInitialType('array<mixed>')); //  ["d", 1]
 
     // typehint as hash, keys will be always considered
     $serializer->serialize([], SerializationContext::create()->setInitialType('array<integer,integer>')); //  {}
     $serializer->serialize([1, 2], SerializationContext::create()->setInitialType('array<integer,integer>')); //  {"0" : 1, "1" : 2}
     $serializer->serialize(['a', 'b'], SerializationContext::create()->setInitialType('array<integer,integer>')); //  {"0" : "a", "1" : "b"}
-    $serializer->serialize(['c' => 'd'], SerializationContext::create()->setInitialType('array<string,string>')); //  {"d" : "d"}
+    $serializer->serialize(['c' => 'd'], SerializationContext::create()->setInitialType('array<string,string>')); //  {"c" : "d"}
+    $serializer->serialize(['c' => 'd', 'i' => 1], SerializationContext::create()->setInitialType('array<string,mixed>')); //  {"c" : "d", "i" : 1}
 
 
 .. note ::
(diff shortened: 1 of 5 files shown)


goetas:
Instead of doing the check at runtime, why not transform `mixed` into `null` directly when doing the type parsing? see https://github.com/schmittjoh/serializer/blob/b2e7f4cbbf74ef8bc7849c51567a42334ce9f85f/src/JMS/Serializer/TypeParser.php#L66

goetas:
Ah, i forgot to say, thanks for the PR :smiley: !

supersmile2009:
Hi, @goetas.
Thanks for the tip, I checked out an option that you suggest.

If I add this into `TypeParser::parseInternal()`
```
        if ($typeName === 'mixed') {
            return null;
        }
```
Then empty arrays with `array<string,mixed>` will not be serialized as JSON objects.
If you look into JsonSerializationVisitor::visitArray(), there is such code:
```
        $isHash = isset($type['params'][1]);  // $type['params'][1] will be null, because we return null in TypeParser

        if (null === $this->root) {
            $this->root = $isHash ? new \ArrayObject() : array(); // So $isHash is false, thus root is array, not object.
            $rs = &$this->root;
        } else {
            $rs = $isHash ? new \ArrayObject() : array();
        }
```
It can be fixed by changing `$isHash = isset($type['params'][1]);` to
```
$isHash = isset($type['params'][1]) || \array_key_exists($type['params'][1]);
``` 


Alternatively I could use this code in TypeParser
```
        if ($typeName === 'mixed') {
            typeName = null;
        }
       // .......
       // which will eventually return this
        return array('name' => $typeName, 'params' => $params);
```
But then in GraphNavigator::accept() i'll have to change this code
```
        if (null === $type || 'mixed' === $type['name']) {
```
to
```
        if (null === $type || null === $type['name']) {
```

With these 2 alternatives we still have do extra checks.
So at the moment my solution seems to be the simplest and the most straightforward, because you get in parsed types what you actually wrote in annotation - potentially easier for debugging by people who are not familiar with that code and that particular override.

However if you like alternative solution more, I'll update my PR, just let me know which one.

goetas:
I did some tests on my own, and the feature works well... However  there are some issues with yaml... 
What do you think about placing this in the 2.x release  (hopefully in the next 2-3 weeks)? In 2.x yaml has been dropped, so the problem solves itself.

Regarding your possible solutions, I still like the extra check in the type parser. The graph navigator should handle less things possible. 

The more things are moved to compiled metadata the better

goetas:
Something that I've realised just now is that this feature will not work for deserialization

aheissenberger:
How should I handle a mixed array without this feature implemented?
I tried `array<string,int|string>` but this is also not supported.

supersmile2009:
The proper way of solving it would be redesigning your API or serialization schema so that you don't need mixed value types. We needed it for one particular piece of legacy API and applied the change from this PR as a composer patch. But this is a very poor API design. A better approach would be either having a consistent value type in your array, or explicitly documenting each object key and type.

We've redesigned the API since I've originally posted it, so I have no interest in pursuing it any further. @aheissenberger feel free to pick it up and implement a solution discussed above. But considering that it can't properly work for deserialization @goetas may never merge it anyway, which would be a reasonable decision. Sometimes not trying to solve silly problems is a better solution. You could also apply this change as is as a patch. We used to run it in prod for a while with no issues. Or just redesign your API if/when you have an opportunity.

PR #979, opened 2018-07-24
Add new ContainerObjectContructor

MaksSlesarenko:
Add new ContainerObjectContructor that allows to create new objects from di container services. For ex in symfony this will allow autowiring in deserialized entity

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no
| License       | MIT

### DIFF
--- a/src/Construction/ContainerObjectConstructor.php
+++ b/src/Construction/ContainerObjectConstructor.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace JMS\Serializer\Construction;
+
+use JMS\Serializer\DeserializationContext;
+use JMS\Serializer\Metadata\ClassMetadata;
+use JMS\Serializer\VisitorInterface;
+use Psr\Container\ContainerInterface;
+
+/**
+ * Container object constructor for new (or existing) objects during deserialization.
+ */
+class ContainerObjectConstructor implements ObjectConstructorInterface
+{
+    private $container;
+    private $fallbackConstructor;
+
+    /**
+     * Constructor
+     *
+     * @param ContainerInterface $container
+     * @param ObjectConstructorInterface $fallbackConstructor
+     */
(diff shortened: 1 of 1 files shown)


goetas:
Hi. 
Thanks for the PR.

I'm not sure about this feature.... autowiring of an entity? Can you explain a bit better the use case? 

Is your PR related to https://github.com/schmittjoh/serializer/issues/79 ?

Majkl578:
I am against this addition, it breaks assumptions of serialization/deserialization concept: both of these processes are meant to bypass the constructor.

MaksSlesarenko:
@goetas 
I do not think it is related  #79 

You see when using ParamConverter I can convert request params to DTO object and it also gets validated, so in controller I will receive 

`public function putPostAction(Post $post, ConstraintViolationListInterface $validationErrors)` 

- deserialized object and validation errors. That is good part. 

Bad part is that current implementation of JMS/Serialiser does this

 `$object = new $className` 

This code breaks dependency inversion principle and it constructs object from class by itself, leaving me without any options to bypass a service or model manager to validate DTO object against database.


https://symfony.com/doc/master/bundles/FOSRestBundle/request_body_converter_listener.html

MaksSlesarenko:
Maybe this should be placed in serializer-bundle, to fix dependency inversion principle that is used by symfony ?

goetas:
@MaksSlesarenko 
> Bad part is that current implementation of JMS/Serialiser does this
> $object = new $className

not true. jms serializer deos not call the constructor, it uses reflection to create a new object.

goetas:
@MaksSlesarenko 
What you are doing is more a FOS rest bundle feature to allow validation and deserialization of an object.

PR #1004, opened 2018-10-28
Allow deserializing of null

kunicmarko20:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| License       | MIT

    
 ## To do
    
- [x] Update documentation
- [x] Update all drivers
- [x] Update all visitors
- [x] Add tests

### DIFF
--- a/src/GraphNavigator/DeserializationGraphNavigator.php
+++ b/src/GraphNavigator/DeserializationGraphNavigator.php
@@ -21,6 +21,7 @@ use JMS\Serializer\GraphNavigator;
 use JMS\Serializer\GraphNavigatorInterface;
 use JMS\Serializer\Handler\HandlerRegistryInterface;
 use JMS\Serializer\Metadata\ClassMetadata;
+use JMS\Serializer\Metadata\PropertyMetadata;
 use JMS\Serializer\NullAwareVisitorInterface;
 use JMS\Serializer\Visitor\DeserializationVisitorInterface;
 use Metadata\MetadataFactoryInterface;
@@ -107,9 +108,24 @@ final class DeserializationGraphNavigator extends GraphNavigator implements Grap
         if (null === $type) {
             throw new RuntimeException('The type must be given for all properties when deserializing.');
         }
+
+        $shouldDeserializeNull = false;
+        $metadataStack = $this->context->getMetadataStack();
+
+        if (!$metadataStack->isEmpty()) {
+            $currentPropertyMetadata = $metadataStack->top();
+            if ($currentPropertyMetadata instanceof PropertyMetadata) {
+                $shouldDeserializeNull = $currentPropertyMetadata->deserializeNull;
+            }
+        }
+
         // Sometimes data can convey null but is not of a null type.
         // Visitors can have the power to add this custom null evaluation
-        if ($this->visitor instanceof NullAwareVisitorInterface && true === $this->visitor->isNull($data)) {
+        // If null is explicitly allowed we should skip this
+        if ($this->visitor instanceof NullAwareVisitorInterface
+            && true === $this->visitor->isNull($data)
+            && !$shouldDeserializeNull
+        ) {
             $type = ['name' => 'NULL', 'params' => []];
         }
 
--- a/tests/Fixtures/ObjectWithNullObject.php
+++ b/tests/Fixtures/ObjectWithNullObject.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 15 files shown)


src/GraphNavigator/DeserializationGraphNavigator.php
```diff
@@ -107,9 +108,24 @@ public function accept($data, ?array $type = null)
...
+            }
+        }
+
         // Sometimes data can convey null but is not of a null type.
         // Visitors can have the power to add this custom null evaluation
-        if ($this->visitor instanceof NullAwareVisitorInterface && true === $this->visitor->isNull($data)) {
+        // If null is explicitly allowed we should skip this
+        if ($this->visitor instanceof NullAwareVisitorInterface
+            && true === $this->visitor->isNull($data)
+            && !$shouldDeserializeNull
```
goetas: Can you move `!$shouldDeserializeNull` as first condition? so in case is true we do not need to call `isNull`

tests/Fixtures/ObjectWithNullObject.php
```diff
@@ -0,0 +1,26 @@
...
+
+use JMS\Serializer\Annotation\DeserializeNull;
+use JMS\Serializer\Annotation\Type;
+
+class ObjectWithNullObject
+{
+    /**
+     * @var null
+     * @Type("NullObject")
+     * @DeserializeNull()
```
goetas: The annotation `DeserializeNull` is fine IMO, but what if we add also `SerializeNull` when doing the serialization.. can we find a name for a single annotation that allows to decide how nulls are handled, both on serialization and deserialization?

tests/Serializer/BaseSerializationTest.php
```diff
@@ -191,6 +192,22 @@ public function testDeserializeNullObject()
         self::assertNull($dObj->getNullProperty());
     }
 
+    public function testDeserializeNullObject()
```
goetas: `testDeserializeNullObjectHandler` ?

src/GraphNavigator/DeserializationGraphNavigator.php
```diff
@@ -107,9 +108,24 @@ public function accept($data, ?array $type = null)
         if (null === $type) {
             throw new RuntimeException('The type must be given for all properties when deserializing.');
         }
+
+        $shouldDeserializeNull = false;
+        $metadataStack = $this->context->getMetadataStack();
+
+        if (!$metadataStack->isEmpty()) {
```
goetas: Really not convinced by this `if`. Both for performance and clarity. But I undersatdn that this lines are the "core" of how the whole feature works... Am I right?

goetas: If you could break compatibility, what would you do?

tests/Fixtures/ObjectWithNullObject.php
kunicmarko20: Hm, the only thing that pops into my mind is:

```
@NullValue(serialize=false, deserialize=false)
```

src/GraphNavigator/DeserializationGraphNavigator.php
kunicmarko20: I don't like it either but the problem was that we need the `deserializeNull` option in both GraphNavigator and Visitor, and this is the only way we can get the current property that is being deserialized and its option.

>If you could break compatibility, what would you do?

Can answer this in a few days when I investigate it better

goetas:
Reading your PR few minutes ago I have realized that something could be wrong...
The current behavior of ignoring NULLs allows to do PATCH operations on an object

If `deserialize_null` is defined in the metadata, it will not allow to to PUT and PATCH with the same metadata.

Said that, I think that "serialize null" is good as it is and we should find a way to make also "deserialize null" working via context.
Otherwise PUT will need some metadata and PATCH others.

I realize that this changes many things, but this came to my mind
What do you think?

tests/Fixtures/ObjectWithNullObject.php
goetas: Another question. What happens if `NullObject` is an exiting class and there is no handler for it? 

From what I can see, 
```
{
 "nullProperty": null
}
```
When unserialized, nullProperty  will be an object instance of `NullObject`, and that should not happen

kunicmarko20:
Don't have time to work on this anymore, will close it. Thank you for helping @goetas and sorry, maybe next time.

PR #1005, opened 2018-10-31
Allow deserializing null

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

Alternative implementation of https://github.com/schmittjoh/serializer/pull/1004


goetas:
The main trick here are the interaction between the changes in the graph navigator and the visitor

tests/Serializer/GraphNavigatorTest.php
```diff
@@ -151,7 +156,7 @@ protected function setUp()
         $this->serializationNavigator->initialize($this->serializationVisitor, $this->context);
 
         $this->deserializationNavigator = new DeserializationGraphNavigator($this->metadataFactory, $this->handlerRegistry, $this->objectConstructor, $this->accessor, $this->dispatcher);
-        $this->deserializationNavigator->initialize($this->deserializationVisitor, $this->context);
+        $this->deserializationNavigator->initialize($this->deserializationVisitor, $this->deserializeContext);
```
goetas: this was just a typo in the previous test

tests/Serializer/BaseSerializationTest.php
```diff
@@ -726,14 +746,14 @@ public function testDeserializingNull()
         self::assertEquals($this->getContent('blog_post_unauthored'), $this->serialize($post, SerializationContext::create()->setSerializeNull(true)));
 
         if ($this->hasDeserializer()) {
-            $deserialized = $this->deserialize($this->getContent('blog_post_unauthored'), get_class($post), DeserializationContext::create()->setSerializeNull(true));
+            $deserialized = $this->deserialize($this->getContent('blog_post_unauthored'), get_class($post));
```
goetas: typo. `DeserializationContext::setSerializeNull` has no effect when doing deserialization

tests/Serializer/BaseSerializationTest.php
```diff
@@ -726,14 +746,14 @@ public function testDeserializingNull()
...
 
             self::assertEquals('2011-07-30T00:00:00+00:00', $this->getField($deserialized, 'createdAt')->format(\DateTime::ATOM));
             self::assertAttributeEquals('This is a nice title.', 'title', $deserialized);
             self::assertAttributeSame(false, 'published', $deserialized);
             self::assertAttributeSame(false, 'reviewed', $deserialized);
             self::assertAttributeEquals(new ArrayCollection(), 'comments', $deserialized);
-            self::assertEquals(null, $this->getField($deserialized, 'author'));
+            self::assertEquals($author, $this->getField($deserialized, 'author'));
```
goetas: This is the BC break. Not sure even if the current behavior makes sense

src/GraphNavigator/DeserializationGraphNavigator.php
```diff
@@ -197,7 +214,10 @@ public function accept($data, ?array $type = null)
                     $this->context->pushPropertyMetadata($propertyMetadata);
                     try {
                         $v = $this->visitor->visitProperty($propertyMetadata, $data);
-                        $this->accessor->setValue($object, $v, $propertyMetadata, $this->context);
+
+                        if (null !== $v || true === $this->shouldDeserializeNull) {
```
kunicmarko20: so here is where you prevent it instead of doing in the visitor

tests/Serializer/BaseSerializationTest.php
kunicmarko20: how is this even happening if deserialize null is false

src/GraphNavigator/DeserializationGraphNavigator.php
```diff
@@ -74,6 +76,11 @@
      */
     private $accessor;
 
+    /**
+     * @var bool
+     */
+    private $shouldDeserializeNull;
```
kunicmarko20: is the property even needed? you can just do `$this->context->shouldDeserializeNull()`

tests/Serializer/BaseSerializationTest.php
```diff
@@ -726,14 +746,15 @@ public function testDeserializingNull()
         self::assertEquals($this->getContent('blog_post_unauthored'), $this->serialize($post, SerializationContext::create()->setSerializeNull(true)));
 
         if ($this->hasDeserializer()) {
-            $deserialized = $this->deserialize($this->getContent('blog_post_unauthored'), get_class($post), DeserializationContext::create());
+            $ctx =  DeserializationContext::create()
+            $deserialized = $this->deserialize($this->getContent('blog_post_unauthored'), get_class($post), $ctx);
```
kunicmarko20: this change looks like it isn't needed

src/GraphNavigator/DeserializationGraphNavigator.php
kunicmarko20: This is what is breaking the tests and now that I think about it, it is not needed at all. Without it, tests pass.

kunicmarko20:
The only problem I see with this that it will make all fields in the request deserialize null, not sure if that is a good thing.

goetas:
@kunicmarko20 I've found some time to work on this.
Te current approach allows to set the property type as nullable (like php 7.4 typed props).

Something as:

```php

class BlogPost
{
    /**
     * @Type("?string")
     */
    private $id = 'what_a_nice_id';

    /**
     * @Type("?DateTime")
     */
    private $createdAt;

    /**
     * @Type("?array<JMS\Serializer\Tests\Fixtures\Comment>")
     */
    private $comments2;

    /**
     * @Type("array<string,?string>")
     */
    private $metadata;
}
```

When `?` is used and `setDeserializeNull` is specified, then `null` properties are deserialized as null instead of having the default object value.

src/GraphNavigator/DeserializationGraphNavigator.php
goetas: this is a micro performance optimization that allows to avoid calling `shouldDeserializeNull()` too many times

kunicmarko20:
Well, this looks amazing. Now you can choose to deserialize null for everything or per prop?

goetas:
If `$context->shouldDeserializeNull()` is not called, `?` is just ignored. 
When `$context->shouldDeserializeNull()` is  called then the props having `?` in the type declaration will be null if the JSON contains null.

kunicmarko20:
ah, so you need to enable it and then select which properties this works on?

goetas:
> ah, so you need to enable it and then select which properties this works on?

yes. main reason backward compatibility.

Thinking it twice, having `?` could be enough, so `shouldDeserializeNull()`  could be not needed....

just trying to understand if there might have other implications...

kunicmarko20:
Ye, it looks like `shouldDeserializeNull` is not needed anymore since you need to add `?` to your configuration and that should be BC?

goetas:
> Ye, it looks like shouldDeserializeNull is not needed anymore since you need to add ? to your configuration and that should be BC?

No, the `?` is not a BC break. This solution is good enough IMO.
I still need to "let some time pass" to see if there is something I'might have missed, since this was the result of some "late night" coding :smile:

kunicmarko20:
What could be an issue or a problem with merging this? Also the target doesn't have to be master anymore

goetas:
> Also the target doesn't have to be master anymore

What do you mean?

kunicmarko20:
Well, it is not a BC break, so no need to merge it on master? (assuming master is next major and not minor?)

goetas:
The master branch is targeting 3.1 and not 4.x.

There are not yet any plans for a new major, so I'm keeping one branch less to manage

kunicmarko20:
My bad then, then we just wait. :smile:

pscheit-lillydoo:
I would guess some time is passed now :)

scaytrase:
Any plans to release this?

PR #1019, opened 2018-12-09
allow empty strings and numbers as metadata type parameters

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/JMSSerializerBundle/issues/706
| License       | MIT

@Majkl578 do you mind having a look at this?

### DIFF
--- a/src/Type/grammar.pp
+++ b/src/Type/grammar.pp
@@ -2,15 +2,17 @@
 
 %token parenthesis_ <
 %token _parenthesis >
+%token empty_string ""|''
+%token number        (\+|\-)?(0|[1-9]\d*)(\.\d+)?
 %token comma        ,
 %token name         (?:[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*\\)*[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*
 
 %token quote_                      "            -> quoted_string
-%token quoted_string:quoted_string (?:[^"]|"")+
+%token quoted_string:quoted_string [^"]+
 %token quoted_string:_quote        "            -> default
 
 %token apostrophe_                           '            -> apostrophed_string
-%token apostrophed_string:apostrophed_string (?:[^']|'')+
+%token apostrophed_string:apostrophed_string [^']+
 %token apostrophed_string:_apostrophe        '            -> default
 
 type:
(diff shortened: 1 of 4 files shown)


tests/Serializer/Type/ParserTest.php
```diff
@@ -72,14 +92,6 @@ public function validTypesProvider(): iterable
             'Foo<"asdf asdf">',
             $type('Foo', ['asdf asdf']),
         ];
-        yield [
```
goetas: removed this as it will not allow `''` (empty strings syntax) and was not available in the 1.x parser

goetas: From what I can guess, what you were trying to do here was handled by YAML

src/Type/grammar.pp
```diff
@@ -2,22 +2,28 @@
 
 %token parenthesis_ <
 %token _parenthesis >
+%token empty_string ""|''
+%token number        (\+|\-)?(0|[1-9]\d*)(\.\d+)?
```
Majkl578: This won't allow floats in the following form: `.123`
Should be written as something similar:
```
[+-]?(?:(?:0|[1-9]\d*(?:\.0*[1-9]\d*)?)|\.0*\d*[1-9])
```
https://regex101.com/r/Nadose/1

Also this doesn't handle exponential floats.

src/Type/grammar.pp
```diff
@@ -2,22 +2,28 @@
 
 %token parenthesis_ <
 %token _parenthesis >
+%token empty_string ""|''
```
Majkl578: This looks weird, I think this should be handled as part of strings themselves, meaning that they will no longer be a single token but a node instead.

goetas: That was the only way I found to make it work with empty strings.
Better solutions are welcome

src/Type/grammar.pp
goetas: That's true. My goal was to solve the reported bug. A more complete implementation is welcome

src/Type/grammar.pp
Majkl578: If you make the inner token optional in simple_type (`<quoted_string>?` and `<apostrophed_string>?`), it will work.

Majkl578: Of course, you made the token optional (nullable), so the visitor needs to be updated accordingly. :)

Issue #636, opened 2016-08-29
JMS Serializer cuts trailing zeros


crashev:
Hello,
jms serializer cuts trailing zeros when serializing an object although there is float or double type configured on certain field. So 100.00 becomes 100 and is treated for example by elastic search as long not as float/double. 

It should be fixed and allow jms serializer to output values like 100.00 in places where it is specified as such type.

goetas:
You can easily create a custom handler for floats and decide the policy for
fliats

On 29 Aug 2016 14:37, "Pawel Furtak" <notifications@github.com> wrote:

Hello,
jms serializer cuts trailing zeros when serializing an object although
there is float or double type configured on certain field. So 100.00
becomes 100 and is treated for example by elastic search as long not as
float/double.

It should be fixed and allow jms serializer to output values like 100.00 in
places where it is specified as such type.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/schmittjoh/serializer/issues/636>, or mute the thread

scaytrase:
I think there should be some built-in configuration for float formatting, like one we have for `\DateTime` formats

goetas:
> I think there should be some built-in configuration for float formatting, like one we have for \DateTime formats

I agree. PR are welcome

goetas:
setting `JSON_PRESERVE_ZERO_FRACTION` in the json visitor options will solve the problem.

If you are using the symfony bundle, see http://jmsyst.com/bundles/JMSSerializerBundle/master/configuration#extension-reference  (`visitors/json/optins` section)

Sevavietl:
How can I preserve zeros in XML serialization? I am using [webservices/xsd2php](https://github.com/goetas-webservices/xsd2php). Tried to create a dedicated handler, but, though I register it, it doesn't work.

Handler looks like this:

```
class FloatHandler implements SubscribingHandlerInterface
{
    public static function getSubscribingMethods()
    {
        return [
            [
                'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
                'format' => 'xml',
                'type' => 'float',
                'method' => 'serializeFloatPreserveZeros'
            ]
        ];
    }

    public function serializeFloatPreserveZeros(XmlSerializationVisitor $visitor, $data, array $type, Context $context)
    {
        var_dump($data);

        return $data;
    }
}
```

And I try to register it like this:

```
[...]

PR #1032, opened 2019-01-16
Serialize float with precision, dec_point and thousands separator

anuri73:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #636
| License       | MIT

It's not best solution, but it's work

### DIFF
--- a/src/JMS/Serializer/JsonSerializationVisitor.php
+++ b/src/JMS/Serializer/JsonSerializationVisitor.php
@@ -7,6 +7,10 @@ use JMS\Serializer\Metadata\ClassMetadata;
 use JMS\Serializer\Metadata\PropertyMetadata;
 use JMS\Serializer\Naming\AdvancedNamingStrategyInterface;
 
+/**
+ * Class JsonSerializationVisitor
+ * @package JMS\Serializer
+ */
 class JsonSerializationVisitor extends GenericSerializationVisitor
 {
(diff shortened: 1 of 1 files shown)


goetas:
Well, a json float does not allow "thousands separator" and the "dec_point" must be ".".

What you are trying to do here is to convert to string a float using some localized representation.

anuri73:
Sorry, didn't check well. It was made in haste. Give another chance for this issue

anuri73:
Can you please help,  where I can store properties for number_format. Is it ok, if I declare the numberFormatDecimals, numberFormatDecPoint and numbgerFormatThousandsSep properties in the JsonSerializationVisitor itself?

goetas:
if you need to make sure that `float 2.0` is always converted to `2.0` you can use the already available `JSON_PRESERVE_ZERO_FRACTION` option for the json visitor.

If you want to use numbgerFormatThousandsSep and numberFormatDecPoint and other stuff, then you are not dealing with a float but with a string. to do so you can use a custom type.

Something as 

```php

 /** @Type("MyFormattedFloat<'', '.', ' '>") */
 public $val = 12.3145
```

and adding a custom type handler for the `MyFormattedFloat` type

PR #1042, opened 2019-02-21
Non-locale aware encoding of doubles, closes #1041

### LINKED ISSUE #1041: Values of type "double" should not use locale-specific encoding
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

1. Use locale with decimal separator other than dot, e.g. `setlocale(LC_ALL, 'ru_RU.UTF-8')`
2. Try to serialize value of type double with fractional part, e.g. 1.234

## Expected Result

* Value should be encoded in using fixed format: no thousands separator, dot as fractional separator, e.g. "1.234"

## Actual Result

* Value encoded in locale-aware way, with locale-specific settings, e.g. "1,234", which will lead to parsing errors on produced serialized file.

This behaviour caused by converting double values to string type by using (string) operator:
```
    public function visitDouble($data, array $type, Context $context)
    {
        ...
            return $this->document->createTextNode((string) $data);
        ...
```
Easiest way to convert double to machine-readable string is using json_encode: `json_encode((double)$data)`

Grundik:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1041 
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,11 +13,10 @@
         {
             "name": "Asmir Mustafic",
             "email": "goetas@gmail.com"
-        }        
+        }
     ],
     "require": {
         "php": "^7.2",
-        "ext-json": "*",
         "jms/metadata": "^2.0",
         "doctrine/annotations": "^1.0",
         "doctrine/instantiator": "^1.0.3",
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
     ],
     "require": {
         "php": "^7.2",
+        "ext-json": "*",
         "jms/metadata": "^2.0",
         "doctrine/annotations": "^1.0",
         "doctrine/instantiator": "^1.0.3",
--- a/src/XmlSerializationVisitor.php
+++ b/src/XmlSerializationVisitor.php
@@ -176,11 +176,7 @@ final class XmlSerializationVisitor extends AbstractVisitor implements Serializa
      */
     public function visitDouble(float $data, array $type)
     {
-        if (floor($data) === $data) {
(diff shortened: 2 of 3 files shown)


goetas:
not sure if JSON encode is the best tool for the job here. 

has anybody a better idea?

@Grundik can you crate a test case for this?

Grundik:
I've added tests. Strictly speaking, existing tests already failing on systems with inconvenient system locale. Its only a matter of default locale.

json_encode exports float value as string in machine-readable form without losing precision. Its the simplest way of accomplishment of such task. Another way could be switching to C locale, but that could cause severe performance degradation.

goetas:
What about this https://3v4l.org/WgoZh ? (does not require json)

Grundik:
Thats better, and does not requires integer hack. I've made changes to PR.

But: var_exports works good on floats only on PHP >= 7.0.2, so backporting this bugfix to older versions should be done via json_encode, if that would be done.

goetas:
This package requires php ^7.2 so it should not be a probelm

Grundik:
We encountered this bug in 5.6, and we have so stick to this version for now, so I have backported this fix to 1.13.0 (https://github.com/Grundik/serializer/commits/bugfix-1041-1.13.0 if you're interested).

goetas:
Thanks!

Issue #1048, opened 2019-02-25
Using @Until and @Since on class level


tjveldhuizen:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | yes/no

The `@VirtualProperty` annotation can be used on classes and methods. The `@Until` and `@Since` annotations can be used on methods and properties. As such, it is not possible to use the \@Since annotation in the options argument of the `@VirtualProperty`, on class level.

In most cases I prefer not to create a method in the class, just to be used in the API. However, such virtual property is limited to specific versions. Therefore, my questions are: Am I right about not being abled to use those annotations in this way? If yes, would it be possible to make `@Until` and `@Since` available on class level, within the options-parameter of `@VirtualProperty`? I guess it might be relatively easy, since `@Groups` is available and seems to be working in the same way.

Example:
```
/**
 * Class Team
 *
 * @package Nevobo\BeachBundle\Entity
 *
 * @ORM\Entity(repositoryClass="Nevobo\BeachBundle\Entity\Repository\TeamRepository")
 * @Serializer\ExclusionPolicy("all")
 * @Serializer\VirtualProperty(
 *      name="players",
 *      exp="object.getTeam().getPlayers()",
 *      options={
 *          @Serializer\Until(0),
 *          @Serializer\Groups({
 *              "beach_team_results",
 *          })
 *      }
 * )
 * @ORM\Table(name="beach_participant")
 */
class Participant {
...
}
```

goetas:
Hmm, it should work out of the box. Can it be that you have a field named players in your class that overrides the virtual property?

tjveldhuizen:
No, there is no players property, nor a players() method (or even getPlayers()). The error message is as following.

> AnnotationException in AnnotationException.php line 54: [Semantical Error] Annotation @Serializer\Until is not allowed to be declared on class Nevobo\BeachBundle\Entity\Participant. You may only use this annotation on these code elements: PROPERTY, METHOD.

(I've made a small correction in the class naming of my simplified example, by the way, maing clear the VirtualProperty doesn't refer to the 'current' class)

goetas:
Well, then if is so, a PR is welcome.

goetas:
Fixed via https://github.com/schmittjoh/serializer/pull/1049

PR #1049, opened 2019-02-26
Allow @Since and @Until within @VirtualProperty on class level

tjveldhuizen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1048
| License       | MIT

As discussed in #1048, it would be nice if the `@Since` and `@Until` annotations could be used within the property parameter of `@VirtualProperty` on class level. It turns out, the serializer does work automatically as soon as the annotation is allowed on class level.

This is my first real experience with unit testing (however I absolutely understand the advantages, my main project is too large already to find a good starting point), so please give feedback if I've done stupid things ;) (Also, first PR, btw)

### DIFF
--- a/src/Annotation/Since.php
+++ b/src/Annotation/Since.php
@@ -6,7 +6,7 @@ namespace JMS\Serializer\Annotation;
 
 /**
  * @Annotation
- * @Target({"ALL"})
+ * @Target({"PROPERTY", "METHOD", "ANNOTATION"})
  */
 final class Since extends Version
 {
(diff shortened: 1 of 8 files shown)


goetas:
The annotation target should not be ALL but ANNOTATION. See https://www.doctrine-project.org/projects/doctrine-annotations/en/1.6/custom.html

Since and Until are still not allowed at class level but allowed inside the virtual property annotation

goetas:
Beside that, very well done :)

goetas:
Thanks a lot!

PR #1061, opened 2019-03-14
use Twig 2.7 namespaces

IonBazan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

Twig 2.7 moved to namespaced classes instead of old pre-PHP5.3 ones.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     },
     "require-dev": {
         "ext-pdo_sqlite": "*",
-        "twig/twig": "^1.12|^2.0",
+        "twig/twig": "^2.7",
         "doctrine/orm": "~2.1",
         "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "doctrine/phpcr-odm": "^1.3|^2.0",
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     },
     "require-dev": {
         "ext-pdo_sqlite": "*",
-        "twig/twig": "^2.7",
+        "twig/twig": "~1.34|~2.4",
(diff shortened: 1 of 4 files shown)


IonBazan:
I think there is a problem with Travis-CI config - the test is shown as passed, but it contains errors: https://travis-ci.org/schmittjoh/serializer/jobs/506178733. Although, it's not caused by this PR.

goetas:
Hi, Thanks for the PR.

Currently I think that is not a good idea to drop Twig 1.x and <2.7 support. C

Can this changes be done in a backward compatible way, preferring the new class names if they are available, and falling back on the old if they are not available?

enumag:
This doesn't necessarily break compatibility with all versions <2.7. Namespaced aliases existed several versions before that, 2.7 only added the deprecation of non-namespaced types.

Btw. not supporting Twig 2.7 is quite a problem now because roave/security-advisories no longer allows Twig 2.0-2.6 because of this [vulnerability](https://symfony.com/blog/twig-sandbox-information-disclosure).

mbabker:
> Can this changes be done in a backward compatible way, preferring the new class names if they are available, and falling back on the old if they are not available?

If the version constraint is set to `~1.34|~2.4` then you're OK, those were the two releases where the namespaced aliases were introduced.

goetas:
Having `~1.34|~2.4` is ok for me

goetas:
Thanks!

PR #1066, opened 2019-03-22
Updated suggestion for `JsonSerializationVisitor::addData` replacement

theoboldt:
When using `:visitProperty(new StaticPropertyMetadata('', 'name', 'value), null)` the 
 `\JMS\Serializer\GraphNavigator\SerializationGraphNavigator::accept()` method is called with `$data = null`, therefore `$type` will be set to `null`. Therefore, value configured in `StaticPropertyMetadata` will not be put in place (as long as it's not `null`). 

In order to make  `\JMS\Serializer\GraphNavigator\SerializationGraphNavigator::accept()` extract the correct type (via `gettype()`), the value needs to be set into `visitProperty()`. 

In conclusion, you should use `:visitProperty(new StaticPropertyMetadata('', 'name', null), 'value')`

| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | no
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no
| License       | MIT

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -17,7 +17,7 @@ in case you have heavily used internal-api here are the most important the chang
 - Removed the abstract classes `GenericSerializationVisito`r and `GenericDeserializationVisitor`.
 - Removed deprecated method `VisitorInterface::getNavigator`, use `Context::getNavigator` instead
 - Removed deprecated method `JsonSerializationVisitor::addData`, 
-  use `:visitProperty(new StaticPropertyMetadata('', 'name', 'value'), null)` instead
+  use `:visitProperty(new StaticPropertyMetadata('', 'name', null), 'value')` instead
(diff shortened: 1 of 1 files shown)


goetas:
@kunicmarko20 As I remember, you made a PR about this months ago.. any comment?

kunicmarko20:
Yes, and this is exactly the change I am using, forgot to make a PR. @theoboldt there is a code coment explain this also, can you find that one and update it too.

kunicmarko20:
https://github.com/schmittjoh/serializer/blob/d12427b868a0f9884dbe2c0285cdaa1fa834becb/src/JsonSerializationVisitor.php#L169

goetas:
Thanks!

Issue #828, opened 2017-10-16
Provide a way to specify libxml options


tehplague:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

Currently JMS Serializer has some issues when handling large XML datasets. However, SimpleXML or rather libxml2, to be precise, allows to specify parameters to `simplexml_load_string` and related methods that influence libxml's handling of incoming XML.

Among the available ones `LIBXML_COMPACT` (trying harder to allocate memory more efficiently) and `LIBXML_PARSEHUGE` (bypassing some hardcoded limitations inside libxml) are interesting when deserializing large XML bodies. As the invocation of `simplexml_load_string` is hardcoded in the `XmlDeserializationVisitor` that class needs to be overridden if one wants to change its parameters.

Thus, I propose to provide a configuration interface that allows to control the configuration of libxml for use in JMS Serializer, i.e. in `\JMS\Serializer\XmlDeserializationVisitor::prepare`  

## Steps required to reproduce the problem

1. Deserialize a large XML body with JMS Serializer

## Expected Result

* Deserialized object structure

## Actual Result

* Exception `parser error : internal error: Huge input lookup` which goes away if one modifies said invocation of `simplexml_load_string` into:

```php
        $doc = simplexml_load_string($data, 'SimpleXMLElement', LIBXML_COMPACT | LIBXML_PARSEHUGE);
```

goetas:
Make sense. something similar was implemented for the json serializer ([here](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/JsonSerializationVisitor.php#L27))

goetas:
PR is welcome

PR #1068, opened 2019-04-01
add options property to XmlDeserializationVisitorFactory and XmlDeserializationVisitor, propagate defined value from factory to simplexml_load_string call

kopeckyales:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #828
| License       | MIT

### DIFF
--- a/src/Visitor/Factory/XmlDeserializationVisitorFactory.php
+++ b/src/Visitor/Factory/XmlDeserializationVisitorFactory.php
@@ -22,9 +22,14 @@ final class XmlDeserializationVisitorFactory implements DeserializationVisitorFa
      */
     private $doctypeWhitelist = [];
 
+    /**
+     * @var int
+     */
+    private $options = 0;
+
     public function getVisitor(): DeserializationVisitorInterface
     {
-        return new XmlDeserializationVisitor($this->disableExternalEntities, $this->doctypeWhitelist);
+        return new XmlDeserializationVisitor($this->disableExternalEntities, $this->doctypeWhitelist, $this->options);
     }
 
     public function enableExternalEntities(bool $enable = true): self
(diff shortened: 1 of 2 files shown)


goetas:
Hi!
This looks a pretty nice thing to have! Thanks!

Do you mind adding some tests for it?

kopeckyales:
Hi goetas, thank you for quick response.
For me, only option how to test this is something similar to this: [php_mock_global_functions_for_unit_tests_with_phpunit](http://marcelog.github.io/articles/php_mock_global_functions_for_unit_tests_with_phpunit.html)
But it is like 50/50 because this approach seems strange for me.

I was thinking about "mock" the simplexml_load_string function and save given options parameter to the global value and then in test check if options parameter was propagated correctly.
This is IMHO bulletproof test, but as I said, I do not really like the approach.

If you are satisfied with mentioned solution, I can implement it.

goetas:
Instead of mocking the behavior of simplexml_load_string, you can try to pass some option to the visitor, then call 'prepare', and check if the return of it followed the $options rules

kopeckyales:
Of course, your solution is possible, but in my opinion it is not suitable test. What if simplexml_load_string changes in future and will return different result for whatever reason? It breaks the test even if nothing in your code was changed. And of course it do not guarantee if all options are propagated correctly, just the one you are testing. For me, this is almost comparable to have no test and it is the reason why i searched for something more bullet-proof. I only found given solution.

goetas:
I think we can keep it as it is. 
Nor happy with both solutions. :/ but would like to have this feature

PR #1072, opened 2019-04-11
doc update: registerHandler() example

cebe:
`registerHandler()` expects first argument to be int, i.e. one of the constants from GraphNavigatorInterface

| Q             | A
| ------------- | ---
| Doc updated   | yes
| License       | MIT


goetas:
thanks

PR #1075, opened 2019-04-18
Fix Travis-CI scripts always passing

IonBazan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| License       | MIT

Currently the tests are always marked as passed on Travis-CI due to lack of `set -ex` on the beginning of the tests script.
This is probably why we missed the failing tests introduced in 8ec6beadfa595957ebf907a1f5aa512216dc385a with Doctrine ORM < 2.5 when Embeddable has been introduced. 

- [x] Skip testing Embeddable entities on Doctrine ORM <2.5
- [x] Bump PHPUnit minimum version to 7.5 which introduced `assertIsCallable()`
- [x] Skip loading `BlogPostSeo` and `BlogPostWithEmbedded` metadata in `JMS\Serializer\Tests\Serializer\Doctrine\ObjectConstructorTest`

### DIFF
--- a/tests/Metadata/Driver/DoctrineDriverTest.php
+++ b/tests/Metadata/Driver/DoctrineDriverTest.php
@@ -25,7 +25,7 @@ class DoctrineDriverTest extends TestCase
 
     public function testMetadataForEmbedded()
     {
-        if (ORMVersion::compare('2.5')) {
+        if (ORMVersion::compare('2.5') >= 0) {
             $this->markTestSkipped('Not using Doctrine ORM >= 2.5 with Embedded entities');
         }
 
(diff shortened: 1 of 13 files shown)


tests/Metadata/Driver/DoctrineDriverTest.php
```diff
@@ -24,6 +25,10 @@ public function getMetadata()
 
     public function testMetadataForEmbedded()
     {
+        if (ORMVersion::compare('2.5')) {
```
goetas: I guess here should be `ORMVersion::compare('2.5')>=0`

> 	Returns -1 if older, 0 if it is the same, 1 if version passed as argument is newer.

Returns -1 if older, 0 if it is the same, 1 if version passed as argument is newer.
--

goetas:
```suggestion
        if (ORMVersion::compare('2.5')>=0) {
```

goetas:
If is too difficult, why not just dropping doctrine <2.5 support?

IonBazan:
@goetas I fixed it by moving the `Article`, `BlogPost` and `Comment` entities to a separate directory and load only the specific metadata in `ObjectConstructorTest `.

goetas:
Great!

goetas:
Thanks!

PR #1078, opened 2019-04-30
Allow loading different YAML extensions

### LINKED ISSUE #1077: Support *.yaml extension
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

`*.yml` extension is currently hardcoded. https://github.com/schmittjoh/serializer/blob/master/src/Metadata/Driver/YamlDriver.php#L290

It would be nice to support both, since `*.yaml` is considered canonical https://yaml.org/faq.html

scaytrase:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | yes (TBD)
| Tests pass?   | yes
| Fixed tickets | #1077 
| License       | MIT

Collaspse `AbstractFileDriver` into `YamlDriver` to allow multiple extensions (`yaml`, `yml`) to processed. Updated some tests

Fixes #1077 

cc. @goetas

### DIFF
--- a/src/Metadata/Driver/YamlDriver.php
+++ b/src/Metadata/Driver/YamlDriver.php
@@ -16,8 +16,11 @@ use JMS\Serializer\Type\Parser;
 use JMS\Serializer\Type\ParserInterface;
 use Metadata\ClassMetadata as BaseClassMetadata;
 use Metadata\Driver\AbstractFileDriver;
+use Metadata\Driver\AdvancedFileLocatorInterface;
 use Metadata\Driver\FileLocatorInterface;
 use Metadata\MethodMetadata;
+use ReflectionClass;
+use RuntimeException;
 use Symfony\Component\Yaml\Yaml;
 
 class YamlDriver extends AbstractFileDriver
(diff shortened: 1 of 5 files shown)


tests/Metadata/Driver/YamlDriverTest.php
```diff
@@ -72,12 +74,45 @@ public function testBlogPostAccessor()
...
+
+    private function getDriverForSubDir($subDir = null, bool $addUnderscoreDir = true): YamlDriver
+    {
+        $dirs = [
             'JMS\Serializer\Tests\Fixtures' => __DIR__ . '/yml' . ($subDir ? '/' . $subDir : ''),
-            '' => __DIR__ . '/yml/_' . ($subDir ? '/' . $subDir : ''),
-        ]), new IdenticalPropertyNamingStrategy(), null, $this->getExpressionEvaluator());
+        ];
+
+        if ($addUnderscoreDir) {
```
scaytrase: This one broke `getAllClassNames` testm since this is invalid dir for most cases, but only `getAllClassNames` iterates over all directories and fails. So I made this optional

tests/Metadata/Driver/yml/multiple_types/Person.yaml
```diff
@@ -0,0 +1,4 @@
+JMS\Serializer\Tests\Fixtures\Person:
```
scaytrase: Just copy-pasted some files in order to check loadability

src/Metadata/Driver/YamlDriver.php
```diff
@@ -15,12 +15,16 @@
...
+use Metadata\Driver\AdvancedFileLocatorInterface;
+use Metadata\Driver\FileLocator;
 use Metadata\Driver\FileLocatorInterface;
 use Metadata\MethodMetadata;
+use ReflectionClass;
+use RuntimeException;
 use Symfony\Component\Yaml\Yaml;
 
-class YamlDriver extends AbstractFileDriver
+class YamlDriver implements AdvancedDriverInterface
```
goetas: Does it make sense just for compatibility to still extend `AbstractFileDriver`?

src/Metadata/Driver/YamlDriver.php
```diff
@@ -32,27 +36,74 @@ class YamlDriver extends AbstractFileDriver
      * @var PropertyNamingStrategyInterface
      */
     private $namingStrategy;
+    /**
+     * @var FileLocatorInterface|FileLocator
```
goetas: Do we need the `FileLocator` typehint here?

src/Metadata/Driver/YamlDriver.php
scaytrase: All of it public and protected contracts were kept. Only `instance of AbstractFileDriver` should break. if it's not enougth we can keep extension with everything overrided

src/Metadata/Driver/YamlDriver.php
scaytrase: Not really, just copy-pasted from `metadata`. Will fix it

scaytrase:
@goetas fixed both comments

goetas:
Thanks a lot!

PR #1080, opened 2019-05-03
Move type check when deserializing into the graph navigator

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

### DIFF
--- a/src/GraphNavigator/DeserializationGraphNavigator.php
+++ b/src/GraphNavigator/DeserializationGraphNavigator.php
@@ -194,6 +194,10 @@ final class DeserializationGraphNavigator extends GraphNavigator implements Grap
                         continue;
                     }
 
+                    if (!$propertyMetadata->type) {
+                        throw new RuntimeException(sprintf('You must define a type for %s::$%s.', $propertyMetadata->class, $propertyMetadata->name));
+                    }
+
                     $this->context->pushPropertyMetadata($propertyMetadata);
                     try {
                         $v = $this->visitor->visitProperty($propertyMetadata, $data);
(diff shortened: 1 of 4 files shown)


src/GraphNavigator/DeserializationGraphNavigator.php
```diff
@@ -194,6 +194,10 @@ public function accept($data, ?array $type = null)
                         continue;
                     }
 
+                    if (!$propertyMetadata->type) {
+                        throw new RuntimeException(sprintf('You must define a type for %s::$%s.', $propertyMetadata->class, $propertyMetadata->name));
```
surikman: Could you describe please, why is this check necessery?

My use-case:
I have an `AbstractCommand` => class in library outside of project and then I have own Command (e.g.: `CreateUser`) which extends `AbstractCommand` and my command has some properties (e.g.: username, email...)
I need to deserialize request such as {"username": "JohnDoe", "email":"john.doe@example.com"} into my `CreateUser` command (BUT `AbstractCommand` has special property `metadata`.

I don't need to use this property in my project, but Library works with this property.
When I updated from 3.0.1 to 3.1.0 I got an error message:
`You must define a type for MyExternalLibraryAKACommandBus\AbstractCommand::$metadata`

Is there any chance to skip this throwing exception instead of doing changes in this library?

surikman: I opened an issue: #1102

PR #1081, opened 2019-05-03
Nullable type definitions using php 7.4 typed properties style

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no 
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

- [ ] feature and tests
- [ ] documentation
- [ ] usecase


simPod:
This should also handle the case when

1) I serialize object with uninitialized nullable property
2) Deserialize
3) The property is uninitialized

Personally, I'd somehow get rid of `$context->setSerializeNull(true);` setting and made it default.

Unintialized properties would then be not part of serialized content while initialized with `null` would.

Tobion:
If this is meant to change the behavior of #1252 and https://github.com/schmittjoh/serializer/pull/1081#issuecomment-704363293 then I don't think it's the right direction. So a property that is not serialized should not initialize the property after deserialization to null. This is because we don't know if the initial property was missing from the serialization because it was unitialized or because of serialize_null = false.

So I think what we need is a different option like `serialize_typed_null` which defaults to true. Then typed nullable properties are serialized by default which solves #1252. And people can still configure the nullable behavior of non-typed properties independently.

PR #1082, opened 2019-05-07
Remove strict return type

ZaiatsDmytro:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

According to JMS\Serializer\VisitorInterface method getResult() allows to return a mixed type of result but in JMS\Serializer\SerializerInterface method serialize() has a strict return type 'string'. I suppose this is a small bug which does not allow to use visitors with different return types.

### DIFF
--- a/src/Serializer.php
+++ b/src/Serializer.php
@@ -148,7 +148,7 @@ final class Serializer implements SerializerInterface, ArrayTransformerInterface
     /**
      * {@InheritDoc}
      */
-    public function serialize($data, string $format, ?SerializationContext $context = null, ?string $type = null): string
+    public function serialize($data, string $format, ?SerializationContext $context = null, ?string $type = null)
     {
         if (null === $context) {
             $context = $this->serializationContextFactory->createSerializationContext();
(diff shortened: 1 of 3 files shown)


goetas:
I think that the "bug" here is the phpdoc signature, not the interface signature. 

Is there any reason why mixed should be preferred over string?

goetas:
The reason why visitor interface has been kept mixed is to have the same base interface for serializers and deserializers.

Probably the best thing is to throw awa the generic visitor interface in favor of specific interfaces. But could  be a big bc break.

ZaiatsDmytro:
> An array can not be stored without an additional step, thus is not serialization.
I completely agree

> If you need an array as output, have you considered ArrayTransformerInterface?
ArrayTransformerInterface:toArray() use 'json'-visitor this is a bit confused me but as I see now I can use it for my purposes.

Thank you for your feedback

ZaiatsDmytro:
[#1082 (comment)](https://github.com/schmittjoh/serializer/pull/1082#issuecomment-490535908)

Issue #260, opened 2014-04-08
Support inheritance when looking for a handler in the lazy registry


vboriesazeau:
Hi,

This issue is more a request for an improvement rather than a bug.
It seems the actual code in the `LazyHandlerRegistry` simply checks if there is a handler registered for exactly a given `$typeName`.
But handlers are not found for sub-classes.
Using this configuration in Symfony:

```
myhandler:
        class: %myhandler.class%
        tags:
            -
                name: jms_serializer.handler
                type: AnAbstractClass
                direction: serialization
                format: json
```

the handler is never used if I try to serialize a concrete implementation (sub-class) of `AnAbstractClass`.
That means one must duplicate this handler configuration for each type supported, and change the 'type' key (there is no need to duplicate the handler class itself I guess, It shall be possible to use abstraction in the code). So the working configuration actually shall be:

```
myhandler_for_subclass1:
        class: %myhandler.class%
        tags:
            -
                name: jms_serializer.handler
                type: SubClass1
                direction: serialization
                format: json
myhandler_for_subclass2:
        class: %myhandler.class%
        tags:
            -
                name: jms_serializer.handler
                type: SubClass2
                direction: serialization
                format: json
```

I would suggest the registry in priority gives a handler for exactly the same type name, if any, and after that, as a fallback, goes up in the hierarchy of the type name and looks if there is a dedicated handler, and so on...
Let me know your opinion :).

BR

johndodev:
Bump,

actually you can declare only one service with a "jms_serializer.subscribing_handler", but the "bug" is still there. 

And since update the "::getSubscribingMethods" does not delete the cache, It was a nightmare to handle it. The documentation with "Datetime" is not really relevant..

goetas:
Closing this for the same reason as in https://github.com/schmittjoh/serializer/pull/1089#issuecomment-497293678

PR #1089, opened 2019-05-29
Support to use same handler for subclasses

edefimov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #260
| License       | MIT

This PR adds a feature to use the same serialization/deserialization handler for class hierarchy in `HandlerRegistry` and `LazyHandlerRegistry`. 

Example:
```php
class A {
  public $value;
}

class B extends A {}

$handler = new MyHandler();
$handlerRegistry->registerHandler(GraphNavigatorInterface::DIRECTION_SERIALIZATION, A::class, 'json', $handler);

...

// Will call serialization handler for A if handler for B is not explicitly registered
$serializer->serialize(new B, 'json');
```

### DIFF
--- a/src/Handler/HandlerRegistry.php
+++ b/src/Handler/HandlerRegistry.php
@@ -70,6 +70,29 @@ class HandlerRegistry implements HandlerRegistryInterface
      */
     public function getHandler(int $direction, string $typeName, string $format)
     {
+        $currentType = $typeName;
+        do {
+            $handler = $this->getSingleHandler($direction, $currentType, $format);
+            if (null !== $handler) {
+                return $handler;
+            }
+        } while ($currentType = get_parent_class($currentType));
+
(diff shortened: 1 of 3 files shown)


goetas:
Hi and thanks for the PR:

This is something has been already discussed in the past (and as example FOS REST have implemented it in their bunde, see [this](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/38046e93a69c3d4dbe650e5def36ff9fd35fcaaf/Serializer/JMSHandlerRegistry.php#L52)) but the decision was to not include it in the serializer.

An almost the same PR was https://github.com/schmittjoh/serializer/pull/556 and there you cam find some reasons why that was not merged.

PR #1091, opened 2019-06-03
Add psalm specific generic return type for deserialize

bdsl:
See https://psalm.dev/ and
https://medium.com/vimeo-engineering-blog/uncovering-php-bugs-with-template-a4ca46eb9aeb

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | no
| Fixed tickets | #... 
| License       | MIT

### DIFF
--- a/src/SerializerInterface.php
+++ b/src/SerializerInterface.php
@@ -22,6 +22,10 @@ interface SerializerInterface
      * Deserializes the given data to the specified type.
      *
      * @return mixed
+     *
+     * @psalm-template T
+     * @psalm-param class-string<T> $type
+     * @psalm-return T
      */
     public function deserialize(string $data, string $type, string $format, ?DeserializationContext $context = null);
 }


bdsl:
These annotations tell the [Psalm static analyser](https://psalm.dev/) that the object returned by `deserialize` will be an instance of the class named in the `$type` parameter. Psalm can use that to find errors, e.g. calling a function that doesn't exist in that class, and also to provide autocompletion suggestions.

goetas:
Im fine with this, but is there a way to test it?

bdsl:
@goetas Probably if you want to have this work reliably you'd need need to add psalm to the build pipeline for serializer, firstly to check that it doesn't error when parsings this docblock, and then also to make sure that it infers the type correctly when a caller uses this method.

For now you can see how this allows psalm to detect a spelling mistake when calling a function on the DTO at https://psalm.dev/r/1f725d788b

goetas:
thanks

bdsl:
You're welcome, thanks for the quick response. Do you have any plan for when this would be in a release? I don't know if there are any set release dates or roadmaps for this project.

goetas:
this project does not have "time scheduled" releases. Generally when I have some time I create a changelog and do the release.

bdsl:
Thanks @goetas . Let me know if you and other maintainers would like a PR to add psalm to the travis build. Psalm has a baselining feature so any issues it finds in when first set up don't have to be immediately fixed.

goetas:
Having some basic Psalm integration into the serializer would be very nice. I guess is not any easy task, but would be happy to see a PR with that

carusogabriel:
@bdsl How are we supposed to annotate our classes that we pass to the `deserialize` method from now on?

bdsl:
@carusogabriel This PR didn't change anything about how you should use deserialize. It just means that when you write something like `$dto = $serializer->deserialize($json, ExampleDto::class, 'json')` Psalm will know that $dto is an `ExampleDto` object. You can take a look at https://psalm.dev/r/1f725d788b for an illustration.

I'm not sure if that answers your question.

carusogabriel:
I'm getting some PHPStan 0.12 errors, like: 

- `Unable to resolve the template type T in call to method JMS\Serializer\SerializerInterface::deserialize()`
- `Parameter #2 $type of method JMS\Serializer\SerializerInterface::deserialize() expects class-string<mixed>, string given.`

bdsl:
@carusogabriel hmm, probably best to raise a new issue for that, either here or at https://github.com/phpstan/phpstan/issues , with a minimal reproducing example if possible.

bdsl:
@carusogabriel How did you get on? Will you make a new issue report?

carusogabriel:
When I figure out what's happening in how to reproduce, yes, I'll report an issue here 👍

bdsl:
Ok, feel free to tag me in when you do - I might have some time to look into how to fix it. Worth linking to this PR as well.

Issue #1087, opened 2019-05-20
Empty namespace


antonioturdo:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | yes

I really am not sure whether this is a bug or not. I have a third party xml that looks like this:

```
<?xml version="1.0" encoding="utf-8"?>
<A xmlns="http://namespace.org/ns">
   <B xmlns="">text</B>
</A>
```

An empty namespace should means no default namespace has to be considered for B.

Current implementation:

```
$namespaces = $data->getDocNamespaces();

if (isset($namespaces[''])) {
    $prefix = uniqid('ns-');
    $data->registerXPathNamespace($prefix, $namespaces['']);
    $nodes = $data->xpath('./' . $prefix . ':' . $name);
} else {
    $nodes = $data->xpath('./' . $name);
}
```
The document has default namespace, but since "B" element is not in the default namespace, I think xpath correctly doesn't found it.

## Steps required to reproduce the problem

1. Deserialize the xml shown above

## Expected Result

* An object "A" with the "B" property set.

## Actual Result

* An object "A" with null "B" property.

goetas:
```xml
<?xml version="1.0" encoding="utf-8"?>
<A xmlns="http://namespace.org/ns">
   <B xmlns="">text</B>
</A>
```

OR 

```xml
<?xml version="1.0" encoding="utf-8"?>
<x:A xmlns="http://namespace.org/ns" y:xmlns="">
   <y:B>text</B>
</x:A>
```

OR
```xml
<?xml version="1.0" encoding="utf-8"?>
<x:A x:xmlns="http://namespace.org/ns">
   <B>text</B>
</x:A>
```

Are all equivalent.

The result depends on your XML mapping info in the entity. Can you share those info please?

antonioturdo:
```
use JMS\Serializer\Annotation as JMS;

/**
 * @JMS\XmlNamespace(uri="http://namespace.org/ns")
 * @JMS\XmlNamespace(uri="", prefix="y")
 * @JMS\XmlRoot(name="A")
 */
class A {
    
	/**
	 * @JMS\Type("MyNamespace\Entity\B")
	 * @JMS\SerializedName("B")
         * @JMS\XmlElement(namespace="")
	 */	
	protected $B;
}
```

In class B I haven't annotations.
The empty namespace annotation in B property does not have effects, because

```
if ($metadata->xmlNamespace) {
[...]

PR #1095, opened 2019-06-13
Implement "empty" XML namespace handling

discordier:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1087
| License       | MIT

This is a basic implementation of `xmlns=""` handling.

The implementation works for me (see the tests).

### DIFF
--- a/src/XmlSerializationVisitor.php
+++ b/src/XmlSerializationVisitor.php
@@ -463,6 +463,15 @@ final class XmlSerializationVisitor extends AbstractVisitor implements Serializa
 
     private function createElement(string $tagName, ?string $namespace = null): \DOMElement
     {
+        // See #1087 - element must be like: <element xmlns="" /> - https://www.w3.org/TR/REC-xml-names/#iri-use
+        // Use of an empty string in a namespace declaration turns it into an "undeclaration".
+        if ('' === $namespace) {
+            // If we have a default namespace, we need to create namespaced.
+            if ($this->parentHasNonEmptyDefaultNs()) {
+                return $this->document->createElementNS($namespace, $tagName);
+            }
+            return $this->document->createElement($tagName);
+        }
         if (null === $namespace) {
             return $this->document->createElement($tagName);
         }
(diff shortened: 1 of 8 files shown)


discordier:
Fixed code style.

src/XmlSerializationVisitor.php
```diff
@@ -491,4 +500,16 @@ private function getClassDefaultNamespace(ClassMetadata $metadata): ?string
     {
         return $metadata->xmlNamespaces[''] ?? null;
     }
+
+    private function parentHasNonEmptyDefaultNs(): bool
+    {
+        $element = $this->currentNode;
+        while ($element) {
```
goetas: Looks good so far. I'm just worried about the performance impact of this loop...
It should be tested carefully

discordier: The impact should not be too severe, as it will only be triggered if the element to be serialized has the explicit `''` (empty string) namespace and it will only loop up until it hits any `xmlns="..."` (non empty default namespace).

I am open to suggestions of how to test this performance wise.

goetas: I have tried to debug this and IMO should be enough.

```php
return (null !== ($uri = $this->currentNode->lookupNamespaceUri(null)) && ('' !== $uri));
```

Consider that `lookupNamespaceUri` is aware of the position of the dom element and will give you the default namespace for a node considering its parents.

See https://3v4l.org/ONH44

discordier: I think you are right. I was thinking that I have to traverse the tree up but we can scrap the loop.

So I will drop the loop but keep the method?

goetas: Sounds good to me. We can keep the method for readability

discordier:
Removed the loop, how shall we approach the feature request of @antonioturdo regarding applying it to collections?

goetas:
From my understanding the same approach can be used for serializing/deserializing collections.

@discordier or @antonioturdo   Let me know if you prefer to continue on it directly in this PR or on a separate one.

discordier:
As I need the behavior implemented in this PR in a project currently and considering that I have not had a look at the collection part (and I think I won't have the time in the near future), I'd favor merging it rather sooner than later. :smile_cat:

goetas:
@discordier I agree! @antonioturdo feel free to send a PR for collections support

antonioturdo:
> @discordier I agree! @antonioturdo feel free to send a PR for collections support

Ok, I'll do if I find some time.

Issue #1094, opened 2019-06-07
Support for iterable


simPod:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

Here https://github.com/schmittjoh/serializer/issues/1023 the generators and iterators support was implemented. I wonder would it be possible to support `iterable`?

Currently, I have to type against concrete type, eg.  
`* @Serializer\Type("array<T>")`  
`* @Serializer\Type("Generator<T>")`

I'd like to type as `* @Serializer\Type("iterable<T>")`. I suppose it could be deserialized eg. as an array when deserializing.

Would it make sense?

goetas:
I think here my answer is the same as for the `Iterator` interface, see in https://github.com/schmittjoh/serializer/pull/1034#discussion_r251248190

simPod:
I was thinking that since we're working with interface the implementation shouldn't matter (if end user respect the type ofc).

The issue is that currently it's not possible to have:

```php
/**
 * @Serializer\Type("Generator<T>")
 * @var iterable<T>
 */
private $collection;
```

`@var` says that it expects interface so implementation doesn't matter. But serializer requires concrete type. This breaks LSP.

goetas:
how to deserialize `@Serializer\Type("iterable<T>")`? 

Do you want to say that using `iterable<T>`, the consumer should not care if it has an array, an iterator or other, the important is that is unserialized into something "iterable" ?

simPod:
Yes exactly. I don't see any better way when working with interfaces. But consumer should not care about implementation when expecting interface. 🤔

goetas:
You convinced me. 

With this in mind, then `Iterator` can also be implemented, amending my comment in https://github.com/schmittjoh/serializer/pull/1034#discussion_r251248190

simPod:
Will look into it, thanks!

PR #1096, opened 2019-06-14
Add support for iterable and Iterator

simPod:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #1094
| License       | MIT

- [x] Add `iterable` support
- [x] Add `Iterator` support
- [x] ~Make `iterable` work with jms/meta~

### DIFF
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -378,6 +378,18 @@ Available Types:
 +----------------------------------------------------------+--------------------------------------------------+
 | T                                                        | Where T is a fully qualified class name.         |
 +----------------------------------------------------------+--------------------------------------------------+
+| iterable                                                 | Similar to array. Will always be deserialized    |
+|                                                          | into array as implementation info is lost during |
+|                                                          | serialization.                                   |
++----------------------------------------------------------+--------------------------------------------------+
+| iterable<T>                                              | Similar to array<T>. Will always be deserialized |
+|                                                          | into array as implementation info is lost during |
+|                                                          | serialization.                                   |
++----------------------------------------------------------+--------------------------------------------------+
+| iterable<K, V>                                           | Similar to array<K, V>. Will always be           |
+|                                                          | deserialized into array as implementation info   |
+|                                                          | is lost during serialization.                    |
++----------------------------------------------------------+--------------------------------------------------+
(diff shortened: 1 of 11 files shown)


simPod:
@goetas jms/meta heavily relies on ReflectionClass. However, it's not possible to create reflection class from `iterable`. Any idea how to tackle it?

composer.json
```diff
@@ -52,7 +52,10 @@
     "autoload": {
         "psr-4": {
             "JMS\\Serializer\\": "src/"
-        }
+        },
+        "files": [
+            "src/Functions/iterable_to_array.php"
```
goetas: Functions can not be auto-loaded, this means that this file will be read on each request. 
You can place this utility function in a static method inside some class, and autoloading will work

src/Functions/iterable_to_array.php
```diff
@@ -0,0 +1,20 @@
...
+use function iterator_to_array;
+
+/**
+ * Copy the iterable into an array. If the iterable is already an array, return it.
+ *
+ * @param bool $use_keys [optional] Whether to use the iterator element keys as index.
+ *
+ * @return array
+ */
+function iterable_to_array(iterable $iterable, bool $use_keys = true) : array
```
goetas: `$use_keys` looks never used in the calling code

goetas:
> @goetas jms/meta heavily relies on ReflectionClass. However, it's not possible to create reflection class from iterable. Any idea how to tackle it?

Can you explain which problems gives this?

simPod:
@goetas ooooooh 👍

src/Functions/iterable_to_array.php
simPod: Copied it from library I use so I don't have to include it whole https://github.com/bpolaszek/php-iterable-functions/blob/master/src/iterable-functions.php#L34

I removed `$use_keys`

composer.json
simPod: I tried Functions, WDYT?

goetas: im good with it

goetas:
Thanks a lot!

goetas:
Well done!

Issue #1098, opened 2019-06-20
ConditionalExpose/Exclude annotation does not work on class level


arneee:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Hi,

I have an array with several Office Objects. When serializing, I would like to exclude some if these objects, based on dynamic conditions.  The [documentation ](https://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#dynamic-exclusion-strategy) shows how to do that with **\@Exclude** and the annoation field **if**.

While it works for normal properties, the dynamic part (if) of **\@Exclude**  does not seem to be used when applied to a class. 

## Steps required to reproduce the problem

```php
/**
 * Office
 *
 * @Exclude(if="false")
 */
class Office {
   public $id = 1;
}
$array = [new Office(), new Office()];
$serializer->serialize($array , 'json');
```
## Expected Result
*  All office objects are included in the serialized result, since the exclude condition evalues to false

## Actual Result
```json
[
{},
{}
]
```
* No office objects are serialized, since the condition seems not be be evaluated. 

I've also triedit with a real world example (isGranted provided by the Symfony bundle), but the isGranted method is never called. There is also no exception when using an invalid method, for example ` @Exclude(if="noSuchMethod('view',object)")` which makes me believe that the condition is not even parsed.

arneee:
Hi,

this functionality does not seem to exist, I've created a draft PR. #1099

arneee:
Possible now with https://github.com/schmittjoh/serializer/pull/1099

PR #1099, opened 2019-06-21
Support conditional exclude for classes

arneee:
Hi,

after further checking #1098, I've noticed that this functionality just isn't there yet. I've created a first draft PR for supporting condtional excludes on class level.

The PR includes the changes in class metadata and SerializationGraphNavigator as well as the annotation driver. 

Still missing is:
* Evaluate condition in DeserializationGraphNavigator
* Other Metadata drivers
* Unit tests

While all existing tests are passing, I was not able to add the corresponding unit tests for this case. I'm also not sure if there are any unintended side effects by not completely excluding the class if there is an \@Exclude annotation like it is done at the moment, but keeping it and do the evaluation during serialization.

Thanks for your help!

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1098
| License       | MIT

### DIFF
--- a/src/Exclusion/ExpressionLanguageExclusionStrategy.php
+++ b/src/Exclusion/ExpressionLanguageExclusionStrategy.php
@@ -8,6 +8,7 @@ use JMS\Serializer\Context;
 use JMS\Serializer\Expression\CompilableExpressionEvaluatorInterface;
 use JMS\Serializer\Expression\Expression;
 use JMS\Serializer\Expression\ExpressionEvaluatorInterface;
+use JMS\Serializer\Metadata\ClassMetadata;
 use JMS\Serializer\Metadata\PropertyMetadata;
 use JMS\Serializer\SerializationContext;
 
@@ -29,6 +30,33 @@ final class ExpressionLanguageExclusionStrategy
     public function __construct(ExpressionEvaluatorInterface $expressionEvaluator)
     {
         $this->expressionEvaluator = $expressionEvaluator;
+    }
+
+
+    /**
+     * {@inheritDoc}
+     */
+    public function shouldSkipClass(ClassMetadata $class, Context $navigatorContext): bool
+    {
+        if (null === $class->excludeIf) {
+            return false;
+        }
+
+        $variables = [
+            'context' => $navigatorContext,
+            'property_metadata' => $class,
+        ];
+        if ($navigatorContext instanceof SerializationContext) {
+            $variables['object'] = $navigatorContext->getObject();
+        } else {
+            $variables['object'] = null;
+        }
+
+        if (($class->excludeIf instanceof Expression) && ($this->expressionEvaluator instanceof CompilableExpressionEvaluatorInterface)) {
+            return $this->expressionEvaluator->evaluateParsed($class->excludeIf, $variables);
+        }
+
+        return $this->expressionEvaluator->evaluate($class->excludeIf, $variables);
     }
 
     /**
--- a/src/Exclusion/ExpressionLanguageExclusionStrategy.php
+++ b/src/Exclusion/ExpressionLanguageExclusionStrategy.php
@@ -44,7 +44,7 @@ final class ExpressionLanguageExclusionStrategy
 
         $variables = [
             'context' => $navigatorContext,
-            'property_metadata' => $class,
+            'class_metadata' => $class,
         ];
         if ($navigatorContext instanceof SerializationContext) {
             $variables['object'] = $navigatorContext->getObject();
--- a/src/Metadata/ClassMetadata.php
+++ b/src/Metadata/ClassMetadata.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace JMS\Serializer\Metadata;
 
 use JMS\Serializer\Exception\InvalidMetadataException;
+use JMS\Serializer\Expression\Expression;
 use JMS\Serializer\Ordering\AlphabeticalPropertyOrderingStrategy;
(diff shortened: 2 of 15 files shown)


goetas:
Very nice addition! Thanks a lot! 
Will to a proper review in the next days.

arneee:
Hi Asmir,

Cool, thanks! 👍 

The PR definitely still needs some work, but since I only used this library yesterday for the first time, I'm not sure how to do the missing parts, especiallcy the unit tests.

Regards,

  Arne

src/Metadata/ClassMetadata.php
```diff
@@ -126,6 +126,11 @@ class ClassMetadata extends MergeableClassMetadata
      */
     public $xmlDiscriminatorNamespace;
 
+    /**
+     * @var bool
```
goetas: This should not be a boolean, probably `string|null|Expression`

src/Exclusion/ExpressionLanguageExclusionStrategy.php
```diff
@@ -56,4 +57,31 @@ public function shouldSkipProperty(PropertyMetadata $property, Context $navigato
...
+            return false;
+        }
+
+        $variables = [
+            'context' => $navigatorContext,
+            'property_metadata' => $class,
+        ];
+        if ($navigatorContext instanceof SerializationContext) {
+            $variables['object'] = $navigatorContext->getObject();
+        } else {
```
spam312sn: I think you can avoid using `else` by adding default value in `$variables` array:

```diff
        $variables = [
            'context' => $navigatorContext,
            'property_metadata' => $class,
+           'object' => null,
        ];

        if ($navigatorContext instanceof SerializationContext) {
            $variables['object'] = $navigatorContext->getObject();
-       } else {
-           $variables['object'] = null;
        }
```

arneee:
Hi @goetas ,

i was working on this again and run into a issue. While serialization works fine, there is one problem during deserialization.

```php
/**
 * @Serializer\Exclude(if="object.expired != true")
 */
class PersonAccount {

    /**
     * @Serializer\Type("string")
     */
    public $name;

    /**
     * @Serializer\Type("boolean")
     */
    public $expired;

}
```

During deserialize() we come to the point where the exclude-If expression is being evaluated. Unfortunately the object is null, since the deserialization has not happened yet and therefore the expression fails.

I see two options.

1. Serialize the object first (including all the properties, childs, etc.) and then evaluate the expression
2. Skip the exclude during deserialization 

What is your optionen on this? Thanks!

goetas:
This is a known issue, and `@Serializer\Exclude(if="object && object.expired != true")` should do the job. Do you think is enough?

arneee:
That would be the easiest solution. For my use case that would be perfectly fine, since I'm interes in serialization only anyway. For the library I don't know, should deserialization offer the same functionality like serialization? If this is ok for you, it's fine for me.

arneee:
Ok, hope I got this right and in sync now. 

I've tried to get it working with the solution you have suggested before, but I wasn't able to completely skip the relevant objects during deserialization. If an object was skipped by a class level exclude-if, the array of the parent element still contained a null item.

For now I've removed the Exclude-If on deserialization, it is only works during serialization.

src/Exclusion/ExpressionLanguageExclusionStrategy.php
```diff
@@ -31,6 +32,33 @@ public function __construct(ExpressionEvaluatorInterface $expressionEvaluator)
...
+     */
+    public function shouldSkipClass(ClassMetadata $class, Context $navigatorContext): bool
+    {
+        if (null === $class->excludeIf) {
+            return false;
+        }
+
+        $variables = [
+            'context' => $navigatorContext,
+            'property_metadata' => $class,
```
goetas, line 47: I believe that should be called class_metadata

src/Metadata/ClassMetadata.php
```diff
@@ -126,6 +126,11 @@ class ClassMetadata extends MergeableClassMetadata
      */
     public $xmlDiscriminatorNamespace;
 
+    /**
+     * @var string
+     */
+    public $excludeIf;
```
goetas, line 132: Can it be also Expression?

arneee: It should... I've taken over this part from the PropertyMetaData. I will also change it in the PropertyMetadata.

src/Exclusion/ExpressionLanguageExclusionStrategy.php
arneee: Yepp!

arneee:
> Can you please add some tests for the Metadata in xml and yaml?

Done (and now the drivers are actually working). 😃

arneee:
> Looks good! Can you please fix the code style issue (you can just run vendor/bin/phpcbf), after that I think it is ready to be merged!

It is coming from GraphNavigatorTest, which I didn't change. phpcbf does not find anything on my installation ("No fixable errors were found")

goetas:
Thanks a lot for your excellent work! very very nice! :)

arneee:
Hi Asmir,
Thanks for the merge, unfortunately I've noticed a mistake on my side. Currently the condition is inverted, means Exclude(if="false") will actually exclude it. This was due to a copy mistake from @Expose in the drivers as well as confusing test. I have simplified the test to make it more clear and fixed the drivers. Do you need a new PR for this?

PR #1107, opened 2019-07-11
Fix typos in UPGRADING.md

jdreesen:
First the document says the new way is to call
`JsonSerializationVisitor::visitProperty(new StaticPropertyMetadata('', 'name', null), 'value')`

Later it says one should call instead
`JsonSerializationVisitor::visitProperty(new StaticPropertyMetadata('', 'name', 'value'), null)`

I'm not sure which one is correct.


goetas:
Hi! Thanks for this PR.


`visitProperty(new StaticPropertyMetadata('', 'name', 'value'), null)` is the correct one. Do you mind updating the PR ? 

After that I think is good to be merged.

jdreesen:
:heavy_check_mark:

goetas:
Thank you!

PR #1108, opened 2019-07-14
Handle array format for dateHandler

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT

### DIFF
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -350,28 +350,27 @@ Available Types:
 +------------------------------------------------------------+--------------------------------------------------+
 | DateTime                                                   | PHP's DateTime object (default format*/timezone) |
 +------------------------------------------------------------+--------------------------------------------------+
-| DateTime<'formats'>                                        | PHP's DateTime object (custom format/default     |
-|                                                            | timezone). Formats can either be a string or an  |
-|                                                            | array of string.                                 |
+| DateTime<'format'>                                         | PHP's DateTime object (custom format/default     |
+|                                                            | timezone).                                       |
 +------------------------------------------------------------+--------------------------------------------------+
-| DateTime<'formats', 'zone'>                                | PHP's DateTime object (custom format/timezone)   |
+| DateTime<'format', 'zone'>                                 | PHP's DateTime object (custom format/timezone)   |
 +------------------------------------------------------------+--------------------------------------------------+
-| DateTime<'formats', 'zone', 'deserializeFormats'>          | PHP's DateTime object (custom format/timezone,   |
+| DateTime<'format', 'zone', 'deserializeFormats'>           | PHP's DateTime object (custom format/timezone,   |
 |                                                            | deserialize format). If you do not want to       |
 |                                                            | specify a specific timezone, use an empty        |
-|                                                            | string ('').                                     |
+|                                                            | string (''). DeserializeFormats can either be a  |
+|                                                            | string or an array of string.                    |
 +------------------------------------------------------------+--------------------------------------------------+
 | DateTimeImmutable                                          | PHP's DateTimeImmutable object (default format*/ |
-|                                                            | timezone). Formats can either be a string or an  |
-|                                                            | array of string.                                 |
+|                                                            | timezone).                                       |
 +------------------------------------------------------------+--------------------------------------------------+
-| DateTimeImmutable<'formats'>                               | PHP's DateTimeImmutable object (custom format/   |
+| DateTimeImmutable<'format'>                                | PHP's DateTimeImmutable object (custom format/   |
 |                                                            | default timezone)                                |
 +------------------------------------------------------------+--------------------------------------------------+
-| DateTimeImmutable<'formats', 'zone'>                       | PHP's DateTimeImmutable object (custom format/   |
+| DateTimeImmutable<'format', 'zone'>                        | PHP's DateTimeImmutable object (custom format/   |
 |                                                            | timezone)                                        |
 +------------------------------------------------------------+--------------------------------------------------+
-| DateTimeImmutable<'formats', 'zone', 'deserializeFormats'> | PHP's DateTimeImmutable object (custom format/   |
+| DateTimeImmutable<'format', 'zone', 'deserializeFormats'>  | PHP's DateTimeImmutable object (custom format/   |
 |                                                            | timezone/deserialize format). If you do not want |
 |                                                            | to specify a specific timezone, use an empty     |
 |                                                            | string (''). DeserializeFormats can either be a  |
@@ -467,7 +466,7 @@ Examples:
         private $endAt;
 
         /**
-         * @Type("DateTime<['Y-m-d', 'Y/m/d']>")
+         * @Type("DateTime<'Y-m-d', '', ['Y-m-d', 'Y/m/d']>")
          */
         private $publishedAt;
 
@@ -482,7 +481,7 @@ Examples:
         private $updatedAt;
 
         /**
-         * @Type("DateTimeImmutable<['Y-m-d', 'Y/m/d']>")
+         * @Type("DateTimeImmutable<'Y-m-d', '', ['Y-m-d', 'Y/m/d']>")
          */
         private $deletedAt;
 
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -45,8 +45,8 @@ This annotation can be defined on a property to define the serialized name for a
 property. If this is not defined, the property will be translated from camel-case
 to a lower-cased underscored name, e.g. camelCase -> camel_case.
 
-Note that this annotation is not used when you're using any other naming 
-strategy than the default configuration (which includes the 
+Note that this annotation is not used when you're using any other naming
+strategy than the default configuration (which includes the
 ``SerializedNameAnnotationStrategy``). In order to re-enable the annotation, you
 will need to wrap your custom strategy with the ``SerializedNameAnnotationStrategy``.
 
@@ -142,7 +142,7 @@ be called to retrieve, or set the value of the given property:
             $this->name = $name;
         }
     }
-    
+
 .. note ::
 
     If you need only to serialize your data, you can avoid providing a setter by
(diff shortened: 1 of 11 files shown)


goetas:
Hi! 
First of all, thanks for your contribution!

Can you add some tests for this? I'm really not able to understand how this works...

VincentLanglet:
I made a mistake. All is good now and I added tests.
Does this helps you ?

`params` can still be `['Y-m-d']` but now can be `[['Y-m-d'], ['Y/m/d'], ...]`.
In this case we try to serialize with the format 'Y-m-d', then, in case of failure, 'Y/m/d', and so on.

goetas:
I think here is missing the update on the type parser. How to provide this new type definition via `@Type` annotation as example?

VincentLanglet:
@goetas I wanted to be used this way

```
@JMS\Type("DateTime<'Y-m-d\TH:i:s'>")
@JMS\Type("DateTime<null, null, 'Y-m-d\TH:i:s'>")
@JMS\Type("DateTime<['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP']>")
@JMS\Type("DateTime<null, null, ['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP']>")
```

Are you saying `['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP']` is not actually parsed as a php array ?

goetas:
> Are you saying ['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP'] is not actually parsed as a php array ?

Exactly. I think this has to be implemented.

VincentLanglet:
> Exactly. I think this has to be implemented.

Do you know how it works and what is needed to be added/changed.

I see in both `Xml` and `Json` deserialisation visitor the function.
```php
public function visitArray($data, array $type): array
```

So it seems like it's already implemented.

goetas:
In https://github.com/schmittjoh/serializer/tree/master/src/Type you have the classes responsible for the type parsing. They convert the type definition syntax in the array of parameters passed to the visitors.

The type parsing is implemented using the hoa lexer and parser (see https://github.com/hoaproject/Compiler)

VincentLanglet:
> In https://github.com/schmittjoh/serializer/tree/master/src/Type you have the classes responsible for the type parsing. They convert the type definition syntax in the array of parameters passed to the visitors.

Yes, the type visitor call `visitCompoundType` which returns
```
[
    'name' => $nameToken->getValueValue(),
    'params' => array_map(
        function (TreeNode $node) use ($handle, $eldnah) {
            return $node->accept($this, $handle, $eldnah);
        },
        $parameters
    ),
];
```

When you look at the `accept` function of the hoa project, it calls the `visit` of the `Visitor`.
And in the `DeserializationVisitorInterface` of you project, there is already a function `visitArray`. So it's seems *already implemented*.

Can you check again ?
Or at least give a failing test with array params I need to fix ?

VincentLanglet:
@goetas You maybe missed my previous message

goetas:
> So it's seems already implemented. 

Most probably is not. You can see here the tests for it https://github.com/schmittjoh/serializer/blob/master/tests/Serializer/Type/ParserTest.php#L36

You can try to add here strings following your syntax and check if they pass.

VincentLanglet:
It seems good to me https://github.com/schmittjoh/serializer/pull/1108/commits/0ff06b7af3b8b5a10c32967f58c21126861c3fd6

It's the same syntax than 
```
'array<Foo\Bar, Baz\Boo>'
```

goetas:
Yeah, but the foundation of your proposal is based on this syntax: `DateTime<null, null, ['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP']>`, that is not tested and most probably not working

VincentLanglet:
> Yeah, but the foundation of your proposal is based on this syntax: `DateTime<null, null, ['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP']>`, that is not tested and most probably not working.
> 
> Can you test this?

Damned ! I forgot what I wanted to test...
I fixed it : https://github.com/schmittjoh/serializer/pull/1108/commits/72f15d17b612227207f8884f2e6e150f18b3f9d6

VincentLanglet:
@goetas I'm sorry, how should I understand your emoji ? 
Why are you confused ?

goetas:
I'm confused because the [grammar file](https://github.com/schmittjoh/serializer/blob/master/src/Type/grammar.pp) has no instructions on how to handle `[]`... so will have to check it on my own when will have some time

VincentLanglet:
> I'm confused because the [grammar file](https://github.com/schmittjoh/serializer/blob/master/src/Type/grammar.pp) has no instructions on how to handle `[]`... so will have to check it on my own when will have some time

I understand. Indeed, `[]` is surely not handled.
But no need to handle `[]` if you handle `array<>`.

`DateTime<null, null, array<'Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP'>>` is good enough.

VincentLanglet:
Hi @goetas, do you have time to check it on your own ?

goetas:
I had a look, and as said in https://github.com/schmittjoh/serializer/pull/1108#issuecomment-520298127
`DateTime<null, null, ['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP']>` is not handled.

`DateTime<null, null, array<'Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP'>>`, with `array` has a different semantic meaning.

@Majkl578 Could you help us to have the parser supporting `[]` array definitions?

goetas:
Ideally the tests case to pass should be 
```php
        yield [
            'DateTime<null, null, [\'Y-m-d\TH:i:s\', \'Y-m-d\TH:i:sP\']>',
            $type('DateTime', [null, null, ['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP']]),
        ];
```

VincentLanglet:
> `DateTime<null, null, array<'Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP'>>`, with `array` has a different semantic meaning.

What are the semantic meaning differences between `array` and `[]` ? This is surprising me

goetas:
`array` is a type. used for the handlers. while  `[]` are just parameters for the current type.

Majkl578:
@goetas Is this what you had in mind? #1125

goetas:
@VincentLanglet If you look at https://github.com/schmittjoh/serializer/pull/1125 , @Majkl578 did exactly what you need!

VincentLanglet:
Great ! Thanks @Majkl578

VincentLanglet:
HI @goetas What are the step now ?

1) Merging https://github.com/schmittjoh/serializer/pull/1125
2) Merging this PR

Am I missing something ?

tests/Serializer/Type/ParserTest.php
```diff
@@ -79,6 +79,10 @@ public function validTypesProvider(): iterable
             'array<Foo\Bar, Baz\Boo>',
             $type('array', [['name' => 'Foo\Bar', 'params' => []], ['name' => 'Baz\Boo', 'params' => []]]),
         ];
+        yield [
+            'DateTime<null, null, array<\'Y-m-d\TH:i:s\', \'Y-m-d\TH:i:sP\'>>',
+            $type('DateTime', [null, null, ['name' => 'array', 'params' => ['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP']]]),
```
goetas: They should be tested together to make sure it works as expected. 

Your current branch is not compatible with  https://github.com/schmittjoh/serializer/pull/1125. 
If you checkout   https://github.com/schmittjoh/serializer/pull/1125 and and test it with your code, it will fail.

`DateTime<null, null, ['Y-m-d\TH:i:s', 'Y-m-d\TH:i:sP'],`  will produce an array format that your `getDeserializationFormatsAndTimeZones`  will not be able to understand.

VincentLanglet: Wouldn't be easier if the #1125 branch was merge into master ?

VincentLanglet:
@goetas Rebased, Fixed, And tested with https://github.com/schmittjoh/serializer/pull/1125

VincentLanglet:
Hi @goetas , happy new year.
I think everything is ready in this PR. Do you have some time to look at ?

doc/reference/annotations.rst
```diff
@@ -463,6 +466,11 @@ Examples:
          */
         private $endAt;
 
+        /**
+         * @Type("DateTime<['Y-m-d', 'Y/m/d']>")
```
goetas: This is not correct. Serialization does not accept more than one format. The feature you are building in only for de-serialization

goetas:
See https://github.com/schmittjoh/serializer/pull/1108#discussion_r365565532 and I think an integration test would be a good thing to do, something like https://github.com/schmittjoh/serializer/blob/1eeb40694ce5b8d21eb4ef756273eec6c96fdd01/tests/Serializer/BaseSerializationTest.php#L675

doc/reference/annotations.rst
VincentLanglet:
```
DateTime<'formats'>
```
is a shortcut for
```
DateTime<'formats', null, 'formats'>
```
that's why I allowed the syntax.

But for serialization I used the first valid format.

If I only allow the third parameter to be an array, it means that someone who only care bout de-serialization, will write
```
@Type("DateTime<'Y-m-d'>")
```
To deserialize only one format, and
```
@Type("DateTime<'Foo', 'Bar', ['Y-m-d', 'Y/m/d']>")
```
To handle multiple syntax. I found this a little sad...

goetas: I can understand that, but `DateTime<['Y-m-d', 'Y/m/d']>` is wrong since the serialization does not accept multiple formats. it will always pick the first.

But to be hones, thinking it twice, I start to like this feature less and less. 

I had a similar problem in the past and i solved it with a custom date handler (see https://github.com/goetas-webservices/xsd2php-runtime/blob/master/src/Jms/Handler/XmlSchemaDateHandler.php#L86)

VincentLanglet: I can only allow the syntax DateTime<'Foo', 'Bar', ['Y-m-d', 'Y/m/d']> if preferred.

> But to be hones, thinking it twice, I start to like this feature less and less.

It would be nice to make your choice, I took time working on it and I prefer to know now if I taking more time on it will be useless.

> I had a similar problem in the past and i solved it with a custom date handler

I feel like, it's really hard to write a custom date handler for someone starting with the library.
When only using the annotation (like we do in our company), it's so much easier to use an array that writing a whole custom dateHandler.

VincentLanglet:
@goetas 
```
@Type("DateTime<['Y-m-d', 'Y/m/d']>")
```
Is now not supported

And there is an integration test

goetas:
thank you @VincentLanglet for your work and your patience!

PR #1110, opened 2019-07-24
adds "toArray" and "fromArray" method suggestions to IDEs for autowired s…

yashmika:
adds toArray and fromArray method suggestions to IDEs for autowired serializer

since the methods 'toArray' and 'fromArray' are not included in the SerializerInterface, IDEs (e.g. PhpStorm) do not suggest them under available methods when the serializer is injected via autowiring.

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| Doc updated   | yes/no
| BC breaks?    | yes/no 
| Deprecations? | yes/no 
| Tests pass?   | yes/no
| Fixed tickets | #... 
| License       | MIT

### DIFF
--- a/src/SerializerInterface.php
+++ b/src/SerializerInterface.php
@@ -28,4 +28,20 @@ interface SerializerInterface
      * @psalm-return T
      */
     public function deserialize(string $data, string $type, string $format, ?DeserializationContext $context = null);
+    
+    /**
+     * Converts the given data to an array.
+     *
+     * @param mixed $data
+     *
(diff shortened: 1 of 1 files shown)


goetas:
Hi! 
First of all, thanks for your contribution!

There is already [ArrayTransformerInterface](https://github.com/schmittjoh/serializer/blob/master/src/ArrayTransformerInterface.php) exactly for that propose. Use that interface if you need that typehint

PR #1117, opened 2019-08-20
JsonSerializationVisitor::visitProperty documentation update

KDederichs:
correct usage of JsonSerializationVisitor::visitProperty

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1116 
| License       | MIT

Update documentation regarding the usage of JsonSerializationVisitor::visitProperty

Issue #1116, opened 2019-08-20
StaticPropertyMetadata first constructor argument not nullable


KDederichs:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Steps required to reproduce the problem

1. Create a post serialisation listener
2. Try to add a property according to https://github.com/schmittjoh/serializer/blob/1792fe207b8addce6c65f9acd75822eacff3408e/src/JsonSerializationVisitor.php#L169

## Expected Result

* It serialises the property correctly

## Actual Result

* ```TypeError: Argument 1 passed to JMS\Serializer\Metadata\StaticPropertyMetadata::__construct() must be of the type string, null given```

Looks like the first argument is typed as not nullable string, if that value can indeed have the value null the typing in the constructor of ```StaticPropertyMetadata``` should be changed.

goetas:
This has been updated in https://github.com/schmittjoh/serializer/blob/master/UPGRADING.md

> Removed deprecated method JsonSerializationVisitor::addData, use `::visitProperty(new StaticPropertyMetadata('', 'name', 'value'), null)` instead

That file has been updated, most probably we forgot to update the comment in the file.

KDederichs:
Ok thanks for pointing that out, I do have a follow up question regarding that though:

when I do ```$visitor->visitProperty(new StaticPropertyMetadata('', 'field', 'value') , null);```
It turns out to serialise as ```null``` though instead of ```value```

When I do ```$visitor->visitProperty(new StaticPropertyMetadata('', 'field', 'value' ), 'value2');```

It put's ```value2``` in the serialised data.
That behaviour is completely different from both Documentations and I'm wondering if I'm missing something here.

goetas:
If you want to update the documentation/comment with the current behavior (that most probably is the one I had in mind when implementing the feature), a PR is welcome

goetas:
see https://github.com/schmittjoh/serializer/pull/1118

Issue #1030, opened 2019-01-12
Instructions for upgrading from addData in 1.x don't work


GKFX:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

The provided instructions for upgrading from `addData` in 1.x don't work. This is code from a project using the Symfony bundle. Instructions were ([UPGRADING.md](https://github.com/schmittjoh/serializer/blob/master/UPGRADING.md))
>Removed deprecated method JsonSerializationVisitor::addData, use :visitProperty(new StaticPropertyMetadata('', 'name', 'value'), null) instead

## Steps required to reproduce the problem

1. My old code (extract from https://github.com/camdram/camdram/blob/38216b3fe4c3b87d0f2f74d73a2521a315c0d920/src/Acts/CamdramApiBundle/Serializer/JsonEventSubscriber.php)
```php
public function onPostSerialize(ObjectEvent $event)
{
    // ...
    $visitor->addData('_links', $linkJson); // $linkJson is an array
    $visitor->addData('_type', strtolower($class->getShortName()));
}
```

2. My new code, now with library version v2.0.2, following instructions provided in UPDATING.md:
```php
$visitor->visitProperty(new StaticPropertyMetadata('', '_links', $linkJson), null);
```

3. My new code, having ignored the documentation and read the source instead ;)
```php
$visitor->visitProperty(new StaticPropertyMetadata('', '_links', null), $linkJson);
```

## Expected Result

```json
    "_links": {
        "self": "\/shows\/2020-the-phantom-of-the-opera",
        "venue": "\/venues\/demo-venue",
        "society": "\/societies\/society-10"
    },
    "_type": "show"
```
formed part of the output with with 1.x and was expected with method 2. 

## Actual Result

The expected JSON did not appear with method 2. It did appear with method 3.

Is this a bug in the code or a bug in the documentation? I.e. should I continue using method 3 or wait until a fix is released that makes method 2 work?

Sander-Toonen:
I have the same experience. 
Are you sure the documentation is right?

CHTJonas:
Bump!

goetas:
Option 3 was the right way and https://github.com/schmittjoh/serializer/pull/1066 updated the documentation

CHTJonas:
That's grand - thanks!

faysh:
[...]

PR #1118, opened 2019-08-23
Explain once and for all the use of StaticPropertyMetadata

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Fixed tickets | -
| License       | MIT

This tries to solve once and for all what has been attempted in https://github.com/schmittjoh/serializer/pull/1117 https://github.com/schmittjoh/serializer/pull/1107 and  https://github.com/schmittjoh/serializer/pull/1107

Also reported in https://github.com/schmittjoh/serializer/issues/1116  https://github.com/schmittjoh/serializer/issues/1030 

The value has to be specified twice. The first for the actual value, the second is for nullability checks and type auto detection.

### DIFF
--- a/src/JsonSerializationVisitor.php
+++ b/src/JsonSerializationVisitor.php
@@ -166,7 +166,7 @@ final class JsonSerializationVisitor extends AbstractVisitor implements Serializ
     }
 
     /**
-     * @deprecated Use visitProperty(new StaticPropertyMetadata(null, 'name', 'value'), null) instead
+     * @deprecated Use `::visitProperty(new StaticPropertyMetadata(null, 'name', 'value'), 'value')` instead
      *
      * Allows you to replace existing data on the current object element.
      *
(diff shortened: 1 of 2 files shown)


src/JsonSerializationVisitor.php
```diff
@@ -166,7 +166,7 @@ public function hasData(string $key): bool
     }
 
     /**
-     * @deprecated Use visitProperty(new StaticPropertyMetadata(null, 'name', 'value'), null) instead
+     * @deprecated Use `::visitProperty(new StaticPropertyMetadata(null, 'name', 'value'), 'value')` instead
```
KDederichs: The first argument here is still null though which will cause the type checking in PHP 7 to fail again

goetas: Good catch, fixed in https://github.com/schmittjoh/serializer/commit/80aec06a874ad196df64b371baa0d67689657596

PR #1129, opened 2019-09-30
Visitor interfaces in handlers

derzkiy:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | 
| License       | MIT

When we need override visitor we also need overriding all handlers. Can we use visitor interfaces instead final class?

### DIFF
--- a/src/Handler/ConstraintViolationHandler.php
+++ b/src/Handler/ConstraintViolationHandler.php
@@ -5,9 +5,8 @@ declare(strict_types=1);
 namespace JMS\Serializer\Handler;
 
 use JMS\Serializer\GraphNavigatorInterface;
-use JMS\Serializer\JsonSerializationVisitor;
 use JMS\Serializer\SerializationContext;
-use JMS\Serializer\XmlSerializationVisitor;
+use JMS\Serializer\Visitor\SerializationVisitorInterface;
 use Symfony\Component\Validator\ConstraintViolation;
 use Symfony\Component\Validator\ConstraintViolationList;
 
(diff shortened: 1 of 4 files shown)


goetas:
I understand the need of this PR, but handlers ofthen have visitor specific code (see https://github.com/schmittjoh/serializer/blob/4c1e4296734385af7718ca71ec0febb4815b4a87/src/Handler/DateHandler.php#L80 as example).

In my experience overriding visitors is often a bad idea... can you share the reasons why you are doing that? Maybe I can suggest a better solution...

derzkiy:
For example, before updating serializer, i could override visitor and modified field value (trim, apply some purifiers) and returned null value if it was empty.

And your example uses an interface instead class.

But I will be glad to know the properly solution

goetas:
> For example, before updating serializer, i could override visitor and modified field value (trim, apply some purifiers) and returned null value if it was empty.

I would suggest using type handlers for that. Something as:

```php
/**
 * @Type("TrimmedString")
 */
protected $name;
```
And then add a type handler for `TrimmedString`. In this way you do not need to change too many parts in the serialization process.

anyx:
@goetas using handlers for this case looks strange in my opinion: if we want to purify every string field in project, we need to add this type for every field (and override other custom handlers) for every property in project's models.

I agree that overriding visitors is often a bad idea, but I think, sometimes it's not. Anyway, it should be library user's responsibility. Now we forbid it for users (and for Barbara Liskov specially))

derzkiy:
I think its not a better solution if you need apply modifiers to all fields in all models or dto.

If you want to teach community use composition over inheritance, its good. But you have to be consistent and use SOLID principles not so selectively.

goetas:
Thanks for this PR. I did a small change in https://github.com/schmittjoh/serializer/commit/95765aa46b497469238badc2fcf5bdb6ea9f5a46 because some handlers depends explicitly on some visitor implementations

PR #1130, opened 2019-10-02
Allow to get a raw result from the serializer

alexander-schranz:
| Q             | A
| ------------- | ---
| Bug fix?      | -
| New feature?  | -
| Doc updated   | -
| BC breaks?    | -
| Deprecations? | -
| Tests pass?   | -
| Fixed tickets | #...
| License       | MIT

In the 1.x version of the jms/serializer it was possible to implement custom serializer format which doesn't return a string.

To make this again to reuse the JMS serializer logi a rawResult could be make available without that convertArrayObject is called on it.

### DIFF
--- a/src/Serializer.php
+++ b/src/Serializer.php
@@ -200,6 +200,8 @@ final class Serializer implements SerializerInterface, ArrayTransformerInterface
     }
 
     /**
+     * @param mixed $data
+     *
      * @return mixed
      */
     public function toRawResult($data, ?SerializationContext $context = null, ?string $type = null)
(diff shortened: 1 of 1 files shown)


goetas:
I'm not sure this makes sense in the context of the jms/serializer... the whole idea of serialization is to get a string. There was a big debate to add the `toArray` methods... and now the idea is to expose the underlying visitor data... Can't this be done using `toArray` ? (that exposes some kind of normalized data...)

alexander-schranz:
@goetas thank you. Yeah its definitly a workaround to recreate a behaviour which was possible in 1.x jms/serializer but think its a bad idea to reimplement this behaviour as the serializer should really only return serialized data not a mix of it. Thank you for your time and work!

PR #1139, opened 2019-11-12
Support new doctrine ODM proxy objects

notrix:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

As from version 2.0 `doctrine/mongodb-odm` uses `ocramius/proxy-manager` serializer needs to handle `LazyLoadingInterface` same as doctrine proxy objects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "doctrine/orm": "~2.1",
         "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "doctrine/phpcr-odm": "^1.3|^2.0",
-        "ocramius/proxy-manager": "^2.0",
+        "ocramius/proxy-manager": "^1.0|^2.0",
         "psr/container": "^1.0",
         "symfony/dependency-injection": "^3.0|^4.0",
         "symfony/yaml": "^3.3|^4.0",
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
         "doctrine/orm": "~2.1",
         "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "doctrine/phpcr-odm": "^1.3|^2.0",
+        "ocramius/proxy-manager": "^2.0",
         "psr/container": "^1.0",
         "symfony/dependency-injection": "^3.0|^4.0",
         "symfony/yaml": "^3.3|^4.0",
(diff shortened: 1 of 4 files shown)


src/EventDispatcher/Subscriber/DoctrineProxySubscriber.php
```diff
@@ -53,7 +53,7 @@ public function onPreSerialize(PreSerializeEvent $event): void
         }
 
         if (($this->skipVirtualTypeInit && $virtualType) ||
-            (!$object instanceof Proxy && !$object instanceof ORMProxy)
+            (!$object instanceof Proxy && !$object instanceof LazyLoadingInterface)
```
goetas: is `ORMProxy` still considered? does this work also with `ocramius/proxy-manager: 1.x`

notrix: For `ORMProxy` I found that it implements `Proxy` interface and think that it could be dropped at all because if `!$object instanceof Proxy` then it is obviously `!$object instanceof ORMProxy`. So I replaced it with `LazyLoadingInterface`

For `ocramius/proxy-manager` you are right. It should work with 1.x and even with 0.1.0 version (very first). Should I lower version in composer.json?

goetas: > For ORMProxy I found that it implements Proxy interface and think that it could be dropped at all because if !$object instanceof Proxy then it is obviously !$object instanceof ORMProxy. So I replaced it with LazyLoadingInterface 

I remember that, just if i'm not wrong wasn't so for all proxy-mamager versions

notrix: > I remember that, just if i'm not wrong wasn't so for all proxy-mamager versions

Yes, proxy-manager has LaLoIn in all version, but I'm wondering how to reflect this in composer.json?
`"ocramius/proxy-manager": "*"` or `"ocramius/proxy-manager": "^0.1|^1.0|^2.0"`?

For ORMProxy, I have double checked with `--prefer-lowest` and `Doctrine\ORM\Proxy\Proxy` is instance of `Doctrine\Common\Proxy\Proxy`. Do you think it is still worth checking?

composer.json
```diff
@@ -33,6 +33,7 @@
         "doctrine/orm": "~2.1",
         "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "doctrine/phpcr-odm": "^1.3|^2.0",
+        "ocramius/proxy-manager": "^2.0",
```
goetas:
```suggestion
        "ocramius/proxy-manager": "^1.0|^2.0",
```

src/EventDispatcher/Subscriber/DoctrineProxySubscriber.php
goetas: I think if the tests are green with https://github.com/schmittjoh/serializer/pull/1139#discussion_r346219582, then we are good to go

goetas:
Thanks

PR #1141, opened 2019-11-13
Add PHP typehints to DefaultAccessor closures

thePanz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| License       | MIT

This also checks that the accessor is not trying  to access a property that does not exist on the object. See the getValue() with  `return $o->$name ?? null;`

### DIFF
--- a/src/Accessor/DefaultAccessorStrategy.php
+++ b/src/Accessor/DefaultAccessorStrategy.php
@@ -71,7 +71,7 @@ final class DefaultAccessorStrategy implements AccessorStrategyInterface
         if (null === $metadata->getter) {
             if (!isset($this->readAccessors[$metadata->class])) {
                 if (true === $metadata->forceReflectionAccess) {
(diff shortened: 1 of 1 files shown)


goetas:
Thanks for the PR!
Much appreciated!


> This also checks that the accessor is not trying to access a property that does not exist on the object. See the getValue() with return $o->$name ?? null;

Can you create a test case for this?

thePanz:
I am trying to create a test case for this, mimicking the bug I had in my project.
Looks like it is more complicated than I expected, I'll post an update ASAP :)

thePanz:
@goetas looks like the whole issue was caused by a weird data being serialized.
The property `@var Classb[]` was an array, but of a different class.

I will try to alter this MR to provide a better exception reporting the class mismatch, so to help to identify the issue

PR #1162, opened 2020-02-08
Improved return type for fluent methods in Context

wouterj:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

While the `self` return type is true, it is too loose. What is actually returned is `$this` (i.e. the class extending this abstract class). This means the following code is - strictly spoken - invalid:

```php
$serializationContext = SerializationContext::create()
  ->setSerializeNull(true)
  ->setGroups(['api'])
;

$this->serializer->serialize($subject, 'json', $serializationContext)
```

As PHP does not have the `static` return type [until PHP 8](https://wiki.php.net/rfc/static_return_type), I think adding this PHPdoc is the best possible solution to make static analysers and IDEs happy.

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -121,6 +121,8 @@ abstract class Context
 
     /**
      * @param mixed $value
+     *
+     * @return $this
      */
     public function setAttribute(string $key, $value): self
     {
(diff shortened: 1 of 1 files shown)


goetas:
thank you!

Issue #880, opened 2018-02-28
Notice when trying to de-serizlize incorrect value type


smilesrg:
```
Notice: Array to string conversion in /var/www/html/protected/vendor/jms/serializer/src/JMS/Serializer/GenericDeserializationVisitor.php on line 61
```

It happens when trying to de-serialize a model, one of the property type is defined as a string in configuration, but actual value is an array. Maybe in this case any exceptions have to be catched and throwed some JMS\Serializer\Exception\Exception?

goetas:
Not sure about it... your underlying data looks to be an array, but you declared it as a string.... so a php error is what to expect.

smilesrg:
@goetas I think it shoul be catched and converted to exception to let developer know of incorrect configuration or incorrect value.

gitnik:
In our project we convert notices to exceptions because they often hide code smells in PHP and we thus want execution to stop. With this behaviour, we have no way to gracefully handle invalid POST data in our API. Would you be willing to accept a pull request here?

goetas:
>Would you be willing to accept a pull request here?

Generally speaking, yes!
What would be the proposed solution? typecheck each property being visited?

smilesrg:
>What would be the proposed solution? typecheck each property being visited?
I think yes, it's what needed!

PR #1167, opened 2020-02-25
Assert types before trying to convert data

gitnik:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes 
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/serializer/issues/880
| License       | MIT

As suggested in https://github.com/schmittjoh/serializer/issues/880, I used the naive approach and just assert the type of all scalars when deserializing. Feedback welcome as I'm not sure if this is the best approach.

Also I'm not sure to what degree you would consider this to be a BC break, would love to have your input on that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,8 @@
     ],
     "require": {
         "php": "^7.2",
+        "ext-simplexml":" *",
+        "ext-libxml":" *",
         "jms/metadata": "^2.0",
         "doctrine/annotations": "^1.0",
         "doctrine/instantiator": "^1.0.3",
--- a/src/AbstractVisitor.php
+++ b/src/AbstractVisitor.php
@@ -4,6 +4,11 @@ declare(strict_types=1);
 
 namespace JMS\Serializer;
 
+use JMS\Serializer\Exception\NonCastableTypeException;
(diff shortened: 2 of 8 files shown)


goetas:
I see this as a recurring problem (see also #1173) and probably should try to come up with a backward compatible solution

composer.json
```diff
@@ -17,6 +17,8 @@
     ],
     "require": {
         "php": "^7.2",
+        "ext-simplexml":" *",
```
goetas: This dependencies are optional (if someone uses it only for json), thue should be removed

goetas:
I'm willing to accept this PR even for the 3.x branch. Can you add tests please?

c0nst4nt:
Any updates? Would be nice to have this improvement!

blackandred:
Are there any updates? :wink:

goetas:
Is anyone willing to continue the work on this and add tests?

scaytrase:
@goetas I've created #1317 Feel free to tell me If we need more tests or new are located in wrong place

goetas:
done in https://github.com/schmittjoh/serializer/pull/1317

Issue #1168, opened 2020-03-11
thrown Exceptions are hidden when serializing complex objects with a handler


Hikariii:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

The try/finally block in the `SerializationGraphNavigator` making sure the context visitingStack is working correctly hides all Exceptions thrown in handlers when the context has started visiting a (nested) object but has not stopped visiting is.
This is the case in many situations where a handler (like the `ArrayCollectionHandler`) is used to serialize a more complex object (or object with objects).

Code hiding the exception is this one:
https://github.com/schmittjoh/serializer/blob/3.5.0/src/GraphNavigator/SerializationGraphNavigator.php#L196-L200
This is because `finally` is executed before the Exception is thrown and the `stopVisiting()` causes an other Exception to be thrown, hiding the first Exception.

### Running `stopVisiting()` first is sometimes necessary
The commit that introduced this change makes it clear it is created for a specific situation where serializing NULL throws an exception: https://github.com/schmittjoh/serializer/commit/ffc0feee11be577df56f3752d9bb7a5e3aa193ca

The fact that this `NotAcceptableException` is needed is also mentioned in the UPGRADING guide: https://github.com/schmittjoh/serializer/blame/3.5.0/UPGRADING.md#L56-L57

## Steps required to reproduce the problem

1. Use a custom serializationHandler and serialize object data with it.
2. In the handler start visiting a (nested) object, changing the visiting stack of the context.
3. Throw an exception while the visiting stack is visiting this nested object.

## Expected Result

* The exception thrown while parsing the compex serialization structure is shown.

## Actual Result

* The exception thrown is an error about the visitingStack not working well:
```php 
RuntimeException('Context visitingStack not working well');
```

goetas:
Thanks for working on this. I think that is something we should fix.

goetas:
fixed in https://github.com/schmittjoh/serializer/pull/1169

PR #1169, opened 2020-03-11
do not hide Exceptions from custom handlers but correctly handle null

Hikariii:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1168
| License       | MIT

This solve the issue of #1168 by catching and handling the specific `NotAcceptableException` case. All other Exceptions are not caught so Exceptions thrown in handlers are visible as requested.

### DIFF
--- a/tests/Serializer/GraphNavigatorTest.php
+++ b/tests/Serializer/GraphNavigatorTest.php
@@ -128,7 +128,32 @@ class GraphNavigatorTest extends TestCase
 
         $navigator = new SerializationGraphNavigator($this->metadataFactory, $this->handlerRegistry, $this->accessor, $this->dispatcher);
         $navigator->initialize($this->serializationVisitor, $this->context);
+        $this->context->initialize(TestSubscribingHandler::FORMAT, $this->serializationVisitor, $navigator, $this->metadataFactory);
+
         $navigator->accept($object, null);
+    }
+
+    public function testExposeAcceptHandlerExceptionOnSerialization()
+    {
+        $object = new SerializableClass();
+        $typeName = 'JsonSerializable';
+        $msg = 'Useful serialization error with relevant context information';
+
+        $handler = static function ($visitor, $data, array $type, SerializationContext $context) use ($msg) {
+            $context->startVisiting(new \stdClass());
+            throw new \RuntimeException($msg);
+        };
+        $this->handlerRegistry->registerHandler(GraphNavigatorInterface::DIRECTION_SERIALIZATION, $typeName, TestSubscribingHandler::FORMAT, $handler);
+
+        $this->context->method('getFormat')->willReturn(TestSubscribingHandler::FORMAT);
+
+        $navigator = new SerializationGraphNavigator($this->metadataFactory, $this->handlerRegistry, $this->accessor, $this->dispatcher);
+        $navigator->initialize($this->serializationVisitor, $this->context);
+        $this->context->initialize(TestSubscribingHandler::FORMAT, $this->serializationVisitor, $navigator, $this->metadataFactory);
+
+        $this->expectException(\RuntimeException::class);
+        $this->expectExceptionMessage($msg);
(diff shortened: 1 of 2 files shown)


Hikariii:
Travis is failing for a reason that does not seem related to this PR: https://travis-ci.org/github/schmittjoh/serializer/jobs/661093321#L318

tests/Serializer/GraphNavigatorTest.php
```diff
@@ -131,6 +132,37 @@ public function testNavigatorChangeTypeOnSerialization()
...
+        $context->method('getFormat')->willReturn($format);
+
+        $navigator = new SerializationGraphNavigator($this->metadataFactory, $this->handlerRegistry, $this->accessor, $this->dispatcher);
+        $navigator->initialize($this->serializationVisitor, $context);
+        $context->initialize($format, $visitor, $navigator, $this->metadataFactory);
+
+        try {
+            $navigator->accept($object, ['name' => $typeName, 'params' => []]);
+        } catch (\RuntimeException $exception) {
+            $this->assertEquals($msg, $exception->getMessage());
```
goetas: can you use `self::expectedException()` and self::expectedExceptionMessage()`  from phpunit instead of the current approach. This test will be green if for some reason the exception is not thrown.

tests/Serializer/GraphNavigatorTest.php
```diff
@@ -131,6 +132,37 @@ public function testNavigatorChangeTypeOnSerialization()
...
+        $format = 'foo';
+        $msg = 'Useful serialization error with relevant context information';
+
+        $handler = static function ($visitor, $data, array $type, SerializationContext $context) use ($msg) {
+            $context->startVisiting(new \stdClass());
+            throw new \RuntimeException($msg);
+        };
+        $this->handlerRegistry->registerHandler(GraphNavigatorInterface::DIRECTION_SERIALIZATION, $typeName, $format, $handler);
+
+        $visitor = $this->createMock(VisitorInterface::class);
```
goetas: can we avoid all this mocks and use existing `$this->serializationVisitor` and `$this->context` objects?

goetas:
Thanks for your contribution. This is something that should be fixed. The approach is ok, I've left just few comments on the tests.

tests/Serializer/GraphNavigatorTest.php
Hikariii: Good point!

tests/Serializer/GraphNavigatorTest.php
Hikariii: Yes, but because it is testing the `startVisiting` and `stopVisiting` internal methods I will have to change which methods are mocked on the `$this-context` and add the `initialize()` call to the `testNavigatorChangeTypeOnSerialization`.

tests/Serializer/GraphNavigatorTest.php
```diff
@@ -128,17 +129,42 @@ public function testNavigatorChangeTypeOnSerialization()
 
         $navigator = new SerializationGraphNavigator($this->metadataFactory, $this->handlerRegistry, $this->accessor, $this->dispatcher);
         $navigator->initialize($this->serializationVisitor, $this->context);
+        $this->context->initialize(TestSubscribingHandler::FORMAT, $this->serializationVisitor, $navigator, $this->metadataFactory);
```
Hikariii: This is added here because the `$this-context` mock is changed and initialization now needs to happen.
Also see https://github.com/schmittjoh/serializer/pull/1169#discussion_r391545324

tests/Serializer/GraphNavigatorTest.php
goetas: :+1:

goetas:
thanks

PR #1172, opened 2020-03-15
Handle ObjectConstructor returning NULL

jankramer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

The DeserializationGraphNavigator calls the construct method on an instance of ObjectConstructorInterface. According to the return type hint of that method, it is allowed to return NULL. However, when this occurs, the navigator will pass NULL to the startVisitingObject method
of the visitor. startVisitingObject expects an object to be passed in, thus causing the serializer to crash with a type error.

We can prevent this error by calling the `visitNull` method if we detect that the ObjectConstructor has returned NULL instead of an object.

### DIFF
--- a/src/GraphNavigator/DeserializationGraphNavigator.php
+++ b/src/GraphNavigator/DeserializationGraphNavigator.php
@@ -184,6 +184,13 @@ final class DeserializationGraphNavigator extends GraphNavigator implements Grap
 
                 $object = $this->objectConstructor->construct($this->visitor, $metadata, $data, $type, $this->context);
 
+                if (null === $object) {
+                    $this->context->popClassMetadata();
+                    $this->context->decreaseDepth();
+
+                    return $this->visitor->visitNull($data, $type);
+                }
+
                 $this->visitor->startVisitingObject($metadata, $object, $type);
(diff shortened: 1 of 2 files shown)


goetas:
Hi, thanks for your contribution. 
I did not realize that the signature was allowing NULL.... in my opinion should not happen... but changing that now will result in a BC break.

jankramer:
Why do you reckon it shouldn't happen? The DoctrineObjectConstructor explicitly defines a NULL strategy, which seems to have valid use cases. At least it does in the application where I encountered this situation.

As it stands now, the NULL strategy is the default strategy, but breaks when encountered at run-time. Are you open to merging this fix, or do you have a suggestion for a different approach to fixing this issue?

goetas:
I'm interested in merging this, was just hoping there was a better solution 😁

In the next week will test it and if is OK, do the merge

jankramer:
Gotcha, thanks!

PR #1173, opened 2020-03-17
Check data type before casting

agustingomes:
| Q                     | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no `(I'll update it if needed)`
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | no `(I'll add them if the idea is accepted)`
| Fixed tickets | #... 
| License       | MIT

Hi @schmittjoh 

I took the liberty of creating this PR to check the idea behind this set of proposed changes.

The idea is to avoid some edge cases that could happen when for instance, an array is sent to a property that is defined to be a string. currently it does the casting without validating the data type, unlike what happens in the array visitor.

Let me know what do you think of this approach to cover this edge case.

### DIFF
--- a/src/JsonDeserializationVisitor.php
+++ b/src/JsonDeserializationVisitor.php
@@ -54,6 +54,10 @@ final class JsonDeserializationVisitor extends AbstractVisitor implements Deseri
      */
     public function visitString($data, array $type): string
     {
+        if (!\is_string($data)) {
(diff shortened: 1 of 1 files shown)


goetas:
I see this as a recurring problem (see also https://github.com/schmittjoh/serializer/pull/1167) and probably should try to come up with a backward compatible solution

goetas:
closing in favour of https://github.com/schmittjoh/serializer/pull/1173

PR #1178, opened 2020-03-25
added optional expression parameter to the ReadOnly annotation

helmut1337:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | Yes
| Fixed tickets | 
| License       | MIT

Hello,
for my project I needed some way to dynamically define if a property is ReadOnly or not.
So I implemented the ExpressionParser the same way as the @Exclude annotation does.

Similar features has also been asked here some years ago. It's not possible with groups though.
https://github.com/schmittjoh/JMSSerializerBundle/issues/628

### DIFF
--- a/src/Exclusion/ExpressionLanguageExclusionStrategy.php
+++ b/src/Exclusion/ExpressionLanguageExclusionStrategy.php
@@ -36,7 +36,7 @@ final class ExpressionLanguageExclusionStrategy
      */
     public function shouldSkipProperty(PropertyMetadata $property, Context $navigatorContext): bool
     {
-        if (null === $property->excludeIf) {
+        if (null === $property->excludeIf && null === $property->readOnlyIf) {
             return false;
         }
 
@@ -44,38 +44,31 @@ final class ExpressionLanguageExclusionStrategy
             'context' => $navigatorContext,
             'property_metadata' => $property,
         ];
+        $readOnlyIf = false;
+        $excludeIf = false;
+
         if ($navigatorContext instanceof SerializationContext) {
             $variables['object'] = $navigatorContext->getObject();
         } else {
             $variables['object'] = null;
+
+            if (null !== $property->readOnlyIf) {
+                if (($property->readOnlyIf instanceof Expression) && ($this->expressionEvaluator instanceof CompilableExpressionEvaluatorInterface)) {
+                    $readOnlyIf = $this->expressionEvaluator->evaluateParsed($property->readOnlyIf, $variables);
+                } else {
+                    $readOnlyIf = $this->expressionEvaluator->evaluate($property->readOnlyIf, $variables);
+                }
+            }
         }
 
-        if (($property->excludeIf instanceof Expression) && ($this->expressionEvaluator instanceof CompilableExpressionEvaluatorInterface)) {
-            return $this->expressionEvaluator->evaluateParsed($property->excludeIf, $variables);
+        if (null !== $property->excludeIf) {
+            if (($property->excludeIf instanceof Expression) && ($this->expressionEvaluator instanceof CompilableExpressionEvaluatorInterface)) {
+                $excludeIf = $this->expressionEvaluator->evaluateParsed($property->excludeIf, $variables);
+            } else {
+                $excludeIf = $this->expressionEvaluator->evaluate($property->excludeIf, $variables);
+            }
         }
 
-        return $this->expressionEvaluator->evaluate($property->excludeIf, $variables);
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function shouldSkipPropertyOnDeserialize(PropertyMetadata $property, Context $navigatorContext): bool
-    {
-        if (null === $property->readOnlyIf) {
-            return false;
-        }
-
-        $variables = [
-            'context' => $navigatorContext,
-            'property_metadata' => $property,
-        ];
-        $variables['object'] = null;
-
-        if (($property->readOnlyIf instanceof Expression) && ($this->expressionEvaluator instanceof CompilableExpressionEvaluatorInterface)) {
-            return $this->expressionEvaluator->evaluateParsed($property->readOnlyIf, $variables);
-        }
-
-        return $this->expressionEvaluator->evaluate($property->readOnlyIf, $variables);
+        return $excludeIf || $readOnlyIf;
     }
 }
--- a/src/Exclusion/ExpressionLanguageExclusionStrategy.php
+++ b/src/Exclusion/ExpressionLanguageExclusionStrategy.php
@@ -72,7 +72,7 @@ final class ExpressionLanguageExclusionStrategy
         ];
         $variables['object'] = null;
 
-        if (($property->readOnly instanceof Expression) && ($this->expressionEvaluator instanceof CompilableExpressionEvaluatorInterface)) {
+        if (($property->readOnlyIf instanceof Expression) && ($this->expressionEvaluator instanceof CompilableExpressionEvaluatorInterface)) {
             return $this->expressionEvaluator->evaluateParsed($property->readOnlyIf, $variables);
         }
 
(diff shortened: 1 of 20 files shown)


goetas:
Hi. Thanks for your contribution. 
Can you please implement the same logic for the other two drivers (yml and xml)?

src/Exclusion/ExpressionLanguageExclusionStrategy.php
```diff
@@ -56,4 +56,26 @@ public function shouldSkipProperty(PropertyMetadata $property, Context $navigato
 
         return $this->expressionEvaluator->evaluate($property->excludeIf, $variables);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function shouldSkipPropertyOnDeserialize(PropertyMetadata $property, Context $navigatorContext): bool
```
goetas: I beilive that this method can be moved into `shouldSkipProperty` and use the `$navigatorContext` to understand if we are in "deserialization". In this way we do not need a new public method in this class.

helmut1337: good idea, I've changed that

helmut1337:
Hey,

I've now also added this functionality to the XML and YML drivers.

src/Exclusion/ExpressionLanguageExclusionStrategy.php
```diff
@@ -36,24 +36,39 @@ public function __construct(ExpressionEvaluatorInterface $expressionEvaluator)
...
+                    $readOnlyIf = $this->expressionEvaluator->evaluateParsed($property->readOnlyIf, $variables);
+                } else {
+                    $readOnlyIf = $this->expressionEvaluator->evaluate($property->readOnlyIf, $variables);
+                }
+            }
         }
 
-        if (($property->excludeIf instanceof Expression) && ($this->expressionEvaluator instanceof CompilableExpressionEvaluatorInterface)) {
-            return $this->expressionEvaluator->evaluateParsed($property->excludeIf, $variables);
+        if (null !== $property->excludeIf) {
```
goetas: if `$readOnlyIf` is true. you can skip the branch.

src/Metadata/Driver/YamlDriver.php
```diff
@@ -275,12 +276,25 @@ protected function loadMetadataFromFile(ReflectionClass $class, string $file): ?
...
+                        $pMetadata->readOnlyIf = $this->parseExpression($pConfig['read_only_if']);
+                        $pMetadata->readOnly = false;
+                    } else {
+                        if (isset($pConfig['read_only'])) {
+                            $pMetadata->readOnly = (bool) $pConfig['read_only'];
+                            $pMetadata->readOnlyIf = null;
+                        }
+                    }
+
+/*
```
goetas: can you avoid commenting code please?

src/Metadata/PropertyMetadata.php
```diff
@@ -113,6 +113,11 @@ class PropertyMetadata extends BasePropertyMetadata
      */
     public $readOnly = false;
 
+    /**
+     * @var string
+     */
+    public $readOnlyIf = null;
```
goetas: instead of adding one new property, we can re-use the previous one, that can be `bool|Expression`. In this way the exposed api will be simpler.

For annotations will be as you suggested `@ReadOnly(if="something")`, but for yaml and XML, we can have:

XML example:

- `read-only="true"`  as it was before
- `read-only="false"`  as it was before
- `read-only="some expression"` works as before but will be evaluated as expression.

In this way most of the works is done by the metadata mappers

goetas:
See https://github.com/schmittjoh/serializer/pull/1178/files#r399651332 that suggests a bit different approach.

goetas:
to fix automatically CS issues, you can run `vendor/bin/phpcbf`

src/Metadata/PropertyMetadata.php
helmut1337: That was actually my first approach, but then I saw how it was done on exclude-if so I thought I should go for the same schema.
I will change that.
Using the same schema on exclude-if would cause a BC break though.

src/Exclusion/ExpressionLanguageExclusionStrategy.php
helmut1337: good idea, added that

src/Metadata/Driver/YamlDriver.php
helmut1337: oops sorry, forgot to remove that block

src/Metadata/Driver/XmlDriver.php
```diff
@@ -68,6 +69,7 @@ protected function loadMetadataFromFile(\ReflectionClass $class, string $path):
         $exclude = $elem->attributes()->exclude;
         $excludeAll = null !== $exclude ? 'true' === strtolower((string) $exclude) : false;
         $classAccessType = (string) ($elem->attributes()->{'access-type'} ?: PropertyMetadata::ACCESS_TYPE_PROPERTY);
+        $readOnlyClass = $elem->attributes()->{'read-only'};
```
helmut1337: for some weird reason I really cannot find out, I cannot do parseExpression() here, because if re-use the variable $readOnlyClass in the property-loop it is only working for the first call. On the second call then, I'm getting error on unserializing the cache. But if I do parseExpression() inside the property-loop it works fine.
Makes really no sense to me.

src/Metadata/Driver/XmlDriver.php
```diff
@@ -303,16 +297,18 @@ protected function loadMetadataFromFile(\ReflectionClass $class, string $path):
...
-                    if (null !== $readOnlyIf = $pElem->attributes()->{'read-only-if'}) {
-                        $pMetadata->readOnlyIf = $this->parseExpression((string) $readOnlyIf);
-                        $pMetadata->readOnly = false;
-                    } else {
-                        if (null !== $readOnly = $pElem->attributes()->{'read-only'}) {
-                            $pMetadata->readOnly = 'true' === strtolower((string) $readOnly);
-                            $pMetadata->readOnlyIf = null;
+                    $readOnly = null === $pElem->attributes()->{'read-only'} ? $readOnlyClass : $pElem->attributes()->{'read-only'};
+                    if(null !== $readOnly) {
+                        if ($readOnly === 'true' || $readOnly === 'false') {
```
helmut1337: Thought of also parsing the true/false values as expression. But I guess it's not good for performance.

goetas: here performance are not a concern as this is done only once, then data are cached

src/Metadata/Driver/XmlDriver.php
goetas: you should cast to string the xml node

helmut1337: If you look into my previous commit, I'm doing that there.
I've noticed that it was actually only working for the caching-call. The second call fails with error:

> Return value of JMS\Serializer\Metadata\PropertyMetadata::unserializeProperties() must be of the type string, null returned [] []

It's really weird because if I do the exact same thing inside of the property-loop, it is working.
So putting this block line 90-96 (previous commit):
`        $readOnlyClass = false;
        $readOnlyIfClass = null;
        if(null === $readOnlyIf = $elem->attributes()->{'read-only-if'}) {
            $readOnlyClass = 'true' === strtolower((string) $elem->attributes()->{'read-only'});
        } else {
            $readOnlyIfClass = $this->parseExpression((string) $readOnlyIf);
        }`

into the loop starting at line 174, it works 

It's just confusing me and I wanted to find out how this happens.
But it does not really matter because doing parseExpression on every porperty will maybe only affect the performance on when creating cache

src/Metadata/Driver/XmlDriver.php
helmut1337: as I saw it is only caching the expression-string and not the result, which makes sense because the expression should be evaluated on deserialize to be dynamically.
And then it would always evalute expressions with "true" and "false" on deserialize.
Not sure if this will produce impact at all though.

helmut1337:
Removed read-only-if tag in yml and xml. read-only is now also parsing expressions.
guess I'm done :)

helmut1337:
hey,
wanted to ask if there's anything missing from my side?

Sorato95:
when this will be merged?

PR #1184, opened 2020-04-06
Embed Hoa\Compiler

### LINKED ISSUE #1182: [RFC] Removing abandoned hoa from serializer
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | yes

Hi guys,

jms serialize use hoa/compiler for parse some data, but looks like hoa project was abandoned since 2018, for now it have problem with php 7.4 opache.preload (https://github.com/hoaproject/Consistency/pull/36, https://github.com/hoaproject/Consistency/issues/35), and nobody react for this. JMS serializer is popular library, that used by lot of peoples in lot of project, but abandoned dependency blocking that projects from migrate to php 7.4, from my side situation looks not very good. 

So depends on above, I propose make decision to remove HOA project from dependency, and use something different, or fork HOA project to another repository, merge exists pull request, and forget about problems with php 7.4.

Best regards.

sanmai:
- [ ] TODO: description is outdated

I've been using [this fork](https://github.com/sanmai/hoa-compiler) for some time in production, and necessary to say I took at least some effort to make sure it works as it should. For example, tests were migrated to PHPUnit, but we still run the original test suite to make sure it's even.

There's a slight breaking change for packages that depend both on the serializer and the compiler: `Hoa\Exception\Exception`-type exceptions are no longer thrown as the compiler does not depend on the exception abstractions it previously used. This is due to I tried to minimize the number of potentially abandoned dependencies, removing unnecessary for day to day operations.

Second breaking change is in the fact that the compiler package no longer includes dev dependecies as non-dev. E.g. if one wants to use `getAST()` to make new Ln compilers on the fly, they'll need to include additional dependencies. The serializer is not affected.

The above change also makes the serializer non-dependent on hoa/consistency, therefore hoaproject/Consistency#35 is no longer a problem when using the serializer.

Currently I'm the sole maintainer for the fork, but I'll be happy to accept a co-maintainer or two. (CI procedure isn't as robust as I'd want it, but that's not a big deal to fix. We'll fix it once we get there.)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | yes TBD
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1182
| License       | MIT

Fixes #1182


goetas:
Personally I would have preferred if the fork was using explicitly a different namespace... so it looked just as a different package.

sanmai:
Fair enough. Currently it uses original `Hoa\Compiler` namespace. What do you think would work better?

sanmai:
Never mind previous question, we can include the bulk of the compiler package in a sub-namespace of the very package. Updated dependencies are only `hoa/iterator` and `hoa/visitor`.

PR #1187, opened 2020-04-19
Fix Support conditional exclude for classes

arneee:
In #1099 support for ExcludeIf on class level was added, unfortunately the condition was inverted.

Current behavior would actually exclude this class:
```php
/**
 * @Serializer\Exclude(if="false")
 */
class PersonAccount
```

However it should only exclude the class if the condition evaluates to true:
```php
/**
 * @Serializer\Exclude(if="true")
 */
class PersonAccount
```

This was due to a copy mistake from _Expose_, which internal also uses _Exclude_, but with the condition negated and also due to a test scenario which was not very clear to udnerstand.

 I have simplified the test to make it more clear and fixed the drivers.


goetas:
Thanks!

PR #1031, opened 2019-01-14
Fix for failing doctrine object constructor on embeddable class

notrix:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no
| License       | MIT

Actual issue is on doctrine/persistence component. I have registered an issue to them https://github.com/doctrine/persistence/issues/37
The case is that doctrine is using XML metadata driver returns isTransient() === false on embeddable class and this class could not be found in database as it has no identifiers.
I have added one additional check to ensure that we have identifier and test that uses different entity manager with XML metadata driver.

PR #1189, opened 2020-04-20
Sending #1031 fix to 1.x branch

notrix:
As requested in https://github.com/schmittjoh/serializer/pull/1031#issuecomment-615556681 sending #1031 fix to 1.x branch


goetas:
Tests are missing. You could have cherry-picked that single commit

PR #1197, opened 2020-05-11
Fix deprecated assertFileNotExist

mpoiriert:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "symfony/form": "^3.0|^4.0|^5.0",
         "symfony/filesystem": "^3.0|^4.0|^5.0",
         "symfony/expression-language": "^3.0|^4.0|^5.0",
-        "phpunit/phpunit": "^7.5||^8.0||^9.0",
+        "phpunit/phpunit": "^8.0||^9.0",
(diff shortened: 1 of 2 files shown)


goetas:
I think that you should increase the minimum required version of phpunit, as some builds are failing

mpoiriert:
Yes you are right. I have remove 7 since it reach it's end of support

mpoiriert:
So it's not compatible with PHPUnit 8.0 either.

Do you want me to remove it ? If there is no need to have a older version require dev. PHPUnit 8 is compatible with php 7.2 so it should be ok.

goetas:
I think that removing phpunit 7 is ok, regarding your changes, unfortunately I think that we have to postpone them

mpoiriert:
There is always a way.

I did that so we don't have the deprecated warning, it's a temporary fix.

goetas:
nice! thanks!

PR #1198, opened 2020-05-12
GroupsExclusionStrategy code refactoring / comment for readability

mpoiriert:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | yes/no 
| Deprecations? | no 
| Tests pass?   | yes/no
| Fixed tickets | #... 
| License       | MIT

This code optimize event more the groups exclusion strategy by relaying on isset in any case.

The path => groups get prepared in constructor and then we just need to build the path from the context and use isset from there on each group of the property.

The issue I have is that the GroupsExclusionStrategy::getGroupsFor is public but not part of the interface. The method is not use in the package itself other than in the test itself. 
The test failed since I the method getGroups for doesn' return the same structure (in fact it's not needed anymore).

All the integration test that rely on the GroupsExclusionStrategy does pass.

Since the getGroupsFor and the class GroupsExclusionStrategy we should pass by a deprecations flow and reimplement the getGroupsFor. Since it's not suppose to be use directly I could just combine both code and use the old way in the code.

The last section is not part of the PR yet.

### DIFF
--- a/src/Exclusion/GroupsExclusionStrategy.php
+++ b/src/Exclusion/GroupsExclusionStrategy.php
@@ -12,10 +12,17 @@ final class GroupsExclusionStrategy implements ExclusionStrategyInterface
 {
     public const DEFAULT_GROUP = 'Default';
 
+    private $groups;
+
     /**
      * @var array<string,array<string,boolean>>
      */
     private $parsedGroups = [];
+
+    /**
+     * @var array<string,boolean>>
+     */
+    private $rootGroups = [];
 
     /**
      * @var bool
@@ -24,7 +31,11 @@ final class GroupsExclusionStrategy implements ExclusionStrategyInterface
 
     public function __construct(array $groups)
     {
-        $this->prepare($groups, 'root');
+        $this->groups = $groups;
+        $this->prepare($groups, '');
+        $this->parsedGroups['.'] = $this->parsedGroups[''];
+        unset($this->parsedGroups['']);
+        $this->rootGroups = $this->parsedGroups['.'];
     }
 
     private function prepare(array $groups, $path)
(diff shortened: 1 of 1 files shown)


mpoiriert:
FYI I did some benchmark with this script (replicating the behaviour in one of the test)

```
$groups = [
    GroupsExclusionStrategy::DEFAULT_GROUP,
    'manager_group',
    'friends_group',

    'manager' => [
        GroupsExclusionStrategy::DEFAULT_GROUP,
        'friends_group',

        'friends' => ['nickname_group'],
    ],
    'friends' => [
        'manager_group',
        'nickname_group',
    ],
];

$adrien = new GroupsUser(
    'John',
    new GroupsUser(
        'John Manager',
        null,
        [
            new GroupsUser(
                'John Manager friend 1',
                new GroupsUser('John Manager friend 1 manager')
            ),
            new GroupsUser('John Manager friend 2'),
        ]
    ),
    [
        new GroupsUser(
            'John friend 1',
            new GroupsUser('John friend 1 manager')
        ),
        new GroupsUser(
            'John friend 2',
            new GroupsUser('John friend 2 manager')
        ),
    ]
);

$oldStrategy = new \JMS\Serializer\Exclusion\GroupsExclusionStrategy($groups);
$newStrategy = new \JMS\Serializer\Exclusion\NewGroupsExclusionStrategy($groups);


$propertyNickname = new \JMS\Serializer\Metadata\PropertyMetadata(GroupsUser::class, 'nickname');
$propertyNickname->groups = ['nickname_group'];

$propertyManager = new \JMS\Serializer\Metadata\PropertyMetadata(GroupsUser::class, 'manager');
$propertyManager->groups = ['manager_group'];

$propertyFriends= new \JMS\Serializer\Metadata\PropertyMetadata(GroupsUser::class, 'friends');
$propertyFriends->groups = ['friends_group'];

$context = SerializationContext::create()->setGroups($groups);

$context->getMetadataStack()->push(
    $propertyManager
);

$context->getMetadataStack()->push(
    $propertyFriends
);

testPerformance('Old', function () use ($oldStrategy, $propertyNickname, $context) {
    $oldStrategy->shouldSkipProperty($propertyNickname, $context);
});

testPerformance('New', function () use ($newStrategy, $propertyNickname, $context) {
    $newStrategy->shouldSkipProperty($propertyNickname, $context);
});

function testPerformance($name, Closure $closure, $runs = 1000000)
{
    $start = microtime(true);
    for (; $runs > 0; $runs--)
    {
        $closure();
    }
    $end = microtime(true);

    printf("Function call %s took %.5f seconds\n", $name, $end - $start);
}
```

Here are the result:

```
Function call Old took 13.70450 seconds
Function call New took 8.95419 seconds
```

FYI The class NewGroupsExclusionStrategy doesn't exist in the PR it's just so both are available in my test.

There is a gain only in with  nested groups. More complex the nest is bigger are the gain.

goetas:
Nice. I like it, but please give me some time to test it as this feature has many side effects not fully covered by tests.

mpoiriert:
I have made some other performance test and the "prepare" portion do have a impact. You have pretty much no difference of a none nested groups so that's ok and you see a performance gain when you check around 50 attribute on a nested groups. I think that when you end up configuring nested group you have more than 50 attribute to check. Just to clarify 50 overall, meaning 10 object with 5 attribute each will see a speed improvement.

I did try to assign a callback in the prepare to decide which function to execute (since we already know if it's a nested group or not) but calling a callback function take more time than doing the if...

I saw other improvement that can be done but would need to deprecate some other method too (In the context for example).

PR #1200, opened 2020-05-13
add ObjectManagerAwareConstructor

re2bit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT


re2bit:
mh... sorry build is failing. have to check it this evening.

tests/Serializer/Doctrine/ObjectConstructorTest.php
```diff
@@ -217,6 +221,19 @@ public function testNamingForIdentifierColumnIsConsidered()
...
     }
 
+    public function testObjectManagerIsInjected(): void
+    {
+        $serializer = $this->createSerializerWithObjectManagerAwareFallback();
+        $jsonData = '{"name":"Hello Thing"}';
+        /** @var Thing $thing */
+        $thing = $serializer->deserialize($jsonData, Thing::class, 'json');
+
+        static::assertInstanceOf(ObjectManager::class, $thing->getObjectManager());
```
goetas, line 231: can you modify the test to be able to use `assertSame` and check the exact instance?

re2bit: i dont use symfony at work but i will try. maybe this weekend.

re2bit:
i dont use symfony at work but i will try. maybe this weekend.

tests/Serializer/Doctrine/ObjectConstructorTest.php
re2bit: done

re2bit:
@goetas I looked at the SymfonyBundle and now think it would a better way to inject the ManagerRegistry instead of the ObjectManagerInterface. I Upgrade my pull request soon together with a Pull Request for the Symfony bundle.

goetas:
Sounds good, looking forward to it.

re2bit:
@goetas I Looked this Weekend at the Symfony Bundle and tried to add the ObjectManagerAwere Constructor. 
But I encountered some Problems. 

First of all I´m typically use only Symfony Components rather than the whole Framework so I had to work me through the Dependency Injection Documentation first.

I tried to add the ObjectManager Aware Constructor as Decorator for any existing Fallback Constructor by updating the DependencyInection/Compiler/DoctrinePass Class but had no Success. 

In fact I had no success to enable the DoctrineObjectConstructor at all with DoctrinePass.
The DoctrinePass code is executed but still the UnserializeObjectConstructor is used for my Controller.

I found Bug #391 and #559 which both address this Problem in some way.

After changing Line 37 of JMSSerializerBundle.php
```
$builder->addCompilerPass(new DoctrinePass(), PassConfig::TYPE_BEFORE_REMOVING);
```
back to
```
$builder->addCompilerPass(new DoctrinePass(), PassConfig::TYPE_BEFORE_OPTIMIZATION);
```
the Doctrine Object Constructor was used by default.

I am using PHP v7.4.5, Symfony v4.14.4 and FOS Rest Bundle v2.7.4

As far I can tell there is a Bug that prevents the ObjectConstructor and DoctrineTypeDriver to be used by default and DoctrinePass is kind of dead Code right now.

I think I need some more Time to investigate this deeply. Im afraid a Change to this Code could lead to serious problems if People are not aware that their Doctrine Entities are loaded during deserialize process in case of changing the DoctrinePass Code.

goetas:
Hi, thanks for your efforts. Is  your work on the serializer library ready? If yes, can try to fix the bundle issues. Please let me know

goetas:
@re2bit 

> In fact I had no success to enable the DoctrineObjectConstructor at all with DoctrinePass.
The DoctrinePass code is executed but still the UnserializeObjectConstructor is used for my Controller.

this was a messy part in the bundle code. as of https://github.com/schmittjoh/JMSSerializerBundle/pull/803 this part should not be a problem anymore

PR #1202, opened 2020-05-13
[HandlerRegistry] case sensitive default methods

wickedOne:
ucfirst for the serialization format while constructing default methods in the handler registry

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Doc updated   | no
| BC breaks?    | yes?
| Deprecations? | no 
| Tests pass?   | yes
| License       | MIT

this fixes an issue where the default DateHandler deserialization methods were improperly defined in a case sensitive environment as well as those of the FormErrorHandler.

not entirely sure whether this is considered a bc break as people might have worked aroud this issue by modifying their method names?

### DIFF
--- a/src/Handler/HandlerRegistry.php
+++ b/src/Handler/HandlerRegistry.php
@@ -23,10 +23,10 @@ class HandlerRegistry implements HandlerRegistryInterface
 
         switch ($direction) {
             case GraphNavigatorInterface::DIRECTION_DESERIALIZATION:
-                return 'deserialize' . $type . 'From' . $format;
+                return 'deserialize' . $type . 'From' . ucfirst($format);
 
             case GraphNavigatorInterface::DIRECTION_SERIALIZATION:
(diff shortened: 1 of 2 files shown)


goetas:
This looks reasonable to me, but PHP methods are not case sensitive, what is the reason for changing  this?

wickedOne:
that's odd... 

my use case is that i constructed a serializer
```php
$this->serializer = SerializerBuilder::create()
     // ...
    ->configureHandlers(static function (HandlerRegistry $registry) {
        $registry->registerSubscribingHandler(new DateHandler());
        // ...
     })
      // ...
     ->build()
 ;
```
which didn't invoke the ``dateHandler::deserializeDateTimeFromXml`` method when using the following annotation.
```php
    /**
     * @Serializer\Type("DateTime")
     */
    private $active;
```
using the modifications from this pr, the method was called, so i assumed it came down to a case sensitivity issue

Issue #1208, opened 2020-05-28
Enable resolving collection type from the var doblock


dgafka:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | yes

Currently JMS Serializer can resolve type declaration from properties (PHP 7.4), yet can not resolve the type, when type is collection (array of specific classes).

The type however is available for us in most of the time in form of the docblock `@var`. 

## Steps required to reproduce the problem

Right now, in order to describe the collection we need to do it like this:

```php
	/**
	 * @JMS\Type("array<Application\Message>")
	 * @var Message[]
	 */
	private array $messages;
``` 
Unfortunately this solution has disadvantages. 
1) We duplicate the type in the form of the `@var `and as well `@JMS\Type`
2) IDE refactoring tools are not able to find and replace, when the class name is changed
3) We can resolve types from property declaration without any additional configuration, however not when it's collection. 

In PHP we are having convention to describe collections as `@var` docblock and mark it with suffix `[]`. 
This is respected by IDE, works with refactoring tools and is widely used in the community in order to describe the collection. 
What I would propose is to fallback to the docblock, when declared type is array. 

Previous code would looks like this:

```php
	/**
	 * @var Message[]
	 */
	private array $messages;
```

## Expected Result

Transforming Message[] into the form respected by JMS array<Application\Message>

## Actual Result

Require annotation in order to work

----

I have built such feature on top of JMS Serializer in integration with [my framework](https://docs.ecotone.tech/modules/jms-converter#example-usage). 
However I think such feature could be really nice addition build into JMS Serializer directly. 

I am willing to prepare implementation for that, if you will find it useful. :)

goetas:
That sound a good idea. I'm curious on how is implemented. Is it something similar to what i did to detect php 7.4 types in https://github.com/schmittjoh/serializer/pull/1192 ?

goetas:
Just saw https://github.com/ecotoneframework/jms-converter/blob/master/src/JMSConverter/ClassDefinitionDriver.php , nice job! :+1:

PR #1214, opened 2020-06-14
Resolve collections from DocBlock

dgafka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/serializer/issues/1208
| License       | MIT

There are few things worth discussion:

1) What should happen, when incorrect DocBlock type is given?

       /**
       * @var \stdClass
       */
       public array $productIds;

Current implementation will throw exception, as above has no sense

2) What should happen, when union type is given?

        /**
        * @var ProductName[]|ProductDescription[]
        */
       public array $productIds;

Current implementation will throw exception, as we will not know, which class should be used for deserialization. 

3) What should happen, when class does not exists?

        /**
        * @var NotExistingClass[]
        */
       public array $productIds;

Current implementation will throw exception as, this class will not be deserializable.

4) If  exception is thrown in any case of `1)` and `2)` `3)` what exception type it should be? 

Current implementation will throw \InvalidArgumentException


dragosprotung:
Any chance to support also this format: 
```
/**
 * @var array<ProductName>
 */
public array $productIds;
```

dgafka:
@dragosprotung yes, I can add it :)

@goetas can you deliver feedback about current implementation and answer above questions? 
So I can know, if current behaviour is expected and continue with possible fixes. :)

phpcs.xml.dist
```diff
@@ -71,6 +71,15 @@
             <property name="searchAnnotations" type="boolean" value="true"/>
         </properties>
     </rule>
+    <rule ref="SlevomatCodingStandard.Namespaces.MultipleUsesPerLine.MultipleUsesPerLine">
```
goetas, line 74: can you explain the reason for this (and the other two) change in the coding style rules?

src/Metadata/Driver/TypedPropertiesDriver.php
```diff
@@ -33,15 +33,20 @@ class TypedPropertiesDriver implements DriverInterface
...
+    /**
+     * @var DocBlockTypeResolver
+     */
+    private $docBlockTypeResolver;
 
     /**
      * @param string[] $whiteList
      */
-    public function __construct(DriverInterface $delegate, ?ParserInterface $typeParser = null, array $whiteList = [])
+    public function __construct(DriverInterface $delegate, DocBlockTypeResolver $docBlockTypeResolver, ?ParserInterface $typeParser = null, array $whiteList = [])
```
goetas, line 44: do we need to pass it as external dependency ? from what is see is a hard dependency for this class (and there are no interfaces for it..).
I would suggest to instantiate it inside.

Another (nicer?) approach would be to let it implement a `DriverInterface` and implement this kind of behavior for all the `array` types, not only for the one detected by php 7.4.

src/Metadata/Driver/TypedPropertiesDriver.php
```diff
@@ -79,7 +85,12 @@ public function loadMetadataForClass(ReflectionClass $class): ?ClassMetadata
             try {
                 $propertyReflection = $this->getReflection($propertyMetadata);
                 if ($this->shouldTypeHint($propertyReflection)) {
-                    $propertyMetadata->setType($this->typeParser->parse($propertyReflection->getType()->getName()));
+                    $type = $propertyReflection->getType()->getName();
+                    if ('array' === $type) {
+                        $type = $this->docBlockTypeResolver->getPropertyDocblockTypeHint($propertyReflection);
```
goetas, line 90: is there a case where `getPropertyDocblockTypeHint` is not able to resolve the type?

phpcs.xml.dist
dgafka: As types in DocBlocks may be typed partially only by the Class Name `Product` instead of Namespace + Class Name `SomeNamespace\Product`, we need to resolve them using `use statements`. 
This allows us to cover the case with group use statement, which is not part of PSR standard, however may be used in user land :)

src/Metadata/Driver/TypedPropertiesDriver.php
dgafka: I have changed declaration to keep the depedency hidden. It will be instantiate inside now.

I did implement it as separate Driver. During the implementation I found out, that it will be weird from the user perspective, why something like this can be serialized

    /**
    * @var Product[]
    */
    private array $products;

and something like this not:

    /**
    * @var Product
    */
    private $product;

This could lead to confusion, so both are correct now and we will be serialized correctly. 
Second thing is it will lower barrier, as PHP 7.4 is not needed anymore in order to use this feature.
The last and probably most valuable benefit. It will be much easier to move to `JMS Serializer` from the code, not using typed properties, as there will be no need for users to annotate it with `@Type` in each of the class properties. 
So install and use, if you have types defined in DocBlock :)

src/Metadata/Driver/TypedPropertiesDriver.php
dgafka: Implementation has changed. Questions from the description are not current anymore, and this code is different.
DocBlockType resolver will return no type, if there is no `DocBlock` or `@var in DocBlock`

dgafka:
@goetas Proposal of @dragosprotung is implemented. 
This will be possible:

     /**
     * @var array<Product>
     */
    private array $products;

To keep this compatible with `JMS Serializer` documentation this will also be possible:

     /**
     * @var array<string, Product>
     */
    private array $products;

dgafka:
@goetas I would like to ask for review now. 
There are 4 failing tests left, I am not able to fix them, as I don't understand why they fail. Can you check them out?

goetas:
This looks great so far! I will try it in the next days/weeks and see if I can help with the failing tests

dgafka:
@goetas will you find time to check the tests?  :)

dgafka:
@goetas can you tell, if you will find time for verifying tests?

I understand that you may have not enough time for it. 
So please tell me, if you will not find the time, then I will try to fix it on my own :)

goetas:
I have tested this and looks great, but is seems that needs to be targeting a 4.0 release...

the issue is the following code that works as expected now, and is broken if we were going to merge this PR.


```php
class User {
     /**
     * @var Country
     */
     private $country;

     public function __construct($country) 
     {
         $this->country = $country;
     }
}

$user = new User(new Car()); // yes we are passing a "car", not a country, it is not a typo

```

Even if the following code is "wrong", it works. A user will be serialized and in the "country" there will be serialized the car object.

If this PR gets merged, the serializer will try ti traverse the "car" by using the props defined for "country".

This is not an uncommon scenario... immagine just how many docblocks might be not up to date in a project.... and now suddenly applications will start crashing.

With Typed props this was not an issue is since PHP will do the check for us and tell the user that the code was wrong... but in our case there is no such check.

dgafka:
This is understandable, you would like to avoid breaking change for current major release.  

Should I leave the MR to you now and wait for 4.0 release? :)

goetas:
> Should I leave the MR to you now and wait for 4.0 release? :)

@dgafka Sadly I have to say yes. The 4.0 is not yet planned but there are a couple of things I would like to place into it (but have no time to work  on it....).

The only option I can see to have it in 3.x to make it opt-in, so who is interested can enable it and deal with potential bc breaks? What do you think? (i think it could be possible just by removing the instantiation of it from https://github.com/schmittjoh/serializer/pull/1214/files#diff-31d5d899bd07e7ac0794ccc1b0facdeaR64 and move it to the serializer builder).

Thanks a lot again for your great work!

dgafka:
@goetas I have added `DocBlock Type Resolver` as opt in feature. :)

You can enable it by using method `setDocBlockTypeResolver` on `SerializerBuilder `

    builder->setDocBlockTypeResolver(true)

dgafka:
@goetas will we make it to 3.x after recent change? :)

dgafka:
If I understand you were trying to replace the part which resolves the docblock types for an external library. 
Plus of the current solution is less dependencies on the JMS\Serializer :)  

Thank you for merging it :)  
  
I am looking forward for next MR, in order to optimize deserializing process. I feel like version 3, had big impact on that as the times increased.
However that's a plan for a future, as I will need to understand the library deeper :)

goetas:
>If I understand you were trying to replace the part which resolves the docblock types for an external library.
> Plus of the current solution is less dependencies on the JMS\Serializer :)

yeah. the advantage of the other lib is that is fully tested and support very exotic syntaxes to express types (see PHPStan)...

>I am looking forward for next MR, in order to optimize deserializing process. 

Looking forward to it.

>I feel like version 3, had big impact on that as the times increased.

did not know that... IMO should be faster than 1.x... but I might be wrong.

goetas:
>However that's a plan for a future, as I will need to understand the library deeper :)

thanks again for this contribution

simPod:
@dgafka how to enable DocBlockDriver?

dgafka:
@simPod you can see example usage in here:
https://github.com/schmittjoh/serializer/pull/1214/commits/901f0729fdba5987f996cbb9e6604ae4017df2b7#diff-9739e997b0a9ef5ef5a7b3304d94d26ba594544fd4d8604c9f0f43262fd40d55R269

simPod:
I see. I'll have to do it differently if I use the sf bundle. Thanks.

PR #1222, opened 2020-06-30
fixed exception for strict_types

ivoba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT

When using declare(strict_types=1); null will not work in RuntimeException constructor and will result in this error:

`Wrong parameters for JMS\Serializer\Exception\RuntimeException([string $message [, long $code [, Throwable $previous = NULL]]])`


goetas:
Thanks

PR #1227, opened 2020-07-04
Make SubscribingHandlerInterface::getSubscribingMethods() return iterable

simPod:
I'm using generators inside of this method so this would allow them to be used officially

| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| License       | MIT


goetas:
is adding this method signature a BC break? how would it behave if someone tries to override it ?

simPod:
You're right, it was `array` before so it's BC

proggeler:
If this is a BC break I would suggest to make this function no longer static. It is not used that way.

simPod:
Not sure if it's BC break. array is covariant to iterable.

Tobion:
The BC break is that implementations of the interface without the return type won't work anymore due to signature incompatibility.

simPod:
@Tobion afaik `implementations of the interface without the return type` are compatible

```php
interface Foo
{
    public function fcn() : iterable;
}

class Bar implements Foo
{
    public function fcn()
    {
    }
}
```

Tobion:
No it's not: https://3v4l.org/c8lWB

PR #1236, opened 2020-07-11
Add support for filterable (de)serialization handlers

bobvandevijver:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

I need to be able to choose between my handler and the default serializer implementation depending on the current deserialization depth, but I couldn't find any way to do so. Therefor, I added support for filterable handlers, which contain a `shouldBeSkipped` method which is called when the handler implements the `FilterableSubscribingHandlerInterface` interface. This way, you can fallback to the default implementation if you choose to do so from you own handler.

### DIFF
--- a/doc/handlers.rst
+++ b/doc/handlers.rst
@@ -78,3 +78,25 @@ Also, this type of handler is registered via the builder object::
         })
     ;
 
+Filterable Subscribing Handlers
+-------------------------------
+
+In case you need to be able to fall back to the default deserialization behavior instead of using your custom
+handler, you can implement the `FilterableSubscribingHandlerInterface` instead of the default
+`SubscribingHandlerInterface`. This interface add a single `shouldBeSkipped` method, which will be called
+before the handler will be used. If it returns `true`, the custom handler will be skipped and the default
+implementation of the serializer will be used instead::
+
+    use JMS\Serializer\Handler\SubscribingHandlerInterface;
+    use JMS\Serializer\GraphNavigator;
+    use JMS\Serializer\JsonSerializationVisitor;
+    use JMS\Serializer\JsonDeserializationVisitor;
+    use JMS\Serializer\Context;
+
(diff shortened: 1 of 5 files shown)


bobvandevijver:
The spacing rule for the doc comment is kinda interesting, I must say...

goetas:
Hi, thanks for your contribution. Unfortunately I think that  this can not be accepted as it will negatively impact on performance , increase the event handler complexity and it and splits the decision logic across two methods.

A better approach was proposed in https://github.com/schmittjoh/serializer/pull/177, by using `SkipStepException` that allows a 
handler to decline the handling of a type.
Unfortunately that PR was lacking tests and had as part of it another much more complex feature, so it never got merged.

There was also https://github.com/schmittjoh/serializer/pull/620 that proposed the same thing, but was also lacking tests. In that specific case the conversation derailed on a completely different direction (was also my fault, I should have handled better that conversation, I was missing some context and it was also a transition period between the 1.x and 2.x release...).

If you are interested in having such feature, would you be interested in continuing the work done in   https://github.com/schmittjoh/serializer/pull/620 by adding tests to it? With tests (and the extra context i've gained in the last years) I think think that there should be no problems in merging such feature.

bobvandevijver:
@goetas Thanks for your feedback! I created #1238, which should address it. If you meant anything else, please just let me know.

I must say, I was thinking about another way of doing this, by returning a dedicated object instead of throwing an exception. That would skip the need for a try/catch construct in the deserializer (the serializer already has one, so it doesn't really matter there I guess), but then you would need to test if that object is returned.

goetas:
Replaced by https://github.com/schmittjoh/serializer/pull/1238

PR #1238, opened 2020-07-12
Add support for skippable (de)serialization handlers

bobvandevijver:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | - 
| License       | MIT

I need to be able to choose between my handler and the default serializer implementation depending on the current deserialization depth, but I couldn't find any way to do so. Therefor, I added support for skippable handlers, by simply throwing a `SkipHandlerException` to do so.

This PR is an updated version of #1236, created as separate PR to be able to illustrate the differences between both easily and to be able to choose which one is more suitable. I did not adjust the test method implementations at all compared to the first PR, only the implementation of the TestSubscriber used in the tests has been adjusted to match throw the required exception. 

@goetas I hope this is what you meant with your [comment](https://github.com/schmittjoh/serializer/pull/1236#issuecomment-657188521), otherwise, please let me know!

### DIFF
--- a/doc/handlers.rst
+++ b/doc/handlers.rst
@@ -78,3 +78,8 @@ Also, this type of handler is registered via the builder object::
         })
     ;
 
+Skippable Subscribing Handlers
+-------------------------------
+
+In case you need to be able to fall back to the default deserialization behavior instead of using your custom
+handler, you can simply throw a `SkipHandlerException` from you custom handler method to do so.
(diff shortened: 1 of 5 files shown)


goetas:
Thanks a lot!

PR #1241, opened 2020-07-27
Change return type of SerializerBuilder::build() to Serializer

icanhazstring:
Using the `SerializerBuilder` build will return an instance of `Serializer` which implements `SerializerInterface` and `ArrayTransformerInterface`.
So the return type of `SerializerInterface` is wrong in this case.

| Q             | A
| ------------- | ---
| Bug fix?      | sort of
| New feature?  |no
| Doc updated   |no
| BC breaks?    |no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/src/SerializerBuilder.php
+++ b/src/SerializerBuilder.php
@@ -499,10 +499,7 @@ final class SerializerBuilder
         return $this;
     }
 
-    /**
-     * @return SerializerInterface&ArrayTransformerInterface
-     */
-    public function build()
+    public function build(): Serializer
     {
         $annotationReader = $this->annotationReader;
         if (null === $annotationReader) {
--- a/src/SerializerBuilder.php
+++ b/src/SerializerBuilder.php
@@ -499,7 +499,10 @@ final class SerializerBuilder
         return $this;
     }
 
-    public function build(): SerializerInterface
+    /**
+     * @return SerializerInterface&ArrayTransformerInterface
+     */
+    public function build()
     {
         $annotationReader = $this->annotationReader;
         if (null === $annotationReader) {


src/SerializerBuilder.php
```diff
@@ -499,7 +499,10 @@ public function setMetadataCache(CacheInterface $cache): self
         return $this;
     }
 
-    public function build(): SerializerInterface
+    /**
+     * @return SerializerInterface&ArrayTransformerInterface
```
goetas, line 503: is this syntax supported? as I know php does not have yet intersection types

icanhazstring: Yes it is, seen here:
- https://blog.jetbrains.com/phpstorm/2018/09/phpstorm-2018-3-eap-183-2635-12/
- https://phpstan.org/blog/union-types-vs-intersection-types
- https://github.com/squizlabs/PHP_CodeSniffer/pull/2919
- https://psalm.dev/docs/annotating_code/type_syntax/intersection_types/

goetas:
the ci job https://travis-ci.org/github/schmittjoh/serializer/jobs/712201589 is failing

goetas:
Since the serializer builder is marked as final, we can just typehint the build method with `Serializer`.

icanhazstring:
@goetas we could do this as well. You are right :)
I just wanted to adhere to the interface approach. Will change that.

goetas:
thanks!

goetas:
Thanks

icanhazstring:
Will this be tagged in a patch release soon?

PR #1257, opened 2020-10-25
Added default_skip_when_empty option as config using the exclusionStrategy

slava-v:
Extended the context to include the option to enable SkipWhenEmpty as a default behaviour for properties.

The ExclusionStrategy approach was used to enable this.

It is also related to https://github.com/schmittjoh/JMSSerializerBundle/issues/789 And I will prepare a pull request for that repo. also when this will be merged

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/JMSSerializerBundle/issues/789 
| License       | MIT


goetas:
Hi, thanks for the PR. I like the idea, but I'm not sure that works as expected. IMO it will skip all the props when enabled. To check it, can you please also add few props in `ObjectWithEmptyArrayAndHashNotAnnotated` that are not empty, and check that those do not get skipped?

src/Context.php
```diff
@@ -101,7 +108,7 @@ public function getNavigator(): GraphNavigatorInterface
         return $this->navigator;
     }
 
-    public function getExclusionStrategy(): ?ExclusionStrategyInterface
+    public function getExclusionStrategy(): ?DisjunctExclusionStrategy
```
slava-v, line 111: If we change this return type, then "we delegate" the control of what interfaces the "exclusionStrategy" property could contain. 

In this case i extended it with a new interface "ValueExclusionStrategyInterface" that defines a method which accepts a `value` also, beside the `metadata` and `context`

src/Context.php
```diff
@@ -142,28 +149,21 @@ public function setAttribute(string $key, $value): self
     }
 
     /**
+     * @param ExclusionStrategyInterface|ValueExclusionStrategyInterface $strategy
+     *
      * @return $this
      */
-    public function addExclusionStrategy(ExclusionStrategyInterface $strategy): self
+    public function addExclusionStrategy($strategy): self
```
slava-v, line 156: I changed the strong type hint for the argument. I annotated accordingly 

But the argument validation is performed in the `$this->exclusionStrategy->addStrategy()`

src/Exclusion/ExpressionLanguageExclusionStrategy.php
```diff
@@ -20,7 +20,7 @@
  *
  * @author Asmir Mustafic <goetas@gmail.com>
  */
-final class ExpressionLanguageExclusionStrategy
+final class ExpressionLanguageExclusionStrategy implements ExclusionStrategyInterface
```
slava-v, line 23: I saw that this class is also a part of "exclusionStrategy", it was missing the interface implementation.

src/Metadata/Driver/AnnotationDriver.php
```diff
@@ -186,6 +186,7 @@ public function loadMetadataForClass(\ReflectionClass $class): ?BaseClassMetadat
                         $propertyMetadata->serializedName = $annot->name;
                     } elseif ($annot instanceof SkipWhenEmpty) {
                         $propertyMetadata->skipWhenEmpty = true;
+                        // @Todo: Check the context about the configuration whether skipWhenEmpty on all properties.
```
slava-v, line 189: remove (

tests/Handler/ArrayCollectionHandlerTest.php
```diff
@@ -48,7 +49,7 @@ public function testSerializeArraySkipByExclusionStrategy()
         $factoryMock = $this->getMockBuilder(MetadataFactoryInterface::class)->getMock();
         $factoryMock->method('getMetadataForClass')->willReturn(new ClassMetadata(ArrayCollection::class));
 
-        $context->method('getExclusionStrategy')->willReturn(new AlwaysExcludeExclusionStrategy());
+        $context->method('getExclusionStrategy')->willReturn(new DisjunctExclusionStrategy([new AlwaysExcludeExclusionStrategy()]));
```
slava-v, line 52: repercussion after changing the return type for [Context::getExclusionStrategy()](https://github.com/schmittjoh/serializer/pull/1257/files#diff-3cce1ea0450771adbc3cae800dc672c1edde9f76fc377da115e1ca05c8da926cR111)

slava-v:
@goetas , you were right. After adding the test  tests/Fixtures/ObjectWithEmptyArrayAndHashNotAnnotated.php i realized where was the mistake. 
I did some refactorings of the implementation and, basically, extended the Exclusion mechanism with additional interface that describes signature for an exclusion method that would accept a property value also, whenever the case.

i came to this implementation, which seems cleaner in my view. I considering the:
* approach of `expressionExclusionStrategy` private property assigned in constructor -- bad: one more prop. polouting class
* extending the original `ExclusionStrategyInterface` with another method -- bad: other classes would have to implement the method 

Maybe there are any other ideas that i haven't thought about ?

slava-v:
Hi @goetas i've rebased the pull-request branch. 

I'll check if i can improve/lower the complexity reported for  
[SerializationGraphNavigator](https://scrutinizer-ci.com/g/schmittjoh/serializer/inspections/2c2d89f3-18f0-4181-a6a0-0ef84dfdf5e9/code-structure/class/JMS%5CSerializer%5CGraphNavigator%5CSerializationGraphNavigator)

Issue #1259, opened 2020-10-30
DocBlockDriver unable to parse single-line DocBlocks


Namoshek:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Steps required to reproduce the problem

1. Use the new `$builder->setDocBlockTypeResolver(true)` feature from #1214.

2. Define single-line DocBlocks to save some space:
    ```php
    class Foo
    {
        /** @var Bar[] */
        public array $x;

        /** @var \Some\Namespace\Bar[] */
        public array $y;
     }
    ```

## Expected Result

Parses successfully.

## Actual Result

Throws exception(s):

**Case `Bar[]`:**
> InvalidArgumentException: Can't use incorrect type Bar[] */ for collection in Some\Namespace\Foo:x

**Case `\Some\Namespace\Bar[]`:**
> JMS\Serializer\Type\Exception\SyntaxError: Syntax error, unexpected "*/" (T_UNKNOWN), expected was T_COMMA

## Additional Information

The same code with multi-line DocBlocks parses and de-/serializes successfully:
```php
class Foo
{
    /**
     * @var Bar[]
     */
    public array $x;

    /**
     * @var \Some\Namespace\Bar[]
     */
    public array $y;
}
```

Interestingly enough, using the `array<Class>` syntax also works with single-line DocBlocks:
```php
class Foo
{
    /** @var array<Bar> */
    public array $x;

    /** @var array<\Some\Namespace\Bar> */
    public array $y;
}
```

Namoshek:
Thank you for this cool addition by the way! Combining this with attributes and named parameters from PHP 8 will be my absolute favorite way to go in future. My vision would be something like this (some random modifier ideas thrown into the mix as well 😄):
```php
class Foo
{
    /** @var Bar[] */
    #[Serializer\Property(
        name: 'theBars',
        nullIfEmpty: true,              // for deserialization of strings and arrays
        nullIfWhitespace: true,         // for deserialization of strings
        ignoreIfNull: true,             // for serialization
        ignoreIfEmpty: true,            // for serialization of strings and arrays
        ignoreIfWhitespace: true,       // for serialization of strings
        ignoreIfNullOrWhitespace: true, // for serialization of strings
    )]
    public ?array $bars;
}
```
_Note: not all added parameters make sense with an array of objects... it's just to illustrate._

Namoshek:
[...]

PR #1256, opened 2020-10-15
Allow interfaces for DocBlock

marein:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #...
| License       | MIT

Currently the `DocBlockDriver` only reads classes (abstract included) and some php defined types. For interfaces, additional metadata in the form of annotations or files must be specified. This pull request makes this obsolete.

PR #1261, opened 2020-10-31
Use phpstan/phpdoc-parser to retrieve additional type information from PhpDoc

Namoshek:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1259
| License       | MIT

As suggested by @goetas in #1259, this PR fixes the mentioned issue by implementing `phpstan/phpdoc-parser` support to read `@var` tags from PhpDoc. Only one test case for single-line PhpDoc parsing has been added, which works fine now.

In my opinion, a BC break is introduced by adding a constructor to `DocBlockTypeResolver`, since none existed before (which means possible extensions probably do not contain and call to the parent constructor so far).

This needs to be reviewed carefully, as I'm not sure if I considered all possible cases. The tests seem to work fine though.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         "doctrine/phpcr-odm": "^1.3|^2.0",
         "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "ocramius/proxy-manager": "^1.0|^2.0",
+        "phpstan/phpdoc-parser": "^0.4",
         "phpunit/phpunit": "^8.0||^9.0",
         "psr/container": "^1.0",
         "symfony/dependency-injection": "^3.0|^4.0|^5.0",
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
         "doctrine/annotations": "^1.0",
         "doctrine/instantiator": "^1.0.3",
         "doctrine/lexer": "^1.1",
-        "jms/metadata": "^2.0"
+        "jms/metadata": "^2.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -35,6 +35,7 @@
         "doctrine/phpcr-odm": "^1.3|^2.0",
         "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "ocramius/proxy-manager": "^1.0|^2.0",
+        "phpstan/phpdoc-parser": "^0.4",
```
goetas, line 38: i think that this should stay in the `require` section, not `require-dev`

goetas:
except only one comment, this looks good to me. @dgafka can you also check it please?

goetas:
since https://github.com/schmittjoh/serializer/pull/1256 has been merged, this needs to be rebased.

composer.json
Namoshek: You are totally right. No idea how it ended up in the dev dependencies.

Namoshek:
I changed the dependency location and rebased.

#1256 is very welcomed by me as well by the way. since I was about to write a custom handler for `CarbonInterface` and realized it doesn't work just yet. 😄

goetas:
yay!

goetas:
thanks a lot!

Issue #379, opened 2014-12-31
User defined ordering function


ogizanagi:
Hi there.

Many thanks for the amazing work on this library.

I wonder if a custom user defined ordering function might be useful ?
**Concrete case:** All my entities use a `TimestampableTrait` (very similar to the one provided by `DoctrineExtensions`), adding both `createdAt` and `updatedAt` fields. Those fields are declared with proper annotations in order to allow serialization. (`VirtualProperty` on a custom method or `Expose` directly on the trait properties).

As for readability, the traits use is declared at the top of my entities, without any defined order policy (`ACCESSOR_ORDER_UNDEFINED`), this results in the fact that `created_at` and `updated_at` fields are always before any other fields after serialization.

For this use case, the aim is to provide a way to use a custom ordering function, which will always put `created_at` and `updated_at` fields at the bottom.

I think there might be some other use cases.

What do you think ?

> **Note:** I guess VirtualProperty are still ignored as mentioned in the #349. So, this will only work for properties for now.

goetas:
Closing, Sorry for the long feedback loop.

This can be already solved with http://jmsyst.com/libs/serializer/master/reference/annotations#accessororder

PR #1272, opened 2020-12-19
Allow to defined custom property ordering strategy

alamirault:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

This PR allow to register new properties order strategy in addition of `undefined`, `alphabetical`, `custom`.
Useful examples can be Reverse Alphabetical Order or set virtual properties after properties in serialization.

There is a BC breaks on signature of `PropertyOrderingInterface` but this interface was used only in `ClassMetadata`. Not sure if I must add this in UPGRADING.md. If I must updating this file, what will be section name ?

If this PR is accepted and merged, I will make changes in JMSSerializerBundle in order to register new strategies when implementing interface (like EventDisptacher)

(Resolve closed issue https://github.com/schmittjoh/serializer/issues/379)

### DIFF
--- a/src/Metadata/ClassMetadata.php
+++ b/src/Metadata/ClassMetadata.php
@@ -9,6 +9,8 @@ use JMS\Serializer\Expression\Expression;
 use JMS\Serializer\Ordering\AlphabeticalPropertyOrderingStrategy;
 use JMS\Serializer\Ordering\CustomPropertyOrderingStrategy;
 use JMS\Serializer\Ordering\IdenticalPropertyOrderingStrategy;
+use JMS\Serializer\Ordering\PropertyOrderingRegistry;
+use JMS\Serializer\Ordering\PropertyOrderingRegistryInterface;
 use Metadata\MergeableClassMetadata;
 use Metadata\MergeableInterface;
 use Metadata\MethodMetadata;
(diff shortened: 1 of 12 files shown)


goetas:
It would be great if this could be implemented in a backward compatible way... I do not have plans in releasing 4.0 anytime soon.

alamirault:
I created a new interface in order to avoid BC.

Scrutinizer failure is not related of my changes

goetas:
Yea, but now classes as `AlphabeticalPropertyOrderingStrategy` are implementing a different interface... and that interface has a different signature than the one you have provided

goetas:
what about having a brand new interface 


```php
interface SomeSorter
{
    // returns a callable compatible with uasort
   function getSorter(ClassMetadata $props):callable;
}
```

alamirault:
I understand.
I can also deprecate all `*PropertyOrderingStrategy` based on original interface and create new based on the new interface named  `*PropertiesOrderingStrategy`.
In this way we are sure that classes returns a new order, `SomeSorter` depending of implementation can do other stuffs that sorter right ?

PR #1273, opened 2021-01-06
Add throws tag

VincentLanglet:
Hi @goetas 

Your implementation of this interface is throwing Runtime exceptions, so it should be good to add the `@throws` tag to the interface ; this way the developer is aware to catch them.

### DIFF
--- a/src/SerializerInterface.php
+++ b/src/SerializerInterface.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace JMS\Serializer;
 
+use JMS\Serializer\Exception\RuntimeException;
+
 /**
  * Serializer Interface.
  *
@@ -15,6 +17,8 @@ interface SerializerInterface
      * Serializes the given data to the specified output format.
      *
      * @param mixed $data
+     *
+     * @throws RuntimeException
      */
     public function serialize($data, string $format, ?SerializationContext $context = null, ?string $type = null): string;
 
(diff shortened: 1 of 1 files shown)


VincentLanglet:
I don't think the scrutinizer error is related to my PR.

Are you interested @goetas ? :)

goetas:
TBH I do not know.

PHP (contrary to Java) does not force you to catch the declared exceptions, so do not see really a value in declaring them.

It would just make a lot of noise, and it we want to be strict we would have to place hundreds of `@throw SomeException`  all over the code

VincentLanglet:
Hi

> PHP (contrary to Java) does not force you to catch the declared exceptions, so do not see really a value in declaring them.

To me, it's like saying (when we were in php5)
> PHP does not force you to respect `@param` and `@return` so do not see really value declaring them.

PHP does not force you, but some static analysis like phpstan/psalm can help you to report uncaught exceptions. But only if they are declared in the phpdoc, of course. 

I got an error in production because I didn't catch the exception, and nothing helps me to know they were an exception to catch. This would have avoid such error. 

> It would just make a lot of noise

Why noise ? Phpdoc is documentation
As an example, symfony interfaces have `@throws` tags: https://github.com/symfony/contracts/search?q=throws

> it we want to be strict we would have to place hundreds of @throw SomeException all over the code

Yes, maybe. But if so, that does mean that currently, people are using method which can throws exception without knowing it ! That's seems to be a risky thing.

People shouldn't wait to get an exception to know they have to catch something.

goetas:
Anyway I think that this will now hurt, so it is ok.

VincentLanglet:
Thanks

VincentLanglet:
Does a 3.11.1 release would be possible ? @goetas :)

PR #1274, opened 2021-01-11
Enable RequireOneLinePropertyDocComment sniff

simPod:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   |no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   |
| Fixed tickets | #... 
| License       | MIT


goetas:
TBH i do not see a value in this, it was intentionally disabled that sniff (and phpstorm automatically uses the multiline version...)

simPod:
I thought that for easier maintenance it would be best to comply with Doctrine standard. That said removing all custom settings, exclusions etc.

Eg. this sniff saves a lot of file vertical space. Phpstorm has somehow outdated file templates, that's true.

In last few years phpdoc are started disappearing rapidly as we have param, return and property types. Will get event better with annotations and union types, mb even generics.

PR #1276, opened 2021-01-11
Introduce PHPStan

simPod:

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -4,9 +4,3 @@ parameters:
 			message: "#^Class JMS\\\\Serializer\\\\Handler\\\\iterable not found\\.$#"
 			count: 2
 			path: src/Handler/IteratorHandler.php
-
-		-
-			message: "#^Variable \\$previousEntityLoaderState might not be defined\\.$#"
-			count: 1
-			path: src/XmlDeserializationVisitor.php
-
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,12 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Class JMS\\\\Serializer\\\\Handler\\\\iterable not found\\.$#"
+			count: 2
+			path: src/Handler/IteratorHandler.php
+
+		-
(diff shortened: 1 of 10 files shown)


phpstan-baseline.neon
```diff
@@ -0,0 +1,12 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Class JMS\\\\Serializer\\\\Handler\\\\iterable not found\\.$#"
```
goetas, line 4: hmm, this is weird... why do we get this error?

phpstan-baseline.neon
```diff
@@ -0,0 +1,12 @@
...
+			count: 2
+			path: src/Handler/IteratorHandler.php
+
+		-
+			message: "#^Variable \\$previousEntityLoaderState might not be defined\\.$#"
```
goetas, line 9: was it possible to fix this?

goetas (commented):
nice! i like it!

Can you please have a look at the two comments i've left?

phpstan-baseline.neon
simPod: I don't know yet. Could not reproduce it on phpstan playground.

phpstan-baseline.neon
simPod: check now

phpstan-baseline.neon
goetas: maybe it needs just a `\`?

simPod: it's not an object so cannot be imported. It's like `string`.

https://phpstan.org/r/84508559-53e5-43da-b083-219d3b9af0a5

I was thinking that would investigate it later, that's why I added it to baseline for now.

phpstan.neon.dist
```diff
@@ -0,0 +1,14 @@
+parameters:
+    level: 1
```
goetas, line 2: hmm, level 1, @simPod do you plan in increasing the level with future pull requests?

simPod: Yup

phpstan-baseline.neon
Tobion: I found the problem, see https://github.com/schmittjoh/serializer/pull/1315

PR #1295, opened 2021-03-09
Doctrine `json` field type can contain not only an array

gam6itko:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT

In my project I have database table with `json` field which can contain not only `array`, but `string`, `int`, `bool` etc.
Using release 3.12.0 is not possible for me


goetas:
Thanks

Issue #1297, opened 2021-03-12
DoctrineObjectConstructor calls Doctrine with null identifier columns


Vincentv92:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | yes
| RFC?             | yes

## Steps required to reproduce the problem

1. Managed entity by doctrine with a single `@ORM\Id` notation
2. From XML deserialisation to managed doctrine object
3. XML contains an ID field which does not exists yet.

```
composer show | grep jms
jms/metadata                         2.5.0     Class/method/property metadata management in PHP
jms/serializer                       3.12.0    Library for (de-)serializing data of any complexity; supports XML, JSON, and YAML.
jms/serializer-bundle                3.8.0     Allows you to easily serialize, and deserialize data of any complexity
composer show | grep doctrine
doctrine/annotations                 1.12.1    Docblock Annotations Parser
doctrine/cache                       1.10.2    PHP Doctrine Cache library is a popular cache implementation that supports many different drivers such as redis, memcache, apc, mongodb and others.
doctrine/collections                 1.6.7     PHP Doctrine Collections library that adds additional functionality on top of PHP arrays.
doctrine/common                      3.1.1     PHP Doctrine Common project is a library that provides additional functionality that other Doctrine projects depend on such as better reflection support, proxies and much more.
doctrine/dbal                        2.10.4    Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection and management.
doctrine/doctrine-bundle             2.2.3     Symfony DoctrineBundle
doctrine/event-manager               1.1.1     The Doctrine Event Manager is a simple PHP event system that was built to be used with the various Doctrine projects.
doctrine/inflector                   2.0.3     PHP Doctrine Inflector is a small library that can perform string manipulations with regard to upper/lowercase and singular/plural forms of words.
doctrine/instantiator                1.4.0     A small, lightweight utility to instantiate objects in PHP without invoking their constructors
doctrine/lexer                       1.2.1     PHP Doctrine Lexer parser library that can be used in Top-Down, Recursive Descent Parsers.
doctrine/orm                         2.8.2     Object-Relational-Mapper for PHP
doctrine/persistence                 2.1.0     The Doctrine Persistence project is a set of shared interfaces and functionality that the different Doctrine object mappers share.
doctrine/sql-formatter               1.1.1     a PHP SQL highlighting library
[...]

PR #1299, opened 2021-03-17
#1297: Fix deserialization of doctrine entity with null identifier

Toilal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1297
| License       | MIT


goetas:
hmm, this sounds strange to me... the `find` method should be supposed to find entities using primary keys.... how is this feature going to behave if we have an entity identified by multiple columns?

Something like 
```php
$authorFetched = $constructor->construct($this->visitor, $class, ['id1' => null, 'id2' => 123 ], $type, $this->context);
```

Toilal:
@goetas I don't have this use case in my current projects, so I can't really test composite primary keys for now. 

According to this issue https://github.com/doctrine/orm/issues/6647, Doctrine doesn't support null fields in a composite primary key. So we probably want to skip find invocation when a single field is null.

PR #1301, opened 2021-03-22
`Undefined offset: 0` when using `@var null|string` instead of `@var string|null`

ruudk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Doc updated   |no
| BC breaks?    |no
| Deprecations? | no 
| Tests pass?   | see CI
| License       | MIT

After upgrading from 3.12.0 to 3.12.1 I received `Undefined offset: 0` errors. After some debugging I found the problem.

In the commits I explain what the problem is and how I fixed it.

### DIFF
--- a/src/Metadata/Driver/DocBlockTypeResolver.php
+++ b/src/Metadata/Driver/DocBlockTypeResolver.php
@@ -143,9 +143,9 @@ class DocBlockTypeResolver
      */
     private function filterNullFromTypes(array $types): array
     {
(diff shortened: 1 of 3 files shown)


ruudk:
@goetas Thanks for the merge. Can this be tagged as the latest one is broken. 🙏

goetas:
https://github.com/schmittjoh/serializer/releases/tag/3.12.2

PR #1305, opened 2021-03-24
Bump doctrine/coding-standard to v8.2

simPod:


goetas:
What is the value in this change? composer is going to use the latest anyway when possible, right?

simPod:
Yes so this is only to prevent older versions

scyzoryck:
Thanks for contribution! Already bumped to version 12.0.

PR #1313, opened 2021-04-21
useful error when data is not an object

dbu:
| Q             | A
| ------------- | ---
| Bug fix?      | (yes)
| New feature?  | no
| Doc updated   | -
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | see ci
| Fixed tickets | -
| License       | MIT

I was spending some time trying to figure out why i get:

    TypeError: "Argument 2 passed to JMS\Serializer\JsonSerializationVisitor::startVisitingObject() must be an object, array given

Turns out i misconfigured a property that was supposed to be an array of objects as an object with the JMS annotations. From the error it is impossible to know what is wrong. With this exception, the path to the offending property is printed and we say what is actually expected.

### DIFF
--- a/src/GraphNavigator/SerializationGraphNavigator.php
+++ b/src/GraphNavigator/SerializationGraphNavigator.php
@@ -236,7 +236,7 @@ final class SerializationGraphNavigator extends GraphNavigator
                 }
 
                 if (!is_object($data)) {
-                    throw new InvalidArgumentException('Value at '.$this->context->getPath().' is expected to be an object of class '.$type['name'].' but is of type '.gettype($data));
+                    throw new InvalidArgumentException('Value at ' . $this->context->getPath() . ' is expected to be an object of class ' . $type['name'] . ' but is of type ' . csgettype($data));
                 }
                 
                 $this->context->pushClassMetadata($metadata);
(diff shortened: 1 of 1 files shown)


src/GraphNavigator/SerializationGraphNavigator.php
```diff
@@ -234,6 +234,10 @@ public function accept($data, ?array $type = null)
                     throw new ExcludedClassException();
                 }
 
+                if (!is_object($data)) {
+                    throw new InvalidArgumentException('Value at '.$this->context->getPath().' is expected to be an object of class '.type['name'].' but is of type '.gettype($data));
+                }
```
dbu, line 239: the startVisitingObject below has the type declaration `object $data`, so we get a TypeError exception if we call it with a non-object.

src/GraphNavigator/SerializationGraphNavigator.php
```diff
@@ -234,6 +235,10 @@ public function accept($data, ?array $type = null)
                     throw new ExcludedClassException();
                 }
 
+                if (!is_object($data)) {
+                    throw new InvalidArgumentException('Value at '.$this->context->getPath().' is expected to be an object of class '.$type['name'].' but is of type '.gettype($data));
```
dbu, line 239: getPath only gives the model class but not the current property. do we know the name of the property we are currently looking at?

goetas:
thanks!

src/GraphNavigator/SerializationGraphNavigator.php
dbu: would probably need some setup, especially for a unit test. i think annotating a model property to be some object and assigning an array to it (or a string/int/bool) and then running serialization should trigger it.

PR #1315, opened 2021-04-23
fix iterable::class that does not exist

Tobion:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

### DIFF
--- a/src/Handler/IteratorHandler.php
+++ b/src/Handler/IteratorHandler.php
@@ -77,14 +77,14 @@ final class IteratorHandler implements SubscribingHandlerInterface
     }
 
     /**
-     * @return array|\ArrayObject
+     * @return array|\ArrayObject|null
      */
     public function serializeIterable(
         SerializationVisitorInterface $visitor,
         iterable $iterable,
         array $type,
         SerializationContext $context
-    ): iterable {
+    ): ?iterable {
         $type['name'] = 'array';
 
         $context->stopVisiting($iterable);
--- a/src/Handler/IteratorHandler.php
+++ b/src/Handler/IteratorHandler.php
@@ -28,22 +28,6 @@ final class IteratorHandler implements SubscribingHandlerInterface
         foreach (self::SUPPORTED_FORMATS as $format) {
             $methods[] = [
                 'direction' => GraphNavigatorInterface::DIRECTION_SERIALIZATION,
-                'type' => iterable::class,
-                'format' => $format,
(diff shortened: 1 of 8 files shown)


src/Handler/IteratorHandler.php
```diff
@@ -25,22 +25,6 @@ public static function getSubscribingMethods()
     {
         $methods = [];
 
-        foreach (self::SUPPORTED_FORMATS as $format) {
-            $methods[] = [
-                'direction' => GraphNavigatorInterface::DIRECTION_SERIALIZATION,
-                'type' => iterable::class,
```
Tobion, line 31: iterable is a psydo class and does not really exist. this is resolved as a class in the current namespace, i.e. `JMS\Serializer\Handler\iterable` which does not exist. this is why phpstan complained rightfully.
also the whole registration of `iterable` is pointless because iterables are already handled directly in the graph visitor (like for array, int etc)

src/Handler/IteratorHandler.php
```diff
@@ -124,25 +108,6 @@ public function deserializeIterator(
         return new ArrayIterator($visitor->visitArray($data, $type));
     }
 
-    /**
-     * @param mixed $data
-     */
-    public function deserializeIterable(
```
goetas, line 130: does it mean that we do not have anymore the deserializaer for iterable?

src/Handler/IteratorHandler.php
goetas: why not put just `'iterable'` as string?

Tobion: Because it wouldn't do anything? iterables are not handled via registration.

src/Handler/IteratorHandler.php
Tobion: we do. it's inside the graph visistor/navigator itself.

Tobion: see https://github.com/schmittjoh/serializer/blob/f908d17afd08f0aab9c083322022682b41483c5b/src/GraphNavigator/DeserializationGraphNavigator.php#L137-L138

goetas: uh, i forgot about that!

goetas:
thanks!

PR #1317, opened 2021-04-30
Check data can be casted before actual casting

scaytrase:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #880 
| License       | MIT

This is retake of #1167 I've cleaned it up and added some tests

I'd rather consider this a bug fix than a new feature since I came here because my tests are failing because of warnings `Array to string conversion` are being caught by phpunit, not by my library exception system.

### DIFF
--- a/src/Exception/NonCastableTypeException.php
+++ b/src/Exception/NonCastableTypeException.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Exception;
+
+class NonCastableTypeException extends RuntimeException
+{
+    /**
+     * @var mixed
+     */
+    private $value;
+
+    /**
+     * @param mixed $value
+     */
+    public function __construct(string $expectedType, $value)
+    {
+        $this->value = $value;
+
+        parent::__construct(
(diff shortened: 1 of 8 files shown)


src/Exception/NonCastableTypeException.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Exception;
+
+class NonCastableTypeException extends RuntimeException
```
goetas, line 7: should this class be abstract?

scaytrase: why not. 

also children classes could be declared final?

goetas: that sounds a good idea!

scaytrase: fixed

goetas:
thanks!

scaytrase:
Sorry, but can this be released?

PR #1320, opened 2021-05-18
Added PHP 8 attributes

joaojacome:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

### DIFF
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -1,5 +1,52 @@
 Annotations
------------
+===========
+
+PHP 8 support
+-------------
+
+JMS serializer now supports PHP 8 attributes, with a few caveats:
+- To avoid BC breaks, all annotations must use named parameters.
+- Due to the missing support for nested annotations, the syntax for a few annotations has changed
+
+Converting your annotations to attributes
+-----------------------------------------
+
+Example:
+
+.. code-block :: php
+
+    /**
+     * @VirtualProperty(
+     *     "classlow",
+     *     exp="object.getVirtualValue(1)",
+     *     options={@Until("8")}
+     * )
+     * @VirtualProperty(
+     *     "classhigh",
+     *     exp="object.getVirtualValue(8)",
+     *     options={@Since("6")}
+     * )
+     */
+    #[VirtualProperty(name: 'classlow', exp: 'object.getVirtualValue(1)', options: [[Until::class, ['8']]])]
+    #[VirtualProperty(name: 'classhigh', exp: 'object.getVirtualValue(8)', options: [[Since::class, ['6']]])]
+    class ObjectWithVersionedVirtualProperties
+    {
+        /**
+         * @Groups({"versions"})
+         * @VirtualProperty
+         * @SerializedName("low")
+         * @Until("8")
+         */
+        #[Groups(groups: ['versions'])]
+        #[VirtualProperty]
+        #[SerializedName(name: 'low')]
+        #[Until(version: '8')]
+        public function getVirtualLowValue()
+        {
+            return 1;
+        }
+    ...
+
 
 @ExclusionPolicy
 ~~~~~~~~~~~~~~~~
@@ -265,6 +312,22 @@ In this example:
 .. note ::
 
     This only works for serialization and is completely ignored during deserialization.
+
+In PHP 8, due to the missing support for nested annotations, in the options array you need to pass an array with the class name and an array with the arguments for its constructor.
+
+.. code-block :: php
+
+    /**
+     * @Serializer\VirtualProperty(
+     *     "firstName",
+     *     exp="object.getFirstName()",
+     *     options={@Serializer\SerializedName("my_first_name")}
+     *  )
+     */
+    #[Serializer\VirtualProperty(name: "firstName", exp: "object.getFirstName()", options: [[Serializer\SerializedName::class, ["my_first_name"]]])]
+    class Author
+    {
+    ...
 
 @Inline
 ~~~~~~~
--- a/src/Annotation/AccessType.php
+++ b/src/Annotation/AccessType.php
@@ -10,11 +10,27 @@ namespace JMS\Serializer\Annotation;
  *
  * @author Johannes M. Schmitt <schmittjoh@gmail.com>
  */
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_PROPERTY)]
 final class AccessType
 {
     /**
      * @Required
-     * @var string
+     * @var string|null
      */
     public $type;
+
+    public function __construct(array $values = [], ?string $type = null)
+    {
+        if ([] !== $values) {
+            if (array_key_exists('value', $values)) {
+                $type = $values['value'];
+            }
+
+            if (array_key_exists('type', $values)) {
+                $type = $values['type'];
+            }
+        }
+
+        $this->type = $type;
+    }
 }
--- a/src/Annotation/AccessType.php
+++ b/src/Annotation/AccessType.php
@@ -21,14 +21,12 @@ final class AccessType
 
     public function __construct(array $values = [], ?string $type = null)
     {
-        if ([] !== $values) {
-            if (array_key_exists('value', $values)) {
-                $type = $values['value'];
-            }
+        if (array_key_exists('value', $values)) {
+            $type = $values['value'];
+        }
 
-            if (array_key_exists('type', $values)) {
-                $type = $values['type'];
-            }
+        if (array_key_exists('type', $values)) {
+            $type = $values['type'];
         }
 
         $this->type = $type;
--- a/src/Annotation/AccessorOrder.php
+++ b/src/Annotation/AccessorOrder.php
@@ -12,16 +12,37 @@ namespace JMS\Serializer\Annotation;
  *
  * @author Johannes M. Schmitt <schmittjoh@gmail.com>
  */
+#[\Attribute(\Attribute::TARGET_CLASS)]
 final class AccessorOrder
 {
     /**
      * @Required
-     * @var string
+     * @var string|null
      */
-    public $order;
+    public $order = null;
 
     /**
      * @var array<string>
      */
     public $custom = [];
+
+    public function __construct(array $values = [], ?string $order = null, array $custom = [])
+    {
+        if ([] !== $values) {
+            if (array_key_exists('value', $values)) {
+                $order = $values['value'];
+            }
+
+            if (array_key_exists('order', $values)) {
+                $order = $values['order'];
+            }
+
+            if (array_key_exists('custom', $values)) {
+                $custom = $values['custom'];
+            }
+        }
+
+        $this->order = $order;
+        $this->custom = $custom;
+    }
 }
--- a/src/Annotation/AccessorOrder.php
+++ b/src/Annotation/AccessorOrder.php
@@ -28,18 +28,16 @@ final class AccessorOrder
 
     public function __construct(array $values = [], ?string $order = null, array $custom = [])
     {
-        if ([] !== $values) {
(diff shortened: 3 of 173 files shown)


joaojacome:
@goetas sorry for tagging you directly here, but would you be able to take a look at this? :) thanks!

src/Annotation/XmlDiscriminator.php
```diff
@@ -21,7 +22,32 @@ class XmlDiscriminator
...
     /**
-     * @var string
+     * @var string|null
      */
-    public $namespace;
+    public $namespace = null;
+
+    public function __construct(array $values = [], bool $attribute = false, bool $cdata = false, ?string $namespace = null)
+    {
+        if (0 !== count($values)) {
```
sanmai, line 31: Why not just `$values !== []` ?

src/XmlDeserializationVisitor.php
```diff
@@ -79,6 +79,8 @@ public function prepare($data)
 
         $previousEntityLoaderState = null;
         if (\LIBXML_VERSION < 20900) {
+            // @todo resolve the deprecation
```
sanmai, line 82: It is already resolved by the `if` above.

src/XmlDeserializationVisitor.php
```diff
@@ -97,6 +99,8 @@ public function prepare($data)
         libxml_use_internal_errors($previous);
 
         if (\LIBXML_VERSION < 20900) {
+            // @todo resolve the deprecation
```
sanmai, line 102: Likewise

doc/reference/annotations.rst
```diff
@@ -266,6 +313,22 @@ In this example:
...
+.. code-block :: php
+
+    /**
+     * @Serializer\VirtualProperty(
+     *     "firstName",
+     *     exp="object.getFirstName()",
+     *     options={@Serializer\SerializedName("my_first_name")}
+     *  )
+     */
+    #[Serializer\VirtualProperty(firstName: "firstName", exp: "object.getFirstName()", options: [[Serializer\SerializedName::class, ["my_first_name"]]])]
```
goetas, line 327: isn't here `name:"firstName"`?

goetas:
@joaojacome woha! this is huge and beautiful! thanks a lot for your work! i hope to find some time to test it locally on a project i have, but looks just great!

doc/reference/annotations.rst
joaojacome: yes :) now fixed

joaojacome: yes :) now fixed

src/Annotation/XmlDiscriminator.php
joaojacome: suggestion applied

src/XmlDeserializationVisitor.php
joaojacome: phpcs is complaining about the deprecated `libxml_disable_entity_loader` call. Should I just leave the `phpcs:ignore` there and remove the @todo line?

sanmai: It _is_ actually deprecated, but it is necessary for this library to run on distributions using older versions of libxml2. I think the only viable replacement here is to throw an exception, and I think this is best to propose in a dedicated PR.

src/Annotation/AccessorOrder.php
```diff
@@ -12,16 +12,37 @@
...
+    public $order = null;
 
     /**
      * @var array<string>
      */
     public $custom = [];
+
+    public function __construct(array $values = [], ?string $order = null, array $custom = [])
+    {
+        if ([] !== $values) {
```
goetas, line 31: are this `[] !== $values` checks really needed? `array_key_exists` should work well even if the array is empty

src/Annotation/AccessType.php
```diff
@@ -10,11 +10,27 @@
...
+            if (array_key_exists('value', $values)) {
+                $type = $values['value'];
+            }
+
+            if (array_key_exists('type', $values)) {
+                $type = $values['type'];
+            }
+        }
+
+        $this->type = $type;
```
niconoe-, line 34: 2 suggestions here:


### By keeping the local variable:

```suggestion
        if (array_key_exists('value', $values)) {
            $type = $values['value'];
        }
    
        if (array_key_exists('type', $values)) {
            $type = $values['type'];
        }

        $this->type = $type;
```

### Removing extra local variable:
```suggestion
        if (array_key_exists('value', $values)) {
            $this->type = $values['value'];
        }
    
        if (array_key_exists('type', $values)) {
            $this->type = $values['type'];
        }
```

I propose the 2 versions as I don't know your coding rules about it.

I could try to find some time to make a proposal that removes all `if ([] !== $values)` as they are not required, following your coding rules, if you want. Just tell me which proposal is OK to you.

Thanks

joaojacome: I'd go with the first suggestion, as the second one wouldn't work if attributes are disabled

niconoe-: Thank you. Do you want me to remove the unecessary `if`s or do you want to do it? I don't want to block you in your work here, creating conflicts for no reason.

joaojacome: I've applied the suggestion ;)

src/Annotation/AccessorOrder.php
joaojacome: I've just pushed a commit with this suggestion :man_dancing:

src/XmlDeserializationVisitor.php
joaojacome: Done! I've left only the phpcs ignore line :)

niconoe-:
@joaojacome Thanks a lot! I can't wait for it to be released :smiley:

src/XmlDeserializationVisitor.php
sanmai: Thanks!

joaojacome:
One thing that came to my mind now - with these changes you can use attributes or annotations, but not both.

Would that be the correct approach?

Maybe both could be supported (by getting rid of the "enableAttributes" flag), and then resolve the attributes, falling back to annotations.

What would ye prefer?

ruudk:
I think in large applications (like ours) it can be desirable to do the migration in steps (instead of a Big Bang).

niconoe-:
Agreed that step by step migration should be followed. This is not incompatible with the will of supporting both ways, neither incompatible with the actual PR IMO.
If I remember correctly, this package is currently in 4.1.x. With the support of the attributes (no matter using the `$enableAttributes` switch or using both ways) it could be a 4.2.x deprecating the usage of annotations, but still applicable.
Then, when preparing the 5.0, the annotations support could be removed in favor of attributes.

It’s basically how Symfony is doing this stuff, and I think it’s a good way to do a step-by-step migration.

My opinion about supporting the usage of both attributes and annotations at the same time without the `$enableAttributes` is that it could create conflicts when information are not the same in the 2 ways used. Secured approach is throwing dedicated exception when occurs, but it’s not errorless. User friendly approach is preferring a way over another, but this might be an unclear behavior.
I personally like the way it is introduced in the PR because the developer choose the method to use and it’s clear.

joaojacome:
> Or what if vendored bundles use annotations but you want to use attributes in `src`?

That's a good point imo. You'd need all your dependencies to support annotations before upgrading the serializer.

> Agreed that step by step migration should be followed. This is not incompatible with the will of supporting both ways, neither incompatible with the actual PR IMO.

Given the previously presented library argument, a project wouldn't be able to do a step by step migration.

> If I remember correctly, this package is currently in 4.1.x. With the support of the attributes (no matter using the `$enableAttributes` switch or using both ways) it could be a 4.2.x deprecating the usage of annotations, but still applicable. Then, when preparing the 5.0, the annotations support could be removed in favor of attributes. It’s basically how Symfony is doing this stuff, and I think it’s a good way to do a step-by-step migration.

I agree with having a major version for it, but I'd say it's too soon to think about that. For this attributes implementation, a few opinionated choices had to be made, specifically regarding the idea of having nested annotations/attributes. I would say it's better to wait for proper nested attributes support before removing any annotation code.

> My opinion about supporting the usage of both attributes and annotations at the same time without the `$enableAttributes` is that it could create conflicts when information are not the same in the 2 ways used.

Can you give us some examples? For me, being able to use both in a project would be better than moving everything to attributes in a single step. We'd just need to define the rules for when you can use one or another, or both.

niconoe-:
Ok, I also understand your point of view.

Examples of conflicts to me could be:


```php
/**
 * @Type("int")
 */
#[Attributes\Type("string")]
```

(sorry for the style, I’m on mobile and it’s complicated to format stuff here)

How does the property behave in this way? 

I agree you can define rules to solve this. My opinion is having such rules can be counter-intuitive by developers using this library. But I totally accept you disagree me on this 😃.

joaojacome:
For the initial attributes support, we could have something like a "metadata resolver priority", with values like "attributes first", "annotation first", "attributes only", "annotations only".

For a next major version, the annotations could be entirely dropped, but that might mean maintaining two versions for a good while, which is something the guys might not want to.

I'd like more inputs on this matter. @goetas

mbabker:
> Maybe both could be supported (by getting rid of the "enableAttributes" flag), and then resolve the attributes, falling back to annotations.

👍  for this, please.  Neither Doctrine (at least with the ORM's attributes) nor Symfony requires only annotations or only attributes for the full application, it'd be great if this library could follow along with this practice.

goetas:
I did some changes in https://github.com/schmittjoh/serializer/pull/1332 to this branch.
now is possible to use both doctrine and php8 attributes.

goetas:
@joaojacome could you please check my changes and let me know if they work as expected?

src/Annotation/VirtualProperty.php
```diff
@@ -12,24 +12,25 @@
  *
  * @author Alexander Klimenkov <alx.devel@gmail.com>
  */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 final class VirtualProperty
 {
     /**
-     * @var string
+     * @var string}null
```
niconoe-, line 19: I missed the typo here.

```suggestion
     * @var string|null
```

src/AttributeReader.php
```diff
@@ -0,0 +1,76 @@
...
+            $result[] = $attribute->newInstance();
+        }
+
+        return $result;
+    }
+
+    public function getClassAnnotation(ReflectionClass $class, $annotationName)
+    {
+        $attributes = $class->getAttributes($annotationName);
+        if (0 === count($attributes) || null === $attributes[0]) {
```
niconoe-, line 28: Same as https://github.com/schmittjoh/serializer/pull/1332#discussion_r679125075

Wouldn't it be better without calling `count`?

goetas:
done in https://github.com/schmittjoh/serializer/pull/1332

src/Annotation/Groups.php
```diff
@@ -8,8 +8,22 @@
...
+        if (array_key_exists('value', $values)) {
+            $groups = $values['value'];
+        }
+
+        if (array_key_exists('groups', $values)) {
+            $groups = $values['groups'];
+        }
+
+        $this->groups = $groups;
+    }
```
mareksuscak, line 28: @schmittjoh It seems that this change may have introduced a subtle breaking change. Upon upgrading `jms/serializer` from `3.13.0` to `3.14.0`, our app broke and we had to make the following update to get it to work again:

CleanShot 2021-10-08 at 15 45 48

This should likely be fixed and a patch version released or, at the very least, documented in the upgrade guide. Though I feel like the breaking change was not intended.

PR #1325, opened 2021-06-23
Context::getCurrentPath() returns indexes of list items

katin-dev:
Implementing `ExclusionStrategyInterface` i realized that there is no way to detect if current element is part on the list with specific index. 

For example, we have a class
```
class Node {
  public string $name;
  /** @var Node[] */
  public array $children;
}
```
When one of `$children` element has been passed to `shouldSkipProperty(PropertyMetadata $property, Context $context)`, i don't know which one. `Context` does not know the index of such element. But i have external rules like so which  define whether property of element should be serialized or not: 
```
$rules = [
  'children' => [
    0 => ['name'],
    2 => ['name', 'children']
  ]
]
```

I suggest `$context->getCurrentPath()` should return array like so: `[children][0]['name']` instead of `[children]['name']` (current behavior)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | likely so
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -253,6 +253,9 @@ abstract class Context
 
         $paths = [];
         foreach ($this->metadataStack as $metadata) {
+            if ($metadata instanceof ClassMetadata && $metadata->positionInList !== null) {
+                array_unshift($paths, $metadata->positionInList);
+            }
             if ($metadata instanceof PropertyMetadata) {
                 array_unshift($paths, $metadata->name);
             }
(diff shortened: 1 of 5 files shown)


katin-dev:
@goetas why Scrutinizer Canceled? What should I do to rerun it?

PR #1329, opened 2021-07-23
Fix phpstan return type

dgafka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT

### DIFF
--- a/src/SerializerInterface.php
+++ b/src/SerializerInterface.php
@@ -30,7 +30,7 @@ interface SerializerInterface
      * @throws RuntimeException
      *
      * @psalm-template T
-     * @psalm-param class-string<T> $type
+     * @psalm-param class-string<T>|array<class-string<T>> $type
      * @psalm-return T
      */
     public function deserialize(string $data, string $type, string $format, ?DeserializationContext $context = null);


dgafka:
@goetas can I kindly ask you for releasing this bug fix? Tests in my work currently failing due to this issue :smiley:

src/SerializerInterface.php
```diff
@@ -30,7 +30,7 @@ public function serialize($data, string $format, ?SerializationContext $context
      * @throws RuntimeException
      *
      * @psalm-template T
-     * @psalm-param class-string<T> $type
```
goetas, line 33: I think that this should be just removed. From my understanding phpstan does not support such notation.

if is not `class-string<T>` nor `array<class-string<T>`.

`$serializer->deserialize($data, 'xml', 'array<Whatever>')` can not be expressed with phpstan:

-  `class-string<T>` , that would be `$serializer->deserialize($data, 'xml', 'Whatever')` 
-  `array<class-string<T>>` , that would be `$serializer->deserialize($data, 'xml', ['Whatever>'])`

dgafka: Done :)

PR #1332, opened 2021-07-29
Add PHP attributes support

goetas:
continuation of https://github.com/schmittjoh/serializer/pull/1320

### DIFF
--- a/src/Metadata/Driver/AttributeDriver/AttributeReader.php
+++ b/src/Metadata/Driver/AttributeDriver/AttributeReader.php
@@ -13,54 +13,58 @@ class AttributeReader implements Reader
 {
     public function getClassAnnotations(ReflectionClass $class)
     {
-        $result = [];
         $attributes = $class->getAttributes();
-        foreach ($attributes as $attribute) {
-            if (0 === strpos($attribute->getName(), 'JMS\Serializer\Annotation\\')) {
-                $result[] = $attribute->newInstance();
-            }
-        }
 
-        return $result;
+        return $this->buildAnnotations($attributes);
     }
 
     public function getClassAnnotation(ReflectionClass $class, $annotationName)
     {
         $attributes = $class->getAttributes($annotationName);
-        if (0 === count($attributes) || null === $attributes[0]) {
-            return null;
-        }
 
-        return $attributes[0]->newInstance();
+        return $this->buildAnnotation($attributes);
     }
 
     public function getMethodAnnotations(ReflectionMethod $method)
     {
-        $result = [];
         $attributes = $method->getAttributes();
-        foreach ($attributes as $attribute) {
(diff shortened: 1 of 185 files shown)


niconoe- (commented):
If I understand the snippets correctly, when both annotations and attributes are set, priority is given to attributes, right?
This looks like perfectly fine to me, for what it matters.

tests/Fixtures/SimpleInternalObject.php
```diff
@@ -7,8 +7,9 @@
 use JMS\Serializer\Annotation as Serializer;
 
 /**
- * @Serializer\Exclude("NONE")
+ * @Serializer\ExclusionPolicy("ALL")
  */
+#[Serializer\ExclusionPolicy(policy: "ALL")]
```
niconoe-, line 12:
```suggestion
#[Serializer\ExclusionPolicy(policy: 'ALL')]
```

src/Metadata/Driver/AttributeDriver/AttributeReader.php
```diff
@@ -0,0 +1,82 @@
...
+declare(strict_types=1);
+
+namespace JMS\Serializer\Metadata\Driver\AttributeDriver;
+
+use Doctrine\Common\Annotations\Reader;
+use ReflectionClass;
+use ReflectionMethod;
+use ReflectionProperty;
+
+class AttributeReader implements Reader
```
niconoe-, line 12: I feel the 3 pairs of methods (for class, for method and for property) are doing all the same, don't they?

If this is due to copy/paste and not a will to let the possibility of each way evolve in their own ways, this could be easily reduced by 2 private (protected?) methods.

```php
    private function buildAnnotation($attributes)
    {
        if (!isset($attributes[0])) {
            return null;
        }
        return $attributes[0]->newInstance();
    }

    private function buildAnnotations($attributes)
    {
        $result = [];
        foreach ($attributes as $attribute) {
            if (0 === strpos($attribute->getName(), 'JMS\Serializer\Annotation\\')) {
                $result[] = $attribute->newInstance();
            }
        }
        
        return $result;
    }
```

And then:
```php
    public function getClassAnnotations(ReflectionClass $class)
    {
        $attributes = $class->getAttributes();
        return $this->buildAnnotations($attributes);
    }
    
    public function getClassAnnotation(ReflectionClass $class, $annotationName)
    {
        $attributes = $class->getAttributes($annotationName);
        return $this->buildAnnotation($attributes);
    }
    
    // Apply the same refactor for getMethodAnnotation(s) and getPropertyAnnotation(s)
```

Otherwise, if you want to keep the idea of making every method to evolve in their own ways independently, then you can keep it as is for now, but you should have 6 classes to split the responsibilities instead of a single one.

Issue #1331, opened 2021-07-25
`@ReadOnly` annotation incompatible with PHP 8.1


mbabker:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

The [readonly properties RFC](https://wiki.php.net/rfc/readonly_properties_v2) makes `readonly` a keyword in PHP 8.1.  This results in `ParseError: syntax error, unexpected token "readonly"` errors when parsing annotations thanks to the `JMS\Serializer\Annotation\ReadOnly` class.

goetas:
:(

the only suggestion I have is to rename it into "ReadOnlyProperty"...

PR #1333, opened 2021-07-31
Deprecate `@ReadOnly` annotation in favor of `@ReadOnlyProperty`

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #1331
| License       | MIT

With `readonly` becoming a keyword in PHP 8.1 thanks to the addition of [readonly properties](https://wiki.php.net/rfc/readonly_properties_v2), classes can no longer use it as a name.  This means the `@ReadOnly` annotation can't be used with PHP 8.1.  As suggested in the issue, this PR creates a new `@ReadOnlyProperty` annotation and deprecates the existing `@ReadOnly` annotation.

### DIFF
--- a/src/Annotation/ReadOnlyProperty.php
+++ b/src/Annotation/ReadOnlyProperty.php
@@ -8,8 +8,12 @@ namespace JMS\Serializer\Annotation;
  * @Annotation
  * @Target({"CLASS","PROPERTY"})
  *
- * @deprecated use `@ReadOnlyProperty` instead
+ * @final
  */
-final class ReadOnly extends ReadOnlyProperty
+/* final */ class ReadOnlyProperty
 {
+    /**
+     * @var bool
+     */
+    public $readOnly = true;
 }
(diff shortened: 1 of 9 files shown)


src/Annotation/ReadOnlyProperty.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace JMS\Serializer\Annotation;
+
+/**
+ * @final
+ *
+ * @Annotation
+ * @Target({"CLASS","PROPERTY"})
+ */
+/* final */ class ReadOnlyProperty
```
mbabker, line 13: This is purposefully left as not final and only annotated `@final` to help with B/C and code complexity.  Otherwise, there'd have to be a lot of PHP version conditionals and checks for both annotation classes.  In 4.0, when the deprecated `@ReadOnly` annotation is removed the final should be uncommented.

goetas: makes sense! i'm ok with it

PR #1335, opened 2021-08-02
Run PHP8 ci on lowest and highest deps

simPod:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "require-dev": {
         "ext-pdo_sqlite": "*",
         "doctrine/coding-standard": "^8.1",
-        "doctrine/orm": "~2.1",
+        "doctrine/orm": "~2.5",
         "doctrine/persistence": "^1.3.3|^2.0|^3.0",
         "doctrine/phpcr-odm": "^1.3|^2.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
     ],
     "require": {
         "php": "^7.2||^8.0",
-        "doctrine/annotations": "^1.0",
+        "doctrine/annotations": "^1.11",
```
goetas, line 20: why do we need this specific version?

simPod: Because any lower with "lowest" set of dependencies fails on PHP8. I've traced it to this though I did not find any particular change that causes it https://github.com/doctrine/annotations/compare/1.10.4...1.11.0. See failures here http://github.com/simPod/serializer/commit/20cf7008

simPod: Ok, might have been some issue in my local installation. Bumping orm is enough.

simPod:
would you mind keeping the co-contribution next time pls? thanks

PR #1337, opened 2021-08-03
Allow positional php8 attributes

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

### DIFF
--- a/src/Annotation/Groups.php
+++ b/src/Annotation/Groups.php
@@ -11,19 +11,25 @@ namespace JMS\Serializer\Annotation;
 #[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY)]
 final class Groups
 {
+    use AnnotationUtilsTrait;
+
     /** @var array<string> @Required */
     public $groups = [];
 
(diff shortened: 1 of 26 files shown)


goetas:
@joaojacome  i've implemented here the support for positional parameters for attributes, could you please have a look?

src/Annotation/Groups.php
```diff
@@ -11,19 +11,25 @@
...
     public function __construct(array $values = [], array $groups = [])
     {
-        if (array_key_exists('value', $values)) {
-            $groups = $values['value'];
-        }
-
-        if (array_key_exists('groups', $values)) {
-            $groups = $values['groups'];
+        $vars = get_defined_vars();
+        /*
```
joaojacome, line 22: TBH I don't like to have a different behaviour depending on the annotation, but I don't see any other way :)

goetas: yea...

Issue #1342, opened 2021-08-21
Serialization of DateTimeInterfaces


andrei-dascalu:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Steps required to reproduce the problem

sample code

```php

use JMS\Serializer\Annotation as JMS;
use Cake\Chronos\Chronos;
use DateTimeInterface;

class SerializeDate
{
    /**
     * @var DateTimeInterface
     * @JMS\Type("DateTimeInterface<'Y-m-d'>")
     * @JMS\SerializedName("geboortedatum")
    */
    public $dateTime;
}

$sendData = new SerializeDate();
$sendData->dateTime = Chronos::now();

/** config for annotations */
$serializer = JMS\Serializer\SerializerBuilder::create()->build();
$serializer->serialize($object, 'json');
```

## Expected Result

```json
{
	"geboortedatum": "2021-08-21"
}
```

## Actual Result

```json
{
	"geboortedatum": {
		"testNow": null,
		"toStringFormat": "Y-m-d H:i:s",
		"weekendDays": [6, 7],
		"diffFormatter": null,
		"_lastErrors": [],
		"days": {
			"1": "Monday",
			"2": "Tuesday",
			"3": "Wednesday",
			"4": "Thursday",
			"5": "Friday",
			"6": "Saturday",
			"7": "Sunday"
		},
		"weekStartsAt": 1,
		"weekEndsAt": 7,
		"relativePattern": "\/this|next|last|tomorrow|yesterday|midnight|today|[+-]|first|last|ago\/i"
	}
}
```

Chronos objects implement DateTimeInterface so I believe they should be serialised accordingly.

If I use a regular DateTime (that also implements DateTimeInterface), the result is as expected.

The behaviour is at the very least inconsistent:
- if Chronos is recognised as a DateTimeInterface (which it seems to be, since there's no type-related error), then it should be formatted accordingly
- if not, then there should be a clear type-related error, rather than silently serializing as an object when it's clearly against the desired configuration.

Cheers!

goetas:
which version of the serialization library are you using? can you please provide a failing test case?

andrei-dascalu:
Hi,

I'm using  php 8.0.9 with JMS 3.14.0 & Chronos 2.2.0

Simple code to reproduce:

```php
<?php

use JMS\Serializer\Annotation as JMS;
use Cake\Chronos\Chronos;
[...]

PR #1344, opened 2021-09-08
Proposed fix for serializing custom DateTimeInterface implementations

andrei-dascalu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | yes/no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1342 
| License       | MIT

Proposed fix for  #1342 

Consideration: there's nothing stopping a custom DateTimeInterface implementation from, say, adding extra fields. This MR does not address that, instead treats the implementation as having the same restricted responsibility as the default implementations (eg: DateTime / DateTimeImmutable) since the ticket was aimed at issues with the two main custom date time libraries (Carbon and Chronos)


goetas:
thank you

PR #1345, opened 2021-09-16
Symfony 6 and DBAL 3 compat

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT

This PR allows for Symfony 6 to be installed in dev and fixes an issue in the tests when DBAL 3 is installed.

For Symfony 6, after forcing dev dependencies to be installed, I had zero test failures.

Forcing dev dependencies to be installed brought in ORM 2.10 which supports DBAL 3 now.  DBAL 3 drops support for the UUID generator, which was used in a couple of the test fixtures.  To fix compatibility and not lose test coverage, I changed the ID columns to the "auto" strategy then added a new `$guid` property to the `BlogPost` fixture as a guid column and updated the test checking the handling of guid columns to use that.


goetas:
thanks!

PR #1354, opened 2021-10-16
Add support for `list<T>`

simPod:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT

- [ ] not sure I've tested that on proper places

### DIFF
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -358,9 +358,15 @@ Available Types:
 +------------------------------------------------------------+--------------------------------------------------+
 | array                                                      | An array with arbitrary keys, and values.        |
 +------------------------------------------------------------+--------------------------------------------------+
-| array<T>                                                   | A list of type T (T can be any available type).  |
+| list                                                       | A list with arbitrary values.                    |
++------------------------------------------------------------+--------------------------------------------------+
+| array<T>                                                   | An array of type T (T can be any available type).|
 |                                                            | Examples:                                        |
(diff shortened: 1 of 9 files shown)


simPod:
List implies numeric and sequenced keys starting from 0. Array<T> has whatever keys of array-key. See https://php.watch/versions/8.1/array_is_list for explanation e. g.

goetas:
in jms `array<T>`  is exactly a list. what is not a list has to be expressed as `array<string,T>`

simPod:
> in jms array<T> is exactly a list

In that case this should be deprecated as it does not comply with current typesystem adopted by ecosystem (phpstan, psalm, IDEs, ...)

-------------

This PR adds support for `@param list<int>` so no additional jms serializer annotations are required for that. I'll add a fixture with that.

dborsatto:
Hey @goetas I've got the same issue with the serializer, the `list<T>` annotation has become the standard for this kind of meaning, so annotating with `array<T>` to make the serializer work breaks other tools as technically `array<T>` is only a parent type of `list<T>` (the Psalm [docs](https://psalm.dev/docs/annotating_code/type_syntax/array_types/) explain this quite nicely).

Is there any way I can help with this PR? :slightly_smiling_face:

goetas:
ok, since this seems on demand, let put it in core

PR #1355, opened 2021-10-16
Fix ObjectWithInlineArray fixture location

simPod:


goetas:
is this file used at all? all tests were green even with it in the wrong location...

simPod:
Seems like it

https://github.com/schmittjoh/serializer/blob/0c83e1f7097a8c88419060591329eba89ede9ee6/tests/Serializer/JsonSerializationTest.php#L321

simPod:
cc @goetas

Issue #1211, opened 2020-06-06
SerializerInterface::deserialize has incorrect psalm annotation


ostrolucky:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

https://github.com/schmittjoh/serializer/blob/5b682e96def251472a0806455275c65521319e22/src/SerializerInterface.php#L26-L28 is incomplete, if nothing else. We use following code:
```php
public function deserializeSnapshot(string $json): array
    {
        return $this->serializer->deserialize($json, 'array', 'json');
    }
```

Annotation used on your interface causes following errors:
```
ERROR: InvalidReturnType - src/AppBundle/Snapshot/Common/Snapshotter.php:50:16 - The declared return type 'array<array-key, mixed>' for AppBundle\Snapshot\Common\Snapshotter::deserializeSnapshot is incorrect, got 'object' (see https://psalm.dev/011)
     * @return mixed[]


ERROR: InvalidReturnStatement - src/AppBundle/Snapshot/Common/Snapshotter.php:54:16 - The inferred type 'object' does not match the declared return type 'array<array-key, mixed>' for AppBundle\Snapshot\Common\Snapshotter::deserializeSnapshot (see https://psalm.dev/128)
        return $this->serializer->deserialize($json, 'array', 'json');


ERROR: ArgumentTypeCoercion - src/AppBundle/Snapshot/Common/Snapshotter.php:54:54 - Argument 2 of JMS\Serializer\Serializer::deserialize expects class-string, parent type string(array) provided (see https://psalm.dev/193)
        return $this->serializer->deserialize($json, 'array', 'json');


ERROR: UndefinedClass - src/AppBundle/Snapshot/Common/Snapshotter.php:54:54 - Class or interface array does not exist (see https://psalm.dev/019)
        return $this->serializer->deserialize($json, 'array', 'json');

```

goetas:
Hi, thanks for the report. Do you know which annotation is the correct one?

ostrolucky:
I don't, unfortunately

ostrolucky:
So, something like this https://psalm.dev/r/84eb3d264b (You should know better what other types can go there and should be able to extend this further)

EvgeniiR:
Just to clarify, with code example above, there'll still be problem with `array<class-string>`, e.g. `array<App\Foo>`

goetas:
[...]

PR #1356, opened 2021-10-25
Psalm annotations for "deserialize" method

EvgeniiR:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1211
| License       | MIT

With this PR psalm will be able to infer returning type for `SerializerInterface::deserialize()` method, when deserializing one instance of a class, e.g. `$s->deserialize($json, A::class);`.

Other variants unfortunately still be `mixed`, so it should cause no BC breaks, but make life a bit easier

Example snippet: https://psalm.dev/r/5e0bce36c4

### DIFF
--- a/src/SerializerInterface.php
+++ b/src/SerializerInterface.php
@@ -28,6 +28,10 @@ interface SerializerInterface
      * @return mixed
      *
      * @throws RuntimeException
+     *
+     * @psalm-template T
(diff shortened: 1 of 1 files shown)


goetas:
see https://github.com/schmittjoh/serializer/issues/1211#issuecomment-998152022

Issue #1312, opened 2021-04-13
Exclude Strategy with Until not working exepcted with `.x`


alexander-schranz:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Steps required to reproduce the problem

We are using the versioning of the jms serializer.

I first did have this:

```
<property name="sections" expose="true" since-version="3.0"/>
<virtual-property name="areas_without_section" serialized-name="areas" method="getAreasWithoutSections" expose="true" since-version="1.0" until-version="2.0" />
<virtual-property name="areas" serialized-name="areas" method="getAreas" expose="true" since-version="2.0" until-version="3.0" />
```

Then I did find out that the `areas` is still returned when using `3.0` version. So I did see in the docs: https://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies

There is something like `1.0.x` used so I did use:

```xml
<property name="sections" expose="true" since-version="3.0"/>
<virtual-property name="areas_without_section" serialized-name="areas" method="getAreasWithoutSections" expose="true" since-version="1.0" until-version="1.x" />
<virtual-property name="areas" serialized-name="areas" method="getAreas" expose="true" since-version="2.0" until-version="2.x" />
```

but this didn't work the `areas` where not returned when requested with `2.0`

## Expected Result

<details>
<summary>**3.0**: Return `sections` without `areas`: </summary>

```js
            "sections": [
                {
                    "_hash": "@string@",
                    "uuid": "@string@",
                    "name": "Section 1",
                    "areas": [
                        {
                            "_hash": "@string@",
                            "uuid": "@string@",
                            "name": "1"
                        },
                        {
                            "_hash": "@string@",
                            "uuid": "@string@",
                            "name": "1A"
                        },
                        {
                            "_hash": "@string@",
                            "uuid": "@string@",
                            "name": "2A"
                        },
                        {
                            "_hash": "@string@",
                            "uuid": "@string@",
                            "name": "2B"
                        }
                    ]
                },
                {
                    "_hash": "@string@",
                    "uuid": "@string@",
                    "name": "Section 2",
                    "areas": [
                        {
                            "_hash": "@string@",
                            "uuid": "@string@",
                            "name": "1"
                        },
                        {
                            "_hash": "@string@",
                            "uuid": "@string@",
                            "name": "1A"
                        },
                        {
                            "_hash": "@string@",
                            "uuid": "@string@",
[...]

PR #1361, opened 2021-10-31
Exclude properties based on semver constraints

W0rma:
…ver constraints

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1312 
| License       | MIT

This PR adds the possibility to exclude properties from serialization based on version constraints accepted by [composer](https://getcomposer.org/doc/articles/versions.md#writing-version-constraints).

Example:

```php
use JMS\Serializer\Annotation\Semver;

class Author
{
    #[Semver(version: '^6.1')]
    public function getAge(): int
    {
        return 61;
    }
}
```


composer.json
```diff
@@ -30,6 +30,7 @@
     },
     "require-dev": {
         "ext-pdo_sqlite": "*",
+        "composer/semver": "^1.7.4 || ^2 || ^3.2",
```
scyzoryck, line 33: Are there any reason why 3.1 would not work? :) 
I would propose to define it similar to other deps: `^1.7.4|^2.0|^3.0`. May it would be nice to add into `suggest` section too?

src/Annotation/VersionConstraints.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Annotation;
+
+use Composer\Semver\Semver as ComposerSemver;
```
scyzoryck, line 7: I would avoid using aliases for namespaces as long as it is really needed to keep code simpler and more readable.

scyzoryck (commented):
Thanks @W0rma for this PR! I left two minor comments.
Do you think that we should keep `Until` and `Since` attributes or they should be deprecated?

src/Annotation/VersionConstraints.php
W0rma: Thank you for the hint. This was a leftover since I used a different name for the annotation before.

composer.json
W0rma: You are right - I fixed the version constraints and added the package to the `suggest` section.

W0rma:
> Do you think that we should keep `Until` and `Since` attributes or they should be deprecated?

@scyzoryck Good point. I think they should be deprecated since version constraints provide even more flexibility. Do you agree?

scyzoryck:
Thanks. Looks good to me. 

> I think they should be deprecated since version constraints provide even more flexibility. Do you agree?

It was my initial thought - as basically new constraint covers all functionalities provided by older constraints. On the other hand new one requires external dependency, so keeping simpler constraints might made sense too. @goetas - what do you think about it?

doc/cookbook/exclusion_strategies.rst
```diff
@@ -57,20 +57,21 @@ expose them via an API that is consumed by a third-party:
     class VersionedObject
     {
         /**
-         * @Until("1.0.x")
```
goetas, line 60: I would not abandon `@Until` and `@Since`, mainly for performance reasons. my guess is that `composer/semver` is way slower that the php's `version_compare` function, in a large object graph in it can make the difference.

BTW, i would be curious to see a benchmark about it

src/Annotation/VersionConstraints.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * @Annotation
+ * @Target({"PROPERTY", "METHOD", "ANNOTATION"})
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY)]
+final class VersionConstraints extends Version
+{
+    public function __construct($values = [], ?string $version = null)
+    {
+        if (!class_exists(Semver::class)) {
```
goetas, line 18: what if this check is done in the metadata drivers? so we avoid to call it at runtime each time the annotation is instantiated?

Issue #1367, opened 2021-11-24
Typed property X must not be accessed before initialization


tmarsteel:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

This topic has come up in #1336 before, but in the context of "this is an issue with 3.14.0-rc1", not "please make this work".
So here goes: please make this work. I would really like to have the strong safety of typed properties.

## Steps required to reproduce the problem
Afaict, this affects all versions under PHP 7.4.

```php
class Foo {
  public $pre74property;
  public string $initialized74property = 'present';
  public ?string $uninitialized74property;
}

$serializer = (new JMS\Serializer\SerializerBuilder())->build();
$context = new JMS\Serializer\SerializationContext();
$context->setSerializeNull(true);

echo $serializer->serialize(new Foo(), 'json', $context);
```

## Expected Result

```json
{
  "pre74property": null,
  "initialized74property": "present",
  "uninitialized74property": null
}
``` 

or, with `setSerializeNull(false)`:

```json
{
  "initialized74property": "present"
}
``` 

## Actual Result

```
PHP Fatal error:  Uncaught Error: Typed property Foo::$uninitialized74property must not be accessed before initialization in /home/tmarstaller/Desktop/jmstest/vendor/jms/serializer/src/Accessor/DefaultAccessorStrategy.php:89
Stack trace:
#0 /home/tmarstaller/Desktop/jmstest/vendor/jms/serializer/src/Accessor/DefaultAccessorStrategy.php(94): Foo::JMS\Serializer\Accessor\{closure}()
#1 /home/tmarstaller/Desktop/jmstest/vendor/jms/serializer/src/GraphNavigator/SerializationGraphNavigator.php(258): JMS\Serializer\Accessor\DefaultAccessorStrategy->getValue()
#2 /home/tmarstaller/Desktop/jmstest/vendor/jms/serializer/src/Serializer.php(252): JMS\Serializer\GraphNavigator\SerializationGraphNavigator->accept()
#3 /home/tmarstaller/Desktop/jmstest/vendor/jms/serializer/src/Serializer.php(163): JMS\Serializer\Serializer->visit()
#4 /home/tmarstaller/Desktop/jmstest/test.php(11): JMS\Serializer\Serializer->serialize()
#5 {main}
[...]

PR #1375, opened 2021-12-06
add check for uninitialized property, and force reflection if type is…

re2bit:
Draft for Discussion of #1367

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no 
| Tests pass?   | ?
| Fixed tickets | #1367
| License       | MIT

### DIFF
--- a/src/Accessor/DefaultAccessorStrategy.php
+++ b/src/Accessor/DefaultAccessorStrategy.php
@@ -72,7 +72,7 @@ final class DefaultAccessorStrategy implements AccessorStrategyInterface
             return $object->{$metadata->getter}();
         }
 
-        if ($metadata->forceReflectionAccess) {
+        if ($metadata->forceReflectionAccess || ($context->getShouldSerializeUnitializedAsNull() && $metadata->hasType)) {
             $ref = $this->propertyReflectionCache[$metadata->class][$metadata->name] ?? null;
             if (null === $ref) {
                 $ref = new \ReflectionProperty($metadata->class, $metadata->name);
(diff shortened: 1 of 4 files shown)


tmarsteel:
Awesome, this looks good :+1: I suspected it would have been more work.
Thanks :)

src/Accessor/DefaultAccessorStrategy.php
```diff
@@ -72,14 +72,18 @@ public function getValue(object $object, PropertyMetadata $metadata, Serializati
...
-        if ($metadata->forceReflectionAccess) {
+        if ($metadata->forceReflectionAccess || ($context->getShouldSerializeUnitializedAsNull() && $metadata->hasType)) {
             $ref = $this->propertyReflectionCache[$metadata->class][$metadata->name] ?? null;
             if (null === $ref) {
                 $ref = new \ReflectionProperty($metadata->class, $metadata->name);
                 $ref->setAccessible(true);
                 $this->propertyReflectionCache[$metadata->class][$metadata->name] = $ref;
             }
 
+            if (PHP_VERSION_ID >= 70400 && !$ref->isInitialized($object)) {
```
scyzoryck, line 83: Should it be executed also for the cases when `$metadata->forceReflectionAccess` is set to true, but `$context->getShouldSerializeUnitializedAsNull()` will return false?

scyzoryck (commented):
Hello @re2bit!
Thank you for your work! :) It looks good, I left few comments. 

In general as forcing Reflection Access might slow down application I would try to limit possible cases when possible - I found that with PHP8.0 has also `ReflectionProperty::hasDefaultValue()` method that might help to avoid forcing usage of reflection to minimum - what do you think about it? 

Best!

Tobion:
Introducing `setShouldSerializeUnitializedAsNull` for this is not good IMO. It makes it impossible to serialize null but not serialize unitialized properties. This is something people will need as that is the whole point of differentiating between unitialized and null.

Issue #1373, opened 2021-12-01
8.1 backed enum functionally not correctly (de)serialize


thepercival:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | yes

## Steps required to reproduce the problem

1. create a backed enum in php 8.1
 enum AgainstRuleSet: int
{
    case DiffFirst = 1;
    case AmongFirst = 2;
}
2. add the enum as a property to a yaml file
MyDomain\MyClass:
  properties:
    pointsCalculation:
      type: MyDomain\AgainstRuleSet
3. serialize 

## Expected Result
The integer or string should be returned. In this case: 1 

## Actual Result
Instead {name: "DiffFirst", value: 1} is returned. Before enums the property-type would be int or string. With enums this is now an enum-type. For a backed enum this should automatically be casted to string or int with the designated functions(which are in the interface). in core.php
For deserialization the opposite should happen.

scyzoryck:
Hello @thepercival ! 

I think that your proposition makes sense and should be an expected once. Current behaviour might be hard to use with real applications :D 
Would you like to create an PR with its implementation? :)

thepercival:
I could give it a try. My proposal would be to specify the enum as a type.
Than check at (de)serialization if the type inherits from interface BackedEnum. 
If so, a string or int should be returned(internal interfaces IntBackedEnum and StringBackedEnum).
This would be myEnum->value
Would this by a correct way to implement?

scyzoryck:
I believe that there are more then one correct way to implement it :) 
I would try to avoid adding new type like `enum< MyDomain\AgainstRuleSet>` - if possible I would try to autodetect it on the Metadata level to improve developer experience. It might introduce a small backward incompatibility, but as PHP8.1 is pretty fresh I should not be a big issue. @goetas WDYT?  
Some crazy idea - in theory it would be possible to handle enums by modifying metadata for classes that implements BackedEnum - by using  `inline`, `getter`, `setter` options. It would require only changes in metadata parsing without a need to modify Serialiser class. But it might be a kind of magic :D

thepercival:
should we wait for @goetas to give his opinion?

scyzoryck:
I would go with PR :)

thepercival:
ok, next week, maybe friday i will make a start.I am not a tech-expert, but have enough php-experience. A did a few pull request before.

Einenlum:
[...]

PR #1384, opened 2021-12-28
added backed enum support

taranovegor:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes(?)
| Deprecations? | no
| Tests pass?   | TODO
| Fixed tickets | #1373
| License       | MIT

This is my proposal for a serialization implementation for sets. I'm ready to listen to suggestions and make changes, It may not be perfect, but it works. :)

Backward compatibility will be broken as the structure of the serialized enum will be changed from `{"enum":{"name":"NAME", "value":"VALUE"}}` to `{"enum":"VALUE"}` for json, and for xml `<enum><name><![CDATA[NAME]]></name><value><![CDATA[VALUE]]></value></enum>`.

Screenshot 2021-12-28 at 23 11 01

### DIFF
--- a/src/GraphNavigator/DeserializationGraphNavigator.php
+++ b/src/GraphNavigator/DeserializationGraphNavigator.php
@@ -146,6 +146,10 @@ final class DeserializationGraphNavigator extends GraphNavigator implements Grap
             default:
                 $this->context->increaseDepth();
 
+                if (function_exists('enum_exists') && enum_exists($type['name'], false)) {
+                    $type = ['name' => \BackedEnum::class, 'enum' => $type['name'], 'params' => $type['params'] ?? []];
+                }
+
                 // Trigger pre-serialization callbacks, and listeners if they exist.
                 // Dispatch pre-serialization event before handling data to have ability change type in listener
                 if ($this->dispatcher->hasListeners('serializer.pre_deserialize', $type['name'], $this->format)) {
--- a/src/GraphNavigator/SerializationGraphNavigator.php
+++ b/src/GraphNavigator/SerializationGraphNavigator.php
@@ -186,8 +186,12 @@ final class SerializationGraphNavigator extends GraphNavigator
                 // If we're serializing a polymorphic type, then we'll be interested in the
                 // metadata for the actual type of the object, not the base class.
                 if (class_exists($type['name'], false) || interface_exists($type['name'], false)) {
-                    if (is_subclass_of($data, $type['name'], false) && null === $this->handlerRegistry->getHandler(GraphNavigatorInterface::DIRECTION_SERIALIZATION, $type['name'], $this->format)) {
-                        $type = ['name' => \get_class($data), 'params' => $type['params'] ?? []];
+                    if (null === $this->handlerRegistry->getHandler(GraphNavigatorInterface::DIRECTION_SERIALIZATION, $type['name'], $this->format)) {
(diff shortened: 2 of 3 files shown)


Einenlum:
I won't open a new PR to avoid duplicate but [here](https://github.com/schmittjoh/serializer/compare/master...Einenlum:feature/support-enums?expand=1) is my personal work on this, with tests. Also shoulndn't you add the BackedEnumHandler in the `SerializerBuilder`?

taranovegor:
> I won't open a new PR to avoid duplicate but here is my personal work on this, with tests:
> 
> https://github.com/schmittjoh/serializer/compare/master...Einenlum:feature/support-enums?expand=1
> 
> Maybe it can help.

Hello

I looked at your solution and it looks good :) But, you are also suggesting to drop support for php version below 8.1, I suggest making the following [changes](https://github.com/schmittjoh/serializer/compare/master...taranovegor:support_for_older_versions?expand=1). In the solution, I did not specifically check for the existence of the `\BackedEnum` interface, relying on the php version to avoid the possibility of interfaces/classes with the same name in the same namespace.

> Also shouldn't you add the BackedEnumHandler in the `SerializerBuilder`?

Yes, this is indeed a mistake on my part.

P.S. I can close this PR if you create your own.

charliefancelli:
any news to merge this branch ?

taranovegor:
> any news to merge this branch ?

Hello, this implementation has a small issue that has been noticed [there](https://github.com/schmittjoh/serializer/issues/1373). If you have any suggestions for a solution, please share them.

goetas:
I'm sorry if it took me so long to express my opinion on this.

I see that the enum support is something needed and there are already some PR with possible solutions.

Most of the implementations assume that a baked enum serializes and deserializes their value, and a ordinary enum the "key" if it.
I'm currently working on a project in which a baked enum gets serialized and deserialized using the "key" despite having the value (baked enum), to make it even more fun we transform the "key" in lowercase...

Thus I think that any suggested implementation should be flexible enough to offer a sane default (serialize/deserialize baked enums using their values and non-baked enums" using their "key") and at the same time allow the user to apply custom handlers on some specific enums.

Einenlum:
I'm not sure it would make sense to serialize/deserialize a non backed enum. Since it has no value it's not supposed to be stored or serialized.
I think there could be an option to use the key as value but the default behavior should be by value.

the-toster:
> I'm not sure it would make sense to serialize/deserialize a non backed enum
It makes very much sense for me, It allows me to use enums in my domain objects, and let me deserialize well typed object.

Einenlum:
> > I'm not sure it would make sense to serialize/deserialize a non backed enum
> 
> It makes very much sense for me, It allows me to use enums in my domain objects, and let me deserialize well typed object.

Real question: why not use backed enum then? Seems like that's what they're for.

the-toster:
> Real question: why not use backed enum then? Seems like that's what they're for.

In my model, I need only distinct one thing from other, so case name is enough. Why should I add redundant value for it?

taranovegor:
Guys, at the moment I can not implement this task.

Einenlum:
Hi @taranovegor, thank you very much for your message and update. No worries, doing Open Source takes time and energy and we are a lot here missing free time to handle these tasks.

PR #1387, opened 2022-01-12
ArrayTransformerInterface::fromArray() DocBlock typehint to allow psalm infer correct return type

the-toster:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | Under some circumstances, It may cause psalm to emit `UnnecessaryVarAnnotation` message. But psalm itself released `UnnecessaryVarAnnotation` as a [patch-level change](https://github.com/vimeo/psalm/releases/tag/3.5.2)
| Deprecations? | no 
| Tests pass?   | yes
| License       | MIT

This PR allows to simplify this type of code:
```php
/** @var Obj $r */
$r = $serializer->fromArray($arr, Obj::class);
return $r;
```
to
```php
return $serializer->fromArray($arr, Obj::class);
```
I don't know how to test it automatically, but hope it can be merged without auto-tests.

Thank you for this great library.

### DIFF
--- a/src/ArrayTransformerInterface.php
+++ b/src/ArrayTransformerInterface.php
@@ -26,8 +26,11 @@ interface ArrayTransformerInterface
      * Restores objects from an array structure.
      *
      * @param array $data
+     * @param class-string<T> $type
      *
-     * @return mixed this returns whatever the passed type is, typically an object or an array of objects
+     * @return T this returns whatever the passed type is, typically an object or an array of objects
+     *
+     * @template T
      */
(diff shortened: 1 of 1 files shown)


goetas:
There is no good way to make it work. it has been discussed in https://github.com/schmittjoh/serializer/issues/1211

if you are using phpstan, you can use https://github.com/goetas/jms-serializer-phpstan-extension regarding psaml i'm not aware of some plugin system to make it work

the-toster:
Thank you for feedback, it's good idea to make psalm plugin for this.

Issue #1381, opened 2021-12-16
Version 3.11+ is not compatible with PHP 8.1


dkreuer:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Steps required to reproduce the problem

1. Use PHP 8.1
2. Create project with composer
3. Require jms/serializer
4. Run composer-require-checker

## Expected Result

* No exception

## Actual Result

```
In LocateASTFromFiles.php line 46:
                                                                                                                                                                          
  Parsing the file [/var/www/vendor/jms/serializer/src/Annotation/ReadOnly.php] resulted in an error: Syntax error, unexpected T_READONLY, expecting T_STRING on line 14  
                                                                                                                                                                          

In ParserAbstract.php line 318:
                                                                      
  Syntax error, unexpected T_READONLY, expecting T_STRING on line 14  
                                                                      
```

## Underlying issue:

"readonly" is a reserved keyword since PHP 8.1. Albeit the Annotation class `JMS\Serializer\Annotation\ReadOnly` is deprecated since version 3.14.0 code analysis running under PHP 8.1 fails if it hits the file. Code might actually break if the annotation is used in an PHP 8.1 project/compatible library.

## Conclusion:

The library is not fully compatible with PHP 8.1 and this is a problem which cannot be fixed easily due to the open version range constraint (`"php": "^7.2 || ^8.0"`). I suggest to release bugfix versions for all maintained version branches which limits the constraint to `"php": "7.2.* || 7.3.* || 7.4.* || 8.0.*"` and a new major version removing the `ReadOnly` annotation and with constraint `"php": "7.2.* || 7.3.* || 7.4.* || 8.0.* || 8.1.*"`. This should help to at least reduce the problem surface.

mbabker:
I'm mildly confused here.  Because the package still provides the deprecated "ReadOnly" class and didn't rush to release a major version to drop that single class (which would be more disruptive than leaving it in place and telling folks to not use it), that must mean the package cannot be used on PHP 8.1 at all?

IMO, this is a case better suited for SA tools being configured to ignore the problem class in incompatible environments instead of trying to say the entire package is unusable because of one optional code path that can lead to a fatal error thanks to a PHP B/C break.

dkreuer:
> thanks to a PHP B/C break

[...]

PR #1390, opened 2022-02-01
Remove ReadOnly since jms serializer cannot be used with PHP 8.1 projects with it

simPod:
I'm performing full codebase scan and php8.1 cannot parse it now with this class being present

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | yes (already specified in UPGRADING)
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT


simPod:
See https://github.com/schmittjoh/serializer/runs/5023529751?check_suite_focus=true for failure

goetas:
why that ? i'm using jms with php 8.1 and symfony 5.4. what is blocking you from using it on php 8.1? do you get some error?

simPod:
Yes, that's because you're not scanning your code base entirely, thus not hitting read-only class. Simply try to load Readonly.php on Php81. The problem is sole presence of the class there.

simPod:
E.g. this https://github.com/maglnet/ComposerRequireChecker <https://github.com/maglnet/ComposerRequireChecker> or whatever else that does the same kind of actions. The class must not exist in PHP 8.1 projects.

mbabker:
This is the same discussion as https://github.com/schmittjoh/serializer/issues/1381

simPod:
Closing in favor of https://github.com/schmittjoh/serializer/pull/1405

Issue #1392, opened 2022-02-08
Support of enum members


the-toster:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

As I understand, enum members aren't supported now. I can try to add support of deserialization, if it's relevant.

## Steps required to reproduce the problem
I added test that demonstrates current state - https://github.com/the-toster/serializer/commit/b0b07bf273c45f26b4a2c2617e78e7136140b149

## Expected Result
Serialize / deserialize objects with enum members.

## Actual Result
`Cannot instantiate enum`

the-toster:
Seems it [doctrine/instantiator](https://github.com/doctrine/instantiator/blob/1.4.x/src/Doctrine/Instantiator/Instantiator.php) problem?

scyzoryck:
Hi! @the-toster 
The topic seems to be related to backend enums that are discussed here: https://github.com/schmittjoh/serializer/issues/1373

Have you thought about the format for serialised Enum value? 
Current serialised value to array is:
```
[
    'enum' =>
        [
            'name' => 'a',
        ],
]
```

Best, scyzoryck.

the-toster:
@scyzoryck yes, I see this thread. For me, current serialized format is fine, and it allows to deserialize it properly.

the-toster:
I checked how `ObjectConstructorInterface` implementation selected, and as I understand, correct way is not add new one, but tweak `UnserializeObjectConstructor::construct()` method:

https://github.com/schmittjoh/serializer/blob/3f8750a21a4f3dc12cf8ac2edd9e39f611667a16/src/Construction/UnserializeObjectConstructor.php#L20-L23

I think about adding `ClassMetadata::isEnum()` method, and then use it here choose way instantiate value.

the-toster:
I can add deserialization by value, but currently both pure and backed serialized by key, which is right in my opinion, I reason this way:
1) purpose of serialization - subsequent deserialization
2) complete deserialization possible only in same environment as serialization (same classes should be declared, etc)
3) so, enum key is enough and preferable (because it obey PHP identifier syntax, while value can be much complex and large string)

Ofcourse I can understand use cases when serialized object used in other environment, sending to JS frontend for example, this case can be handled with custom handlers.

goetas:
Done in https://github.com/schmittjoh/serializer/pull/1448

PR #1393, opened 2022-02-09
Enum deserialization support

the-toster:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1392 
| License       | MIT

Please check it, hope this can be accepted. I used undocumented `ReflectionClass::isEnum()` but I hope it will be added to documentation soon.

Also, codesniffer is not support enum syntax, so it report couple errors about fixture enum.

### DIFF
--- a/src/Construction/UnserializeObjectConstructor.php
+++ b/src/Construction/UnserializeObjectConstructor.php
@@ -19,6 +19,10 @@ final class UnserializeObjectConstructor implements ObjectConstructorInterface
      */
     public function construct(DeserializationVisitorInterface $visitor, ClassMetadata $metadata, $data, array $type, DeserializationContext $context): ?object
     {
+        if ($metadata->isEnum) {
+            return (new \ReflectionEnum($type['name']))->getCase($data['name'])->getValue();
+        }
+
         return $this->getInstantiator()->instantiate($metadata->name);
     }
 
--- a/src/Metadata/ClassMetadata.php
+++ b/src/Metadata/ClassMetadata.php
@@ -85,6 +85,13 @@ class ClassMetadata extends MergeableClassMetadata
     public $isMap = false;
 
     /**
+     * @internal
+     *
+     * @var bool
+     */
+    public $isEnum = false;
+
+    /**
      * @var bool
      */
     public $discriminatorDisabled = false;
--- a/src/Metadata/ClassMetadata.php
+++ b/src/Metadata/ClassMetadata.php
@@ -231,6 +231,7 @@ class ClassMetadata extends MergeableClassMetadata
 
         $this->isMap = $object->isMap;
         $this->isList = $object->isList;
+        $this->isEnum = $object->isEnum;
 
         $this->xmlNamespaces = array_merge($this->xmlNamespaces, $object->xmlNamespaces);
 
@@ -319,6 +320,7 @@ class ClassMetadata extends MergeableClassMetadata
             $this->xmlRootPrefix,
             $this->isList,
             $this->isMap,
+            $this->isEnum,
             parent::serializeToArray(),
         ];
     }
@@ -349,6 +351,7 @@ class ClassMetadata extends MergeableClassMetadata
             $this->xmlRootPrefix,
             $this->isList,
             $this->isMap,
+            $this->isEnum,
             $parentData,
         ] = $data;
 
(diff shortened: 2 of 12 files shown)


the-toster:
`ReflectionClass::isEnum()` was added to https://github.com/php/doc-en

simPod (approved):

src/Construction/UnserializeObjectConstructor.php
```diff
@@ -19,6 +19,10 @@ final class UnserializeObjectConstructor implements ObjectConstructorInterface
      */
     public function construct(DeserializationVisitorInterface $visitor, ClassMetadata $metadata, $data, array $type, DeserializationContext $context): ?object
     {
+        if ($metadata->isEnum) {
+            return (new \ReflectionEnum($type['name']))->getCase($data['name'])->getValue();
```
simPod, line 23: Can you add a backed enum test case?

My code requires this change:

```suggestion
            return (new \ReflectionEnum($type['name']))->getCase($data)->getValue();
```

(for

```php
enum FlopNetworkEventType: int
{
    case UPDATE = 1;
    case DELETE = 2;
}
```
)

simPod (changes requested):

src/Metadata/ClassMetadata.php
```diff
@@ -84,6 +84,13 @@ class ClassMetadata extends MergeableClassMetadata
      */
     public $isMap = false;
 
+    /**
+     * @internal
+     *
+     * @var bool
+     */
+    public $isEnum = false;
```
simPod, line 92: see e.g. `isMap` in this class. It needs to implement the same behaviour for isEnum. Now the caching does not work

simPod (changes requested):

simPod:
This should do https://github.com/the-toster/serializer/pull/1

goetas:
some feedback in https://github.com/schmittjoh/serializer/pull/1384#issuecomment-1123345019

the-toster:
I can add deserialization by value, but currently both pure and backed serialized by key (at least according to added test), which is right in my opinion, I reason this way:
1) purpose of serialization - subsequent deserialization
2) complete deserialization possible only in same environment as serialization (same classes should be declared, etc)
3) so, enum key is enough and preferable (because it obey PHP identifier syntax, while value can be much complex and large string)

Ofcourse I can understand use cases when serialized object used in other environment, sending to JS frontend for example, this case can be handled with custom handlers.

simPod:
TBH IDK. I guess we should try to align with php behaviour https://3v4l.org/YM5gu?

Pure enums use keys. Backed enums use value. 🤷🏾‍♀️

the-toster:
I checked how it works now, serializer treats enum as regular class, that's why backed & pure deserialized correctly.
So I suggest merged as is (because it have no breaking changes), and then add serialization logic (which with BC anyway), for example from #1384.

simPod:
Here they used values for backed enums https://github.com/spatie/data-transfer-object/pull/277/files#diff-08e8db6ece41c64a6374a66ebb8df6c9c567919411684dbfc699dc4b7113d734  
I'd use keys for basic enums.

Einenlum:
It would be nice to first know the exact expected behavior about this feature: I think @goetas wanted to have the possibility to serialize/unserialize by key instead of value (see [here](https://github.com/schmittjoh/serializer/pull/1384#issuecomment-1123345019))? I think it should be optional (and the default behavior would be to serialize/deserialize the value of the Backed Enum). Seems like some people (like @the-toster [here](https://github.com/schmittjoh/serializer/pull/1384#issuecomment-1125341660)) would like the possibility to serialize/unserialize non backed enums. I'm quite against it, but it should be discussed by the community.

Also, do we know what is missing from @the-toster's code to work as expected and to be merged?

I don't have so much time either right now to really get into the code, but I can maybe find some time in the following weeks to dive a bit more into it.

Anyway, **thank you all for your work**. I know OpenSource takes a lot of time and energy :heart:.

enumag:
The behavior I personally expect is to serialize/deserialize backed enums using their values. Other cases feel more situational so I think it would be best if they were not too complicated to implement in userland code (by implementing a customized SubscribingHandlerInterface or something) but not necessarily supported by the library out of the box.

drakulis:
Guys, I'm really happy to see this PR. Is there anything which prevents you from merging it and releasing new version of serializer?

@the-toster thank you for this PR!

goetas:
Done in https://github.com/schmittjoh/serializer/pull/1448

PR #1394, opened 2022-02-20
Serialization Context Naming Strategy

dgafka:
This introduces possibility to set custom naming strategy for given serialization/deserialization using SerializationContext. 

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

### DIFF
--- a/src/GraphNavigator/SerializationGraphNavigator.php
+++ b/src/GraphNavigator/SerializationGraphNavigator.php
@@ -24,6 +24,7 @@ use JMS\Serializer\GraphNavigator;
 use JMS\Serializer\GraphNavigatorInterface;
 use JMS\Serializer\Handler\HandlerRegistryInterface;
 use JMS\Serializer\Metadata\ClassMetadata;
+use JMS\Serializer\Naming\PropertyNamingStrategyInterface;
 use JMS\Serializer\NullAwareVisitorInterface;
 use JMS\Serializer\SerializationContext;
 use JMS\Serializer\Visitor\SerializationVisitorInterface;
@@ -80,6 +81,10 @@ final class SerializationGraphNavigator extends GraphNavigator
      * @var bool
      */
     private $shouldSerializeNull;
+    /**
+     * @var PropertyNamingStrategyInterface|null
+     */
+    private $contextPropertyNamingStrategy = null;
 
     public function __construct(
         MetadataFactoryInterface $metadataFactory,
@@ -104,6 +109,7 @@ final class SerializationGraphNavigator extends GraphNavigator
 
         parent::initialize($visitor, $context);
         $this->shouldSerializeNull = $context->shouldSerializeNull();
+        $this->contextPropertyNamingStrategy = $context->getPropertyNamingStrategy();
     }
 
     /**
@@ -253,6 +259,10 @@ final class SerializationGraphNavigator extends GraphNavigator
 
                     if (null !== $this->expressionExclusionStrategy && $this->expressionExclusionStrategy->shouldSkipProperty($propertyMetadata, $this->context)) {
                         continue;
+                    }
+
+                    if ($this->contextPropertyNamingStrategy) {
+                        $propertyMetadata->serializedName = $this->contextPropertyNamingStrategy->translateName($propertyMetadata);
                     }
 
                     $v = $this->accessor->getValue($data, $propertyMetadata, $this->context);
--- a/src/SerializationContext.php
+++ b/src/SerializationContext.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace JMS\Serializer;
 
 use JMS\Serializer\Exception\RuntimeException;
+use JMS\Serializer\Naming\PropertyNamingStrategyInterface;
 use Metadata\MetadataFactoryInterface;
 
 class SerializationContext extends Context
@@ -24,6 +25,10 @@ class SerializationContext extends Context
      * @var bool
      */
     private $serializeNull = false;
+    /**
+     * @var PropertyNamingStrategyInterface|null
+     */
+    private $propertyNamingStrategy = null;
 
     public static function create(): self
     {
@@ -50,6 +55,15 @@ class SerializationContext extends Context
         return $this;
     }
 
+    public function setPropertyNamingStrategy(PropertyNamingStrategyInterface $propertyNamingStrategy): self
+    {
+        $this->assertMutable();
+
+        $this->propertyNamingStrategy = $propertyNamingStrategy;
+
+        return $this;
+    }
+
     /**
      * Returns TRUE when NULLs should be serialized
      * Returns FALSE when NULLs should not be serialized
@@ -57,6 +71,11 @@ class SerializationContext extends Context
     public function shouldSerializeNull(): bool
     {
         return $this->serializeNull;
+    }
+
+    public function getPropertyNamingStrategy(): ?PropertyNamingStrategyInterface
+    {
+        return $this->propertyNamingStrategy;
     }
 
     /**
--- a/tests/Fixtures/CustomDeserializationObjectWithInnerClass.php
+++ b/tests/Fixtures/CustomDeserializationObjectWithInnerClass.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Tests\Fixtures;
+
+use JMS\Serializer\Annotation\Type;
+
+class CustomDeserializationObjectWithInnerClass
+{
+    /**
+     * @Type("JMS\Serializer\Tests\Fixtures\CustomDeserializationObject")
+     */
+    #[Type(name: CustomDeserializationObject::class)]
+    private $someProperty;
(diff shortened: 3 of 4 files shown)


dgafka:
I can't find the cause of error in scrutinizer.
@scyzoryck can you point me to it? :)

scyzoryck:
Hi! @dgafka - code looks good, thanks! Could you try to rebase your branch based on the newest master? I hope it might help :) 

Best, scyzoryck.

dgafka:
@scyzoryck @goetas can this be reviewed / merged? :)

scyzoryck (changes requested):
The rest of the code looks good, thanks for contribution! Could you add two missing test scenarios to make sure that we are not missing anything, please? 
It would be nice to add new test case for PHPBench to check how it will impact performance of serialisation ;) 

Best, Marcin.

dgafka:
@scyzoryck can we push this forward? :)

src/GraphNavigator/SerializationGraphNavigator.php
```diff
@@ -257,8 +257,15 @@ public function accept($data, ?array $type = null)
                         continue;
                     }
 
+                    /** Metadata changes based on context, should not be cached  */
+                    $contextSpecificMetadata = $propertyMetadata;
+                    if (null !== $this->context->getPropertyNamingStrategy()) {
+                        $contextSpecificMetadata = clone $propertyMetadata;
+                        $contextSpecificMetadata->serializedName = $this->context->getPropertyNamingStrategy()->translateName($propertyMetadata);
```
scyzoryck, line 264: I'm a bit worried about the performance of this operation - could you add PHPBench test that will allow us to compare performance, please? See `JsonMaxDepthSerializationBench` for reference.

scyzoryck (changes requested):
@dgafka Thanks for changes! I left one additional request. Could you also describe this feature in the docs, please?

src/GraphNavigator/SerializationGraphNavigator.php
dgafka: Ok, added :)

dgafka:
@scyzoryck docs updated.
Can review?

cc:
@goetas

scyzoryck (approved):

dgafka:
@scyzoryck thanks for reviewing :) 

@goetas can we merge this?

src/GraphNavigator/DeserializationGraphNavigator.php
```diff
@@ -209,13 +209,20 @@ public function accept($data, ?array $type = null)
                         continue;
                     }
 
-                    $this->context->pushPropertyMetadata($propertyMetadata);
+                    /** Metadata changes based on context, should not be cached  */
+                    $contextSpecificMetadata = $propertyMetadata;
+                    if (null !== $this->context->getPropertyNamingStrategy()) {
```
dgafka, line 214: @goetas I know you mentioned that this may be slow. 
This is optional however, as `Serializer Builder` exposes `default` property naming strategy:

```php
->setPropertyNamingStrategy()
```

So this will be only used conditionally in case someone demand this. 
Will this be fine?

dgafka:
Will fixing the conflicts make it possible to merge?

scyzoryck:
Hi! @dgafka - fine for me to merge it :)

dgafka:
Hey @scyzoryck, PR rebased and green :)

src/GraphNavigator/DeserializationGraphNavigator.php
```diff
@@ -209,15 +209,21 @@ public function accept($data, ?array $type = null)
...
-                        $v = $this->visitor->visitProperty($propertyMetadata, $data);
-                        $this->accessor->setValue($object, $v, $propertyMetadata, $this->context);
+                        $v = $this->visitor->visitProperty($contextSpecificMetadata, $data);
+                        $this->accessor->setValue($object, $v, $contextSpecificMetadata, $this->context);
                     } catch (NotAcceptableException $e) {
-                        if (true === $propertyMetadata->hasDefault) {
-                            $cloned = clone $propertyMetadata;
-                            $cloned->setter = null;
-                            $this->accessor->setValue($object, $cloned->defaultValue, $cloned, $this->context);
+                        if (true === $contextSpecificMetadata->hasDefault) {
```
scyzoryck, line 224: I'm afraid it can has some side effects if `$this->context->getPropertyNamingStrategy()` is null as we are not going to clone at all.

src/Metadata/PropertyMetadata.php
```diff
@@ -33,6 +33,11 @@ class PropertyMetadata extends BasePropertyMetadata
      */
     public $serializedName;
 
+    /**
+     * @var string|null
+     */
+    public $preContextSerializedName;
```
scyzoryck, line 39: Seems to be not used anymore.

scyzoryck (commented):
Thanks! Great coverage with tests, no impact on the performance. 👍 
I left 2 small comments :)

src/GraphNavigator/DeserializationGraphNavigator.php
dgafka: Ah true. Should we clone always or leave this part code as it was before?

scyzoryck: I would left it as it used to be :)

scyzoryck:
Just in general I wonder about use case that it would solve - I do not know so many cases when it would require dynamic change during of the (de-)serialisation naming. On the other hand generating name of the function was one of the biggest performance updates when migration from 1.x to 2.x so using this feature might cause some performance issues. ;-)

dgafka:
> Just in general I wonder about use case that it would solve - I do not know so many cases when it would require dynamic change during of the (de-)serialisation naming. On the other hand generating name of the function was one of the biggest performance updates when migration from 1.x to 2.x so using this feature might cause some performance issues. ;-)

[The feature is opt-in](https://github.com/schmittjoh/serializer/pull/1394/files#diff-23a7c0ac06a63b607424023dc0e8736603bd6d44a628fa2fe8b62c6526c7e8fdR263), so should not affect current usage if I am not mistaken?

And the use case is to use same JMS instance for two different purposes, like serialization of API responses (camel cases needed) and to do Database Mapping (snake cases needed).

scyzoryck:
> > Just in general I wonder about use case that it would solve - I do not know so many cases when it would require dynamic change during of the (de-)serialisation naming. On the other hand generating name of the function was one of the biggest performance updates when migration from 1.x to 2.x so using this feature might cause some performance issues. ;-)
> 
> [The feature is opt-in](https://github.com/schmittjoh/serializer/pull/1394/files#diff-23a7c0ac06a63b607424023dc0e8736603bd6d44a628fa2fe8b62c6526c7e8fdR263), so should not affect current usage if I am not mistaken?

Yes it is optin - so only performance impact will be with custom naming strategy. 

> And the use case is to use same JMS instance for two different purposes, like serialization of API responses (camel cases needed) and to do Database Mapping (snake cases needed).
Ok, but IMO especially for those operations we should care about performance - so having 2 instances that are cached seems to be still better option for me - specially that they will be in a different application layers and can be configure independently :) For the cases that you mentioned you might also need more options  soon - like different config for enums or datetime handlers. 
We can also make some call if'd be helpful for you :D 

When I was taking another look I spotted another potential issue - when passing it into context it overrides all names - also the name defined in attribute/annotation. This behaviour is different than the default one - so it can potential source of the issues for other users. :(

dgafka:
> so having 2 instances that are cached seems to be still better option for me - specially that they will be in a different application layers and can be configure independently :) For the cases that you mentioned you might also need more options soon - like different config for enums or datetime handlers.

The configuration will stay the same, so that's fine. I've checked how much bootstrapping (`$builder->build`) takes and it around 5ms for the simplest case and I think it may increase when there is more configuration involved. 
Therefore having like three instances (identical property, camel case, underscores), seems like a bit overkill, especially that only one of them will be used in most of the cases.

I may hack my way around, and try to bootstrap them conditionally depending on the usage. This is not ideal but at least there wouldn't be bootstraping penalty. 
But do we know, how much using the solution from the PR would affect performance? As if the percentage is small, then I would still consider it preferable over multiple instances of JMS.

scyzoryck:
With following bench added:
```
class JsonCamelCaseSerializationBench extends JsonSerializationBench
{
    protected function createContext(): SerializationContext
    {
        return parent::createContext()->setPropertyNamingStrategy(new CamelCaseNamingStrategy());
    }
}
```
Results on my local was:
```
\JMS\Serializer\Tests\Benchmark\Performance\JsonCamelCaseSerializationBench

    benchSerialization......................I2 - Mo973.740ms (±2.40%)

\JMS\Serializer\Tests\Benchmark\Performance\JsonSerializationBench

    benchSerialization......................I2 - Mo853.821ms (±2.72%)
```

So the difference seems to be around 15% ;-)

PR #1395, opened 2022-03-09
Add support for Doctrine mapped superclasses

dontub:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT

This PR adds support for the mapped superclasses feature of Doctrine: https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/reference/inheritance-mapping.html


goetas:
I'm confused by this MR. The serializer should already be able to serialize/deserialize parent classes. can you please provide more info on what is the use case?

dontub:
Properties in parent classes are not handled for Doctrine entities. Without the change `DoctrineDriverTest` fails when a mapped superclass is used.

PR #1397, opened 2022-03-14
Fix #1396

### LINKED ISSUE #1396: Deprecation error on Lexer.
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | no

When running JMS on my projects, I can see a deprecation error message in the logs: 
```
php.INFO: 
User Deprecated: Method "Doctrine\Common\Lexer\AbstractLexer::getType()" might add "int|string|null" as a native return type declaration in the future. Do the same in child class "JMS\Serializer\Type\Lexer" now to avoid errors or add an explicit @return annotation to suppress this message.
```

My proposal is to do as the error message from Doctrine says, by already support `int|string|null` as a subset of `mixed` currently used by Doctrine (no return type specified).

EDIT: As the minimum PHP version supported here is PHP 7.2, and the `int|string|null` return type syntax is PHP 8.0+, the proposal must change in favor of the usage of the `@return` annotation. Once minimum supported version will be PHP 8.0+, the return type should be explicitly set instead of using the annotation.

niconoe-:
Set explicit return type on Lexer::getType into annotations to remove the related deprecation error message.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | N/A
| BC breaks?    | no 
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1396
| License       | MIT

### DIFF
--- a/src/Type/Lexer.php
+++ b/src/Type/Lexer.php
@@ -54,6 +54,7 @@ final class Lexer extends AbstractLexer implements ParserInterface
 
     /**
      * {@inheritDoc}
+     * @return int|string|null
      */
     protected function getType(&$value)
     {


ruudk:
Thanks for fixing this. 

@goetas can this be merged and tagged please?

ruudk:
@goetas Could you please tag a new release that includes this fix? 🙏 Thank you.

PR #1404, opened 2022-03-30
fix AttributeReader:getPropertyAnnotation AttributeReader:getMethodAnnotation

andriinavrotskii:
| Bug fix?      | yes

### DIFF
--- a/src/Metadata/Driver/AttributeDriver/AttributeReader.php
+++ b/src/Metadata/Driver/AttributeDriver/AttributeReader.php
@@ -46,8 +46,8 @@ class AttributeReader implements Reader
     {
         $attributes = $method->getAttributes($annotationName);
 
-        $classReflection = new ReflectionClass($method->class);
-        return $this->reader->getClassAnnotation($classReflection, $annotationName) ?? $this->buildAnnotation($attributes);
+        return $this->reader->getClassAnnotation(new ReflectionClass($method->class), $annotationName)
+            ?? $this->buildAnnotation($attributes);
     }
 
     public function getPropertyAnnotations(ReflectionProperty $property): array
@@ -61,8 +61,8 @@ class AttributeReader implements Reader
     {
         $attributes = $property->getAttributes($annotationName);
 
-        $classReflection = new ReflectionClass($property->class);
-        return $this->reader->getClassAnnotation($classReflection, $annotationName) ?? $this->buildAnnotation($attributes);
+        return $this->reader->getClassAnnotation(new ReflectionClass($property->class), $annotationName)
+            ?? $this->buildAnnotation($attributes);
     }
 
     private function buildAnnotation(array $attributes): ?object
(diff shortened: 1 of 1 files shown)


scyzoryck:
Hello @andriinavrotskii ! 

Thank you for your PR! Could you tell a bit more about the issue that is solves? Is it possible to add some unit tests that was not working before this changes? 

Best, Marcin.

goetas:
is there a way to provide a test case for this?

ixarlie:
Looking closer at the AttributeReader class it seems the `getMethodAnnotation` and `getPropertyAnnotation` are not using their counterparts from the delegated reader.

I think the solution from [this commit](https://github.com/schmittjoh/serializer/pull/1404/commits/4d5a08856b2b4f7abe83fc6202979bbb3abd69be) is not correct because when you call `getMethodAnnotation` or `getPropertyAnnotation` you expect to get the annotation instance from the method or from the property respectively, not from the class annotations.

IMO the solution is to call to the delegated reader with the same method names.
```php
public function getMethodAnnotation(ReflectionMethod $method, $annotationName): ?object
{
    $attributes = $method->getAttributes($annotationName);

    return $this->reader->getMethodAnnotation($method, $annotationName) ?? $this->buildAnnotation($attributes);
}

public function getPropertyAnnotation(ReflectionProperty $property, $annotationName): ?object
{
    $attributes = $property->getAttributes($annotationName);

    return $this->reader->getPropertyAnnotation($property, $annotationName) ?? $this->buildAnnotation($attributes);
}
```

src/Metadata/Driver/AttributeDriver/AttributeReader.php
```diff
@@ -46,7 +46,8 @@ public function getMethodAnnotation(ReflectionMethod $method, $annotationName):
     {
         $attributes = $method->getAttributes($annotationName);
 
-        return $this->reader->getClassAnnotation($method, $annotationName) ?? $this->buildAnnotation($attributes);
+        return $this->reader->getClassAnnotation(new ReflectionClass($method->class), $annotationName)
+            ?? $this->buildAnnotation($attributes);
```
ixarlie, line 50:
```suggestion
        return $this->reader->getMethodAnnotation($method, $annotationName) ?? $this->buildAnnotation($attributes);
```

src/Metadata/Driver/AttributeDriver/AttributeReader.php
```diff
@@ -60,7 +61,8 @@ public function getPropertyAnnotation(ReflectionProperty $property, $annotationN
     {
         $attributes = $property->getAttributes($annotationName);
 
-        return $this->reader->getClassAnnotation($property, $annotationName) ?? $this->buildAnnotation($attributes);
+        return $this->reader->getClassAnnotation(new ReflectionClass($property->class), $annotationName)
+            ?? $this->buildAnnotation($attributes);
```
ixarlie, line 65:
```suggestion
        return $this->reader->getPropertyAnnotation($property, $annotationName) ?? $this->buildAnnotation($attributes);
```

goetas:
@mbabker you have worked on this area...would you mind having a look at this?

PR #1330, opened 2021-07-23
Handle union types for typed properties and docblocks

dgafka:
This handle union types for typed properties and DocBlock type resolver. 

The solution allows for defining union types for above, which was previously failing during cache building phase. 
As union type is non implicit in what we are actually serializing / deserializing, we are not trying to recognize which type user want to use in specific context.

How does serialization works?
Serializing knows which type to use, as it actually holds the value that is going to be serialized. It use that value to infer the type and do the serialization. 

How does deserialization works? 
The deserialization will throw an exception as we can't guess, which type user would like to pick. 
If developer will define Type Attribute, it will be chosen as default one. 
This is a place for improvement. Possible we could introduce custom union deserializing handler, that would decide on his own what class he would like to deserialize too. 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/serializer/issues/1282 https://github.com/schmittjoh/serializer/issues/1302
| License       | MIT

PR #1408, opened 2022-04-28
Feature/uninitialize property exception

pag1903:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1367
| License       | MIT

Catch uninitialized property PHP Error and convert it into a UninitializedPropertyException if the Error message match the defined regex.

### DIFF
--- a/src/Exception/UninitializedPropertyException.php
+++ b/src/Exception/UninitializedPropertyException.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Exception;
+
+class UninitializedPropertyException extends RuntimeException implements Exception
+{
+}
--- a/src/Accessor/DefaultAccessorStrategy.php
+++ b/src/Accessor/DefaultAccessorStrategy.php
@@ -7,6 +7,7 @@ namespace JMS\Serializer\Accessor;
 use JMS\Serializer\DeserializationContext;
 use JMS\Serializer\Exception\ExpressionLanguageRequiredException;
 use JMS\Serializer\Exception\LogicException;
+use JMS\Serializer\Exception\UninitializedPropertyException;
 use JMS\Serializer\Expression\CompilableExpressionEvaluatorInterface;
 use JMS\Serializer\Expression\Expression;
 use JMS\Serializer\Expression\ExpressionEvaluatorInterface;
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -36,6 +36,7 @@
         "doctrine/phpcr-odm": "^1.3|^2.0",
         "jackalope/jackalope-doctrine-dbal": "^1.1.5",
         "ocramius/proxy-manager": "^1.0|^2.0",
+        "php": "^7.4||^8.0",
```
scyzoryck, line 39: We should not require different version of PHP for `require-dev` section.

scyzoryck (changes requested):

scyzoryck:
Thank you for your PR! I left we small comments :)

pag1903:
Hi scyzoryck,
Thank you for your feedbacks! I'm replying to your comments.

src/Exception/UninitializedPropertyException.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Exception;
+
+/**
+ * UninitializedPropertyException for the Serializer.
```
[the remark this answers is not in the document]
pag1903: OK fixed.

composer.json
pag1903: I'm agree with that, but I don't what is the best option, because I've added some typed properties in the unit tests (implemented in PHP 7.4) :
* update the global requirements to PHP 7.4, this will force all users to upgrade their PHP version : definitely no!
* replace the typed properties in the test Fixtures by a mock of the DefaultAccessorStrategy which will throw an UninitializedPropertyException.

scyzoryck: I would propose to move typed properties into separate fixtures classes - see: https://github.com/schmittjoh/serializer/pull/1330 for reference.

pag1903: OK thanks, I've implemented a new test case.

scyzoryck:
@goetas Could you approve running workflow for this PR please? ;)

goetas:
this approach to me looks really "brute force". is there a way to check if a prop is initialized in the accessor and return null in that case?

goetas:
but i see that [symfony uses the same approach](https://github.com/symfony/symfony/blob/acef571dc6014328373db0367ba5b0a363a21f7a/src/Symfony/Component/PropertyAccess/PropertyAccessor.php#L434)... so maybe there are no other options ... what about moving then that try/catch in the accessor and return null? in this way we do not need custom code in the graph visitor.

pag1903:
>  this approach to me looks really "brute force". is there a way to check if a prop is initialized in the accessor and return null in that case?

Definitely, I had the same reaction when I saw this code in Symfony. But there is a gap in the PHP typed properties, and there is actually no good way to solve this.

scyzoryck (approved):
Looks good for me. Thanks! @goetas could you take a look? :)

goetas:
thank you!

PR #1413, opened 2022-05-25
ignore proxy object if not in hierarchy

mpoiriert:
| Q             | A
| ------------- | ---
| Bug fix?      | maybe
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | maybe 
| Deprecations? | no 
| Tests pass?   | will see
| Fixed tickets | #... 
| License       | MIT

This is to cover a use case like this:

```PHP
class MyEntity
{

  /**
   * @Serializer\Type("App\DTO\OtherEntity")
   */
  private OtherEntity $composition;
}

```

Currently the code will only work if the class is not currently loaded (since the class_exists don't use the autoload). But if the class is already loaded (E.g. Second object that go trough that loop).

This give the option to create a handler on a existing class like this:

```PHP

 public function serializeOtherEntity(
        JsonSerializationVisitor $visitor,
        $value,
        array $type,
        Context $context
    ) {
        if (null === $value) {
            return null;
        }

        if ($value instanceof OtherEntity) {
            throw new SkipHandlerException();
        }

        return $context->getNavigator()->accept(new OtherEntity($value));
    }

```
From my point of view it's a bug fix since I would have expect this to work like this but we can consider it a new feature. It should be backward compatible since nobody would have done this since the behaviour is bugged.

### DIFF
--- a/src/EventDispatcher/Subscriber/DoctrineProxySubscriber.php
+++ b/src/EventDispatcher/Subscriber/DoctrineProxySubscriber.php
@@ -37,10 +37,9 @@ final class DoctrineProxySubscriber implements EventSubscriberInterface
         $object = $event->getObject();
         $type = $event->getType();
 
-        // If the set type name is not an actual class, but a faked type for which a custom handler exists, we do not
-        // modify it with this subscriber. Also, we forgo autoloading here as an instance of this type is already created,
-        // so it must be loaded if its a real class.
-        $virtualType = !class_exists($type['name'], false);
+        // If the set type name is not an actual class, but a faked type for which a custom handler exists or
+        // object is not an instance of that class we do not modify it with this subscriber.
+        $virtualType = !class_exists($type['name']) || !$object instanceof $type['name'];
 
         if (
             $object instanceof PersistentCollection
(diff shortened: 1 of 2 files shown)


mpoiriert:
@goetas It there any change this would be merge ? Other wise I will do something else on my code base.

goetas:
hmm, this is weird. the uses case covered by that piece of code should handle this situation:

```php
class MyEntity
{

  /**
   * @Serializer\Type("SOME_INTENTIONAL_WRONG_CLASS_NAME_USED_FOR_A_TYPE_HANDLER")
   */
  private OtherEntity $composition;
}```

So a type handler registers on `SOME_INTENTIONAL_WRONG_CLASS_NAME_USED_FOR_A_TYPE_HANDLER` and does what needs to be done....

if the type-hint is the same as the class name, than i'm not sure what is the expected behaviour. can you please clarify?

(sorry for the late answer)

mpoiriert:
The issue is that your test is done with an arbitrary string that is not a real class. My use case is with a real class that does exists 'App\DTO\OtherEntity'.

Check at my PR if you want more context.

After that I will also have another issue for my use case but I will open another ticket for it. (Using two different type for the same object in the same execution flow).

mpoiriert:
@goetas Any decision on this ?

PR #1416, opened 2022-06-17
float/double add third parameter to define fixed numbers of decimals.…

### LINKED ISSUE #1415: Float serialization in XML
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

visitDouble for XmlSerializationVisitor https://github.com/schmittjoh/serializer/blob/0207dc23f26d1dff4930c0b7780948edb789592f/src/XmlSerializationVisitor.php#L189 return a float and so not respect decimal paramater

Should it be more similar to json, https://github.com/schmittjoh/serializer/blob/0207dc23f26d1dff4930c0b7780948edb789592f/src/JsonSerializationVisitor.php#L71 ?

```php
use JMS\Serializer\Annotation\Type;

class Example
{
  /**
   * @Type("float<0>")
   */
  protected $entry;
}
```

## Expected Result

```xml
<?xml version="1.0" encoding="utf-8"?>
<example>
  <entry>5</entry>
</example>
```

## Actual Result

```xml
<?xml version="1.0" encoding="utf-8"?>
<example>
  <entry>5.0</entry>
</example>
```

riccardonar:
… In json this param force to cast the result as a string. fix #1415

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1415
| License       | MIT

### DIFF
--- a/src/XmlSerializationVisitor.php
+++ b/src/XmlSerializationVisitor.php
@@ -180,11 +180,11 @@ final class XmlSerializationVisitor extends AbstractVisitor implements Serializa
     public function visitDouble(float $data, array $type)
     {
         $dataResult = $data;
-        $percision = $type['params'][0] ?? null;
-        if ($percision) {
+        $precision = $type['params'][0] ?? null;
+        if (is_int($precision)) {
             $roundMode = $type['params'][1] ?? null;
             $roundMode = $this->mapRoundMode($roundMode);
-            $dataResult = round($dataResult, $percision, $roundMode);
+            $dataResult = round($dataResult, $precision, $roundMode);
         }
 
         $decimalsNumbers = $type['params'][2] ?? null;
--- a/src/XmlSerializationVisitor.php
+++ b/src/XmlSerializationVisitor.php
@@ -179,14 +179,26 @@ final class XmlSerializationVisitor extends AbstractVisitor implements Serializa
      */
     public function visitDouble(float $data, array $type)
     {
+        $dataResult = $data;
         $percision = $type['params'][0] ?? null;
-        if (is_int($percision)) {
+        if ($percision) {
             $roundMode = $type['params'][1] ?? null;
             $roundMode = $this->mapRoundMode($roundMode);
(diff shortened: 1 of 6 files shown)


src/XmlSerializationVisitor.php
```diff
@@ -179,14 +179,26 @@ public function visitInteger(int $data, array $type)
      */
     public function visitDouble(float $data, array $type)
     {
+        $dataResult = $data;
         $percision = $type['params'][0] ?? null;
-        if (is_int($percision)) {
+        if ($percision) {
```
scyzoryck, line 184: What about precision that equals `0`? Can we add test case to make sure that it works as previously?

scyzoryck (commented):
Thank you @riccardonar for your contribution! I left few comments - could you take a look at them please? :)

src/XmlSerializationVisitor.php
riccardonar: you are right, my error

src/XmlSerializationVisitor.php
```diff
@@ -179,14 +179,26 @@ public function visitInteger(int $data, array $type)
...
+        if (is_int($precision)) {
             $roundMode = $type['params'][1] ?? null;
             $roundMode = $this->mapRoundMode($roundMode);
-            $data = round($data, $percision, $roundMode);
+            $dataResult = round($dataResult, $precision, $roundMode);
         }
 
-        return $this->document->createTextNode(var_export((float) $data, true));
+        $decimalsNumbers = $type['params'][2] ?? null;
+        if ($decimalsNumbers === null) {
```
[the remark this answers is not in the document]
scyzoryck: @goetas what do you think? It looks like it was a bug introduced long time ago.

goetas:
Done in https://github.com/schmittjoh/serializer/pull/1456

PR #1362, opened 2021-10-31
Accept FormInterface instead of Form in FormErrorHandler

W0rma:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/schmittjoh/serializer/issues/601
| License       | MIT

PR #1430, opened 2022-09-07
fix(FormErrorHandler): add method to FormInterface

titoko:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT

Since Symfony 6.1 FormErrorIterator property $form is typehint to FormInterface since https://github.com/symfony/symfony/commit/85065e4f76eb0e75163277498bbc4fadb3ae49f8 .

This PR https://github.com/schmittjoh/serializer/pull/1362 add the possibility to serialize Interface but lead to the following error
`call_user_func_array(): Argument #1 ($callback) must be a valid callback, class JMS\Serializer\Handler\FormErrorHandler does not have a method "serializeFormInterfaceTojson"`
( see also https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2371 )

Adding the method option fix this issue

### DIFF
--- a/src/Handler/FormErrorHandler.php
+++ b/src/Handler/FormErrorHandler.php
@@ -43,6 +43,7 @@ final class FormErrorHandler implements SubscribingHandlerInterface
                 'direction' => GraphNavigatorInterface::DIRECTION_SERIALIZATION,
                 'type' => FormInterface::class,
                 'format' => $format,
+                'method' => 'serializeFormTo' . ucfirst($format),
             ];
             $methods[] = [
                 'direction' => GraphNavigatorInterface::DIRECTION_SERIALIZATION,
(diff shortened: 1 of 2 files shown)


scyzoryck:
Hello! 
Thanks for the contribution! 
Would it be possible to add some test cases for it, please? Just to make sure that it works for older & new versions. 

Best, Marcin.

titoko:
Well, don't know if the test is really relevant but at least it should avoid regression :) .
Tell me if it's enough for you

scyzoryck:
I think it would be better to check if it is callable - to make sure that it is not dependent on function name :) 
What do you think about something like:
```php
public function testSerializeFormInterface()
    {
        $registry = new HandlerRegistry([]);
        $registry->registerSubscribingHandler($this->handler);
        $callback = $registry->getHandler(GraphNavigatorInterface::DIRECTION_SERIALIZATION, FormInterface::class, 'json');
        $serialised = $callback($this->visitor, $this->createForm(), []);
        // ...assert result? :) 
    }
```
It looks like it fails with similar result:
```
Error: Call to undefined method JMS\Serializer\Handler\FormErrorHandler::serializeFormInterfaceTojson()
```

goetas:
done in https://github.com/schmittjoh/serializer/pull/1432

PR #1445, opened 2022-12-23
Allow doctrine/lexer 2

derrabus:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT

This PR allows the installation of `doctrine/lexer` 2.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "php": "^7.2||^8.0",
         "doctrine/annotations": "^1.13",
         "doctrine/instantiator": "^1.0.3",
-        "doctrine/lexer": "^1.1",
+        "doctrine/lexer": "^1.1 || ^2",
         "jms/metadata": "^2.6",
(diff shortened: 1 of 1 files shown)


scyzoryck (commented):
Thanks for contribution! 
@goetas could you run CI tests to see if we have any tests that will use that package?

goetas:
hmm, that is weird. can't understand why phpstan fails...

derrabus:
> hmm, that is weird. can't understand why phpstan fails...

Because you're referencing a class from the doctrine/cache package that has been removed in version 2. This is unrelated to my change and can be observed on the master branch as well.

derrabus:
My attempt to fix the PHPStan failure: #1447.

goetas:
Thanks @derrabus ! really appreciated!

PR #1448, opened 2023-01-02
add php 8.1 enum support

goetas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | maybe??
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1393  #1392 #1373
| License       | MIT

### DIFF
--- a/doc/reference/annotations.rst
+++ b/doc/reference/annotations.rst
@@ -316,6 +316,7 @@ the object has been deserialized.
 ~~~~~~~~~~~~~~
 
 .. versionadded : 0.12
+
     @Discriminator was added
 
 This annotation allows serialization/deserialization of relations which are polymorphic, but
@@ -377,6 +378,22 @@ Available Types:
 | array<K, V>                                                | A map of keys of type K to values of type V.     |
 |                                                            | Examples: array<string, string>,                 |
 |                                                            | array<string, MyNamespace\MyObject>, etc.        |
++------------------------------------------------------------+--------------------------------------------------+
+| enum<'Color'>                                              | Enum of type Color, use its case values          |
+|                                                            | for serialization and deserialization            |
+|                                                            | if the enum is a backed enum,                    |
+|                                                            | use its case names if it is not a backed enum.   |
++------------------------------------------------------------+--------------------------------------------------+
+| enum<'Color', 'name'>                                      | Enum of type Color, use its case names           |
+|                                                            | (as string) for serialization                    |
+|                                                            | and deserialization.                             |
++------------------------------------------------------------+--------------------------------------------------+
+| enum<'Color', 'value'>                                     | Backed Enum of type Color, use its case value    |
+|                                                            | for serialization and deserialization.           |
++------------------------------------------------------------+--------------------------------------------------+
+| enum<'Color', 'value', 'integer'>                          | Backed Enum of type Color, use its case value    |
+|                                                            | (forced as integer) for serialization            |
+|                                                            | and deserialization.                             |
 +------------------------------------------------------------+--------------------------------------------------+
 | DateTime                                                   | PHP's DateTime object (default format*/timezone) |
 +------------------------------------------------------------+--------------------------------------------------+
@@ -917,3 +934,23 @@ Example:
             return 1;
         }
     ...
+
+Enum support
+~~~~~~~~~~~~
+
+Enum support is disabled by default, to enable it run:
+
+.. code-block :: php
+
+    $builder = SerializerBuilder::create();
+    $builder->enableEnumSupport();
+
+    $serializer = $builder->build();
+
+
+With the enum support enabled, enums are automatically detected using typed properties typehints.
+When typed properties are no available (virtual properties as example), it is necessary to explicitly typehint
+the underlying type using the ``@Type`` annotation.
+
+- If the enum is a ``BackedEnum``, the case value will be used for serialization and deserialization by default;
+- If the enum is not a ``BackedEnum``, the case name will be used for serialization and deserialization by default;
--- a/src/Metadata/Driver/EnumPropertiesDriver.php
+++ b/src/Metadata/Driver/EnumPropertiesDriver.php
@@ -1,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Metadata\Driver;
+
+use JMS\Serializer\Metadata\ClassMetadata as SerializerClassMetadata;
+use JMS\Serializer\Metadata\ExpressionPropertyMetadata;
+use JMS\Serializer\Metadata\PropertyMetadata;
+use JMS\Serializer\Metadata\StaticPropertyMetadata;
+use JMS\Serializer\Metadata\VirtualPropertyMetadata;
+use Metadata\ClassMetadata;
+use Metadata\Driver\DriverInterface;
+use ReflectionClass;
+use ReflectionException;
+use ReflectionProperty;
+
+class EnumPropertiesDriver implements DriverInterface
+{
+    /**
+     * @var DriverInterface
+     */
+    protected $delegate;
+
+    public function __construct(DriverInterface $delegate)
+    {
+        $this->delegate = $delegate;
+    }
+
+    public function loadMetadataForClass(ReflectionClass $class): ?ClassMetadata
+    {
+        $classMetadata = $this->delegate->loadMetadataForClass($class);
+        \assert($classMetadata instanceof SerializerClassMetadata);
+
+        if (null === $classMetadata) {
+            return null;
+        }
+
+        // We base our scan on the internal driver's property list so that we
+        // respect any internal white/blacklisting like in the AnnotationDriver
+        foreach ($classMetadata->propertyMetadata as $propertyMetadata) {
+            // If the inner driver provides a type, don't guess anymore.
+            if ($propertyMetadata->type || $this->isVirtualProperty($propertyMetadata)) {
+                continue;
+            }
+
+            try {
+                $propertyReflection = $this->getReflection($propertyMetadata);
+                if ($enum = $this->getEnumReflection($propertyReflection)) {
+                    $serializerType = ['name' => 'enum', 'params' => [$enum->getName(), $enum->isBacked() ? 'value' : 'name']];
+                    $propertyMetadata->setType($serializerType);
+                }
+            } catch (ReflectionException $e) {
+                continue;
+            }
+        }
+
+        return $classMetadata;
+    }
+
+    private function isVirtualProperty(PropertyMetadata $propertyMetadata): bool
+    {
+        return $propertyMetadata instanceof VirtualPropertyMetadata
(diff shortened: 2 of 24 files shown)


tests/Fixtures/ObjectWithEnums.php
```diff
@@ -0,0 +1,57 @@
...
+    public BackedSuitInt $backedIntAutoDetect;
+
+    /**
+     * @Serializer\Type("enum<'JMS\Serializer\Tests\Fixtures\Enum\BackedSuitInt', 'value'>")
+     */
+    public BackedSuitInt $backedInt;
+
+    /**
+     * @Serializer\Type("enum<'JMS\Serializer\Tests\Fixtures\Enum\BackedSuit', 'name'>")
+     */
```
ruudk, line 36: Would be great to also have a testcase that uses Attributes.

doc/reference/annotations.rst
```diff
@@ -917,3 +930,13 @@ Example:
...
     ...
+
+Enum support
+~~~~~~~~~~~~
+
+Enums are automatically detected using typed properties typehints.
+When typed properties are no available (virtual properties as example), it is necessary to explicitly typehint
+the underlying type using the ``@Type`` annotation.
+
+- If the enum is a ``BackedEnum``, the cae value will be used for serialization and deserialization by default;
```
ruudk, line 941:
```suggestion
- If the enum is a ``BackedEnum``, the case value will be used for serialization and deserialization by default;
```

tests/Fixtures/ObjectWithEnums.php
goetas: unfortunately the `@Serializer\Type` annotation takes only one param as a string, so except of doing string interpolations, there is no other way to make this wortk.

but that is something that would be interesting to refactor, adding support for something like

`#Type(name: BackedSuit::class, params: [....])`

ruscon: @goetas Hi. I'm trying to use your branch to check how it works.
Could you give an example of how to use `VirtualProperty` along with enum `Type`?

doc/reference/annotations.rst
```diff
@@ -917,3 +930,13 @@ Example:
             return 1;
         }
     ...
+
+Enum support
+~~~~~~~~~~~~
+
+Enums are automatically detected using typed properties typehints.
```
scyzoryck, line 937: It looks like it is not added to `SerializerBuilder::addDefaultHandlers()` so it needs to be added manually? 
Is it possible how many people adopted Serializer at PHP8.1 already? I wonder if it can be a breaking change for them.

src/Metadata/Driver/EnumPropertiesDriver.php
```diff
@@ -0,0 +1,83 @@
...
+        foreach ($classMetadata->propertyMetadata as $propertyMetadata) {
+            // If the inner driver provides a type, don't guess anymore.
+            if ($propertyMetadata->type || $this->isVirtualProperty($propertyMetadata)) {
+                continue;
+            }
+
+            try {
+                $propertyReflection = $this->getReflection($propertyMetadata);
+                if ($enum = $this->getEnumReflection($propertyReflection)) {
+                    $serializerType = ['name' => 'enum', 'params' => [$enum->getName(), $enum->isBacked() ? 'value' : 'name']];
```
scyzoryck, line 50: May it would be better to change the type on ClassMetadata? At the moment it will create inconsistency between handling Enums in objects and arrays. 
Object:
```json
{
    "ordinary_auto_detect":"Clubs"
}
```
Array of Enums:
```json
[{"name":"Clubs"}]
```
Single Enum:
```json
{"name":"Clubs"}
```

src/Metadata/Driver/EnumPropertiesDriver.php
```diff
@@ -0,0 +1,83 @@
...
+use JMS\Serializer\Metadata\PropertyMetadata;
+use JMS\Serializer\Metadata\StaticPropertyMetadata;
+use JMS\Serializer\Metadata\VirtualPropertyMetadata;
+use Metadata\ClassMetadata;
+use Metadata\Driver\DriverInterface;
+use ReflectionClass;
+use ReflectionException;
+use ReflectionProperty;
+
+class EnumPropertiesDriver implements DriverInterface
```
scyzoryck, line 18: I think it is nice approach to override the type in general, but the risk that I can see here is a situation when someone will not have EnumHandler enabled. I would add one test with this enum disabled - that would also answer for question about possible BC breaks.

tests/Fixtures/ObjectWithEnums.php
ruudk: Or create a specific EnumType attribute? 

```
#[EnumType(BackedSuit::class)
```

It can then also have more properties for the `value` / `integer` thing. 

No need to support Annotations for this, as most people that use an PHP 8.0 enum are also people that already upgraded to Attributes 😉

ruscon: Any solution with attributes for this case?
```php
#[Type(name: 'list<enum<MyEnum>>')]
```

ruudk: Or composable attributes like this:
`#[Type(new ListType(new EnumType(MyEnum::class))]`

doc/reference/annotations.rst
```diff
@@ -917,3 +930,13 @@ Example:
...
+When typed properties are no available (virtual properties as example), it is necessary to explicitly typehint
+the underlying type using the ``@Type`` annotation.
+
+- If the enum is a ``BackedEnum``, the case value will be used for serialization and deserialization by default;
```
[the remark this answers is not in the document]
Einenlum: (apart from that, thanks a lot for your work :heart: )

tests/Fixtures/ObjectWithEnums.php
goetas: changing the `@Type` syntax of out of the scope of the current PR (of course i will accept a PR that improves that).

I would like to see something as 

```php
#[Type('enum', [MyEnum::class]])]
#[Type('array, [['enum', [MyEnum::class]]])]
#[Type('array, ['string', ['enum', [MyEnum::class]]])]
```

that is equivalent to

```
@Type("enum<MyEnum>")
@Type("array<enum<MyEnum>>")
@Type("array<enum<string,MyEnum>>")
```

maybe there is a mess with `[]`....

src/Metadata/Driver/EnumPropertiesDriver.php
goetas: good catch! that is not easy to solve at all :(

tests/Fixtures/ObjectWithEnums.php
enumag: I think I'd suggest a syntax like this. It's consistent in how generic types are noted no matter how deep you go.

```php
#[Type('enum', new Type(MyEnum::class))]
#[Type('array', new Type('enum', new Type(MyEnum::class)))]
#[Type('array', 'string', new Type('enum', new Type(MyEnum::class)))]
```

ruudk: I agree with @enumag. When designing the Attribute syntax, take full advantage of the possibilities there.

ruscon: @ruudk your solution could be as aproxy on top of @enumag's solution

goetas: >I think I'd suggest a syntax like this using nested attributes, variadic parameter and union type.

that looks very good

src/Metadata/Driver/EnumPropertiesDriver.php
goetas: i have solved this with a subscriber

doc/reference/annotations.rst
goetas: fixed

goetas:
Im thinking if is worth trying to make it backward compatible or just make a major release...

goetas:
@Einenlum 

> It seems like using enum<'Color'> makes it use the name of the enum and not its value, according to what you wrote before (line 382), so I'm a bit confused.

I have updated the documentation.

- If an enum is a `BackedEnum`, buy default it will be serialized using its `value`
   -   user can change it to `name` it prefers
- If an enum is NOT `BackedEnum`, it will be serialized using its `name`

Despite the fact that the RFC suggests that non-backed enums should not be serialized, I do not see anything wrong with offering such possibility.

Regarding the `integer` option, TBH is an overkill, i can remove it. I've dealt with some "weird" systems in which the backed enum was a string but containing integers... but maybe is not worth solving

ruudk (approved):
Thanks for this. Hopefully the Attribute syntax can be changed later. I think referencing the type using a `::class` reference makes it more robust.

Einenlum:
Thanks @goetas !
The implementation sounds way better and simple to me this way.
For the integer option, sounds indeed overkill to me since it's probably to solve some edge case that shouldn't happen in most codebases.

One thing that I don't really get though (but probably a misunderstanding on my side): why make it opt-in and not do the same as with other drivers [here](https://github.com/schmittjoh/serializer/blob/master/src/Builder/DefaultDriverFactory.php#L48-L71) ?

Sounds strange to make that opt-in since it sounds like a normal behavior that should work out of the box.

goetas:
>One thing that I don't really get though (but probably a misunderstanding on my side): why make it opt-in and not do the same as with other drivers [here](https://github.com/schmittjoh/serializer/blob/master/src/Builder/DefaultDriverFactory.php#L48-L71) ?

the issue is that some people might be using already the serializer with the default enum serialization, see `testEnumDisabledByDefault`. the serializer was already able to serialize enums... and with this PR the format would change...

tests/Serializer/BaseSerializationTest.php
```diff
@@ -2119,8 +2169,14 @@ static function (DeserializationVisitorInterface $visitor, $data, $type, Context
 
         $this->dispatcher = new EventDispatcher();
         $this->dispatcher->addSubscriber(new DoctrineProxySubscriber());
+        $this->dispatcher->addSubscriber(new EnumSubscriber());
 
         $builder = SerializerBuilder::create($this->handlerRegistry, $this->dispatcher);
+
+        if (PHP_VERSION_ID >= 80100) {
```
scyzoryck, line 2176: Just wonder if we should move this PHP version check to `enableEnumSupport` to be sure that it will not fail on lower versions. 🤔 But on the other hand it will be better to fail fast.

scyzoryck (approved):
Looks good! ❤️

tests/Serializer/BaseSerializationTest.php
goetas: done

ruudk:
Would be great to have this tagged so that we can start using it 💙

Einenlum:
:tada: Thanks a lot for you work @goetas ! Great news for JMS Serializer users.

PR #1454, opened 2023-01-06
fix ignored null type

simPod:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Tests pass?   | yes/no
| License       | MIT

### DIFF
--- a/src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php
+++ b/src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php
@@ -325,7 +325,7 @@ final class DocBlockTypeResolver
         return class_exists($typeHint) || interface_exists($typeHint);
     }
 
-    private function resolveTypeFromDocblock(\ReflectionProperty $reflectionProperty): ?array
+    private function resolveTypeFromDocblock(\ReflectionProperty $reflectionProperty): array
     {
         $docComment = $reflectionProperty->getDocComment();
         if (!$docComment && PHP_VERSION_ID >= 80000 && $reflectionProperty->isPromoted()) {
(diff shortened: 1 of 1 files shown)


simPod:
v3.19 is broken since there's `count()` called on `null` value in `DocBlockTypeResolver::getPropertyDocblockTypeHint()`

goetas:
@dgafka can you please have a look?

dgafka:
@simPod can you add test case for this, to show the error?

simPod:
I'd not add tests to test types but keep it for phpstan https://github.com/schmittjoh/serializer/pull/1446

dgafka (approved):
@simPod I can spot the issue.

I would still add the test case where it fails for you, as it means that there is real life scenario that we've not covered.

simPod:
@dgafka yup, that will be covered by phpstan. Such tests, that can be replaced by SA, are basically a mess that is hard to get rid of then SA is enabled.

smoench:
I think https://github.com/schmittjoh/serializer/blob/master/tests/Fixtures/ObjectWithPhpDocProperty.php needs to be extended with a property without a docblock and in https://github.com/schmittjoh/serializer/blob/master/tests/Metadata/Driver/DocBlockTypeResolverTest.php following test needs to be added:
```php
public function testGetPropertyDocblockTypeHintDoesNotCrashOnAPropertyWithoutADocBlock(): void
    {
        $resolver = new DocBlockTypeResolver();
        self::assertNull(
            $resolver->getPropertyDocblockTypeHint(
                new ReflectionProperty(ObjectWithPhpDocProperty::class, 'noneDocBlock')
            )
        );
    }

```

simPod:
Maybe but that's not relevant to this PR. I just fixed a wrong type in a private method.

dgafka:
ping @goetas

goetas:
thank you!

dgafka:
@goetas can you release? :)

goetas:
https://github.com/schmittjoh/serializer/releases/tag/3.21.0

PR #1455, opened 2023-01-06
Make doblock type resolver compatibile with newest phpdoc

dgafka:
This fixes compatibility with newest version of `phpstan/phpdoc-parser`:
Screenshot from 2023-01-06 18-36-21

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT


tests/Metadata/Driver/DocBlockDriverTest.php
```diff
@@ -387,6 +388,16 @@ public function testInferTypeForPhpstanArray()
         );
     }
 
+    public function testInferTypeForPhpstanNestedArrayShape()
```
dgafka, line 391: just additional test for higher confidence.

dgafka:
@goetas this fixes `3.19.0` in case someone uses phpstan array type with `phpstan/phpdoc-parser` >= `1.0`

goetas:
@dgafka I have to say that `src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php` is getting a Mess... so  many edge cases... maybe is doing too much?

dgafka:
That's true, there are a lot of concepts hidden inside single class, so it could be refactored. 
If nobody will do it, I may pick it up, when I will be adding some new feature. :)

Issue #1457, opened 2023-01-09
[3.20.0] The provided class "MyEnum" is an enum, and cannot be instantiated


ruudk:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? |no
| BC Break report? |no
| RFC?             |no

I'm super grateful for the work on enums in https://github.com/schmittjoh/serializer/pull/1448. I'm just testing it out with `jms/serializer v3.20.0` together with [`jms/serializer-bundle dev-enable-enum`](https://github.com/schmittjoh/JMSSerializerBundle/pull/915) and noticed something wrong.

## Steps required to reproduce the problem

```php
enum SocialPlatform : string
{
    case Spotify = 'spotify';
    case Instagram = 'instagram';
}

class MyEvent {
    public function __construct(
        private int $userId,
        private SocialPlatform $socialPlatform,
    ) {
    }
}

$object = new MyEvent(1, SocialPlatform::instagram);

// This is in a functional test, using Symfony Bundle, with `enum_support` enabled.
$this->serializer = self::getContainer()->get('jms_serializer');

$this->context = SerializationContext::create()
    ->enableMaxDepthChecks()
    ->setSerializeNull(true);

$serialized = $this->serializer->serialize($object, 'json', $this->context);
// serializes to: {"user_id":1,"social_platform":"instagram"}

$deserialized = $this->serializer->deserialize($serialized, MyEvent::class, 'json');
// fails Doctrine\Instantiator\Exception\InvalidArgumentException : The provided class "SocialPlatform" is an enum, and cannot be instantiated
```

## Expected Result

No error, deserialized back to an object.

## Actual Result

```
Doctrine\Instantiator\Exception\InvalidArgumentException : The provided class "SocialPlatform" is an enum, and cannot be instantiated
 /Volumes/CS/www/app/vendor/jms/serializer/src/Construction/UnserializeObjectConstructor.php:22
 /Volumes/CS/www/app/vendor/jms/serializer/src/Construction/DoctrineObjectConstructor.php:72
 /Volumes/CS/www/app/vendor/jms/serializer/src/GraphNavigator/DeserializationGraphNavigator.php:189
 /Volumes/CS/www/app/vendor/jms/serializer/src/JsonDeserializationVisitor.php:188
 /Volumes/CS/www/app/vendor/jms/serializer/src/GraphNavigator/DeserializationGraphNavigator.php:214
[...]

PR #1458, opened 2023-01-10
Ignore enums in TypedPropertiesDriver

ruudk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Doc updated   |no
| BC breaks?    |no 
| Deprecations? |no
| Tests pass?   | yes
| Fixed tickets | #1457
| License       | MIT

This is handled by the EnumPropertiesDriver.

If the TypedPropertiesDriver sets the type, it will be skipped by the EnumPropertiesDriver and causes deserialization to fail.

### DIFF
--- a/src/Metadata/Driver/TypedPropertiesDriver.php
+++ b/src/Metadata/Driver/TypedPropertiesDriver.php
@@ -117,6 +117,10 @@ class TypedPropertiesDriver implements DriverInterface
             return true;
         }
 
+        if (PHP_VERSION_ID >= 80100 && enum_exists($reflectionType->getName())) {
+            return false; // handled by EnumPropertiesDriver
+        }
+
         return class_exists($reflectionType->getName())
             || interface_exists($reflectionType->getName());
     }


ruudk:
@goetas When you have time, I'd like to know if this is the idea. If so, we should add some tests to guard it. Ideally, the [JsonSerializationTest](https://github.com/schmittjoh/serializer/blob/master/tests/Serializer/JsonSerializationTest.php) also sets up this TypedPropertiesDriver properly.

simPod (changes requested):
I tried this commit in my project but does not fix the issue. The TypedPropertiesDriver is not hit at all.

This is my constructor

```php
/** @param iterable<Pair<IdValue, EnumType>>|null $a*/
    private function __construct(
        #[Serializer\Type(name: 'iterable<Ds\Pair<IdValue, EnumType>>')]
        public iterable|null $a,
    ) {
    }
```

ruudk:
AFAIK the TypedPropertiesDriver is only hit when the type is not defined. In your case, it is because there is a Type attribute?

ruudk:
@goetas Could you assist in how to proceed with this PR? Looking at @simPod's comment, I think we're overseeing something.

ruudk:
@goetas Hope you're feeling better. Do you have time to help us find the problem here? Is my solution correct?

goetas:
i have tried to look into this but it is confusing... `EnumPropertiesDriver` is registered before the `TypedPropertiesDriver`, thus this should not be a problem at all... is there a way to reproduce the issue?

PR #1460, opened 2023-01-10
fix: do not try to tokenize doc comment when there's none

simPod:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| License       | MIT

This fixes another type issues (will be covered by phpstan when we ship https://github.com/schmittjoh/serializer/pull/1446 and following PRs)

### DIFF
--- a/src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php
+++ b/src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php
@@ -329,7 +329,15 @@ final class DocBlockTypeResolver
     {
         $docComment = $reflectionProperty->getDocComment();
         if (!$docComment && PHP_VERSION_ID >= 80000 && $reflectionProperty->isPromoted()) {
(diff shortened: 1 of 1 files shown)


goetas:
@dgafka ?

dgafka:
Thank @simPod for rising the PR.
From my understanding this solves scenario where there is no constructor and we went to the path for checking that. 
Is this correct @simPod ?

If that's the case, this is totally valid scenario that having test case for would secure us for future refactors. 
If you're unsure how to write such test, please take a look on `testInferTypeForConstructorPropertyPromotion` as a reference :)

simPod:
@dgafka TBH I did not investigate in what scenario it fails, I just considered it a "compile" error because of invalid types.

I had no time to investigate what code is not supported exactly and had to hot fix it production with my fork so I don't have a reproducible.

I just knew that here `->getConstructor()->getDocComment()` can occur a NPE and `getDocComment` returns `string|false` so I handled it.

dgafka:
This can be closed now #1462 :)

PR #1466, opened 2023-01-25
Add support for symfony/uid UUID type

lhess:
The SymfonyUidHandler is lacking support for the `\Symfony\Component\Uid\Uuid` super type that is available since v5.1 of the `symfony/uid` component.

Supporting this super type is helpful in situations in which an interface does not care about which exact UUID version is actually used.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

### DIFF
--- a/src/Handler/SymfonyUidHandler.php
+++ b/src/Handler/SymfonyUidHandler.php
@@ -13,6 +13,7 @@ use JMS\Serializer\Visitor\SerializationVisitorInterface;
 use JMS\Serializer\XmlSerializationVisitor;
 use Symfony\Component\Uid\AbstractUid;
 use Symfony\Component\Uid\Ulid;
+use Symfony\Component\Uid\Uuid;
 use Symfony\Component\Uid\UuidV1;
 use Symfony\Component\Uid\UuidV3;
 use Symfony\Component\Uid\UuidV4;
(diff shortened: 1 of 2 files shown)


mbabker:
For reference, I purposefully didn't add the base Uuid class as I was trying to keep feature parity with the normalizer in Symfony's component.  Based on what they're testing and https://github.com/symfony/symfony/pull/44721 which deprecates support for the base abstract class in favor of requiring a version-specific sub-class, I'm not sure adding support for the base class is a good idea.

lhess:
> For reference, I purposefully didn't add the base Uuid class as I was trying to keep feature parity with the normalizer in Symfony's component. Based on what they're testing and [symfony/symfony#44721](https://github.com/symfony/symfony/pull/44721) which deprecates support for the base abstract class in favor of requiring a version-specific sub-class, I'm not sure adding support for the base class is a good idea.

The difference here is, that `\Symfony\Component\Uid\Uuid` is not abstract. It also valid to use `\Symfony\Component\Uid\Uuid` as a lower boundary in an interface: My API does not care which UUID version was used to generate an UUID, I only care about that it is an UUID.

scyzoryck (approved):
Thanks for contribution! 
@goetas do you have some time to take a look? :)

PR #1467, opened 2023-01-25
Add support for UUIDv7 and UUIDv8 classes

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT

Symfony 6.2 adds support for UUIDv7 and UUIDv8, this adds those classes to the serialization handler.

### DIFF
--- a/src/Handler/SymfonyUidHandler.php
+++ b/src/Handler/SymfonyUidHandler.php
@@ -18,6 +18,8 @@ use Symfony\Component\Uid\UuidV3;
 use Symfony\Component\Uid\UuidV4;
 use Symfony\Component\Uid\UuidV5;
 use Symfony\Component\Uid\UuidV6;
+use Symfony\Component\Uid\UuidV7;
+use Symfony\Component\Uid\UuidV8;
 
 final class SymfonyUidHandler implements SubscribingHandlerInterface
 {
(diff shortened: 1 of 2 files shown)


scyzoryck (commented):
Thanks for contribution! 
Could you take a look at failing PHPStan in lower PHP versions, please?

mbabker:
The PHP 8.2 PHPStan issue looks to need a baseline update (the ignored `~Access to an undefined property ReflectionProperty::\$info~` error looks to no longer be in the error list).

The PHP 8.0 PHPStan issue, before being hard cancelled, is because it runs on a Symfony version without the new classes and groans they're referenced unconditionally (which isn't an issue for the serializer runtime itself) and doesn't like the conditionals used to add the extra test cases.  I don't do a lot with PHPStan where I have it running across multiple PHP versions (everything OSS I use it on only checks the latest versions of the dependency tree as you start ending up with a lot of false positive incompatibilities otherwise), so I really have no idea how to best deal with those issues as they aren't going to be PHP version specific but Symfony component version specific (PHP 8.0 and earlier are never going to be able to use a Symfony version with those classes, but PHP 8.1 and later can run with a Symfony version where they don't exist).

scyzoryck:
I will take a look how we can fix it for lower Symfony versions. :)

goetas:
PHPstan is getting annoying :(

Issue #1452, opened 2023-01-04
Decouple `AttributeReader` from `doctrine/annotations`


mbabker:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

The `JMS\Serializer\Metadata\Driver\AttributeDriver\AttributeReader` class has a hard dependency to the `doctrine/annotations` package in that it implements the annotation reader interface and requires a reader to be provided.  It looks like this is because attributes support was added by decorating a `Doctrine\Common\Annotations\Reader` implementation to inject it into the `JMS\Serializer\Metadata\Driver\AnnotationDriver` class, which rightfully is dependent on an annotation reader. It would be great if this class could function without the annotations package.

As far as downstream impact goes, in one of my more complex Symfony apps, this implementation blocks me from disabling the annotations integration at the framework level (this class requires an annotation reader, the serializer bundle registers this class as a service requiring the `annotation_reader` service, and the `annotation_reader` service is not enabled if the `framework.annotations.enabled` config is set to false).

goetas:
Hi, I think that it would be an improvement. do you think you can provide a PR?

mbabker:
I'd have to think of something that doesn't end up being a pain in the neck to maintain.

I ended up cheating in that app and copied the `JMS\Serializer\Metadata\Driver\AnnotationDriver` class into my app as `App\Serializer\Metadata\Driver\AttributeDriver`, removed the annotation reader argument, and just inlined all the attribute handling stuff instead of using a separate reader (and then overloaded the `jms_serializer.metadata_driver` service definition since that hardcodes the list of drivers for the chain instead of using service tags).

With B/C considerations, you can't drop implementing the interface from `JMS\Serializer\Metadata\Driver\AttributeDriver\AttributeReader`, so that change is out of the question.

Maybe something like this can work to create separate drivers for annotations and attributes without duplicating a lot of logic?

```php
namespace JMS\Serializer\Metadata\Driver;

use JMS\Serializer\Expression\CompilableExpressionEvaluatorInterface;
use JMS\Serializer\Naming\PropertyNamingStrategyInterface;
use JMS\Serializer\Type\Parser;
use JMS\Serializer\Type\ParserInterface;
use Metadata\ClassMetadata as BaseClassMetadata;
use Metadata\Driver\DriverInterface;

abstract class AnnotationOrAttributeDriver  implements DriverInterface
{
[...]

PR #1469, opened 2023-01-29
Create an attribute driver decoupled from the annotations driver, deprecate the annotation based attribute reader

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes/no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #1452
| License       | MIT

As suggested with https://github.com/schmittjoh/serializer/issues/1452#issuecomment-1377511748 this makes a new base driver class for annotations and attributes, moves the `loadMetadataForClass()` method to that base class for reuse by child objects, and declares methods for child classes to handle loading whatever's needed.  The current attribute reader (which implements the Annotations `Reader` interface) which is used as a decorator around an annotations reader for use inside the annotations driver is deprecated in favor of the new attributes driver.

(By design, the first commit here is a rename to try and retain the file history, otherwise doing it all in one go (rightfully) treats `AnnotationOrAttributeDriver.php` as a new file and the blame history gets really ugly; the commits individually show the moves right, even if GitHub's PR viewer doesn't)

### DIFF
--- a/src/Metadata/Driver/AnnotationDriver.php
+++ b/src/Metadata/Driver/AnnotationDriver.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Metadata\Driver;
+
+use Doctrine\Common\Annotations\Reader;
+use JMS\Serializer\Expression\CompilableExpressionEvaluatorInterface;
+use JMS\Serializer\Naming\PropertyNamingStrategyInterface;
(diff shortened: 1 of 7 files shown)


goetas:
whoa, this is cool :)

goetas:
this is nice :)

thanks!

PR #1471, opened 2023-02-05
[B/C Break] Make the Annotations package optional in the builder API

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | N/A
| License       | MIT

The below are the changes needed to (in theory) make the Annotations package an optional thing (note that Annotations is hard required by the dev PHPCR ODM and PHPBench dependencies, so right now it's not possible to actually not install the package without playing with Composer)

### DIFF
--- a/src/Builder/DefaultDriverFactory.php
+++ b/src/Builder/DefaultDriverFactory.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace JMS\Serializer\Builder;
 
 use Doctrine\Common\Annotations\Reader;
+use JMS\Serializer\Exception\RuntimeException;
 use JMS\Serializer\Expression\CompilableExpressionEvaluatorInterface;
 use JMS\Serializer\Metadata\Driver\AnnotationOrAttributeDriver;
 use JMS\Serializer\Metadata\Driver\DefaultValuePropertyDriver;
@@ -55,8 +56,12 @@ final class DefaultDriverFactory implements DriverFactoryInterface
         $this->enableEnumSupport = $enableEnumSupport;
     }
 
-    public function createDriver(array $metadataDirs, Reader $annotationReader): DriverInterface
+    public function createDriver(array $metadataDirs, ?Reader $annotationReader = null): DriverInterface
     {
+        if (PHP_VERSION_ID < 80000 && empty($metadataDirs) && !interface_exists(Reader::class)) {
+            throw new RuntimeException(sprintf('To use "%s", either a list of metadata directories must be provided, the "doctrine/annotations" package installed, or use PHP 8.0 or later.', self::class));
+        }
+
         /*
          * Build the sorted list of metadata drivers based on the environment. The final order should be:
          *
@@ -65,7 +70,11 @@ final class DefaultDriverFactory implements DriverFactoryInterface
          * - Annotations/Attributes Driver
          * - Null (Fallback) Driver
          */
-        $metadataDrivers = [new AnnotationOrAttributeDriver($this->propertyNamingStrategy, $this->typeParser, $this->expressionEvaluator, $annotationReader)];
+        $metadataDrivers = [];
+
+        if (PHP_VERSION_ID >= 80000 || $annotationReader instanceof Reader) {
+            $metadataDrivers[] = new AnnotationOrAttributeDriver($this->propertyNamingStrategy, $this->typeParser, $this->expressionEvaluator, $annotationReader);
+        }
 
         if (!empty($metadataDirs)) {
             $fileLocator = new FileLocator($metadataDirs);
--- a/src/Builder/DocBlockDriverFactory.php
+++ b/src/Builder/DocBlockDriverFactory.php
@@ -26,7 +26,7 @@ class DocBlockDriverFactory implements DriverFactoryInterface
         $this->typeParser = $typeParser;
     }
 
-    public function createDriver(array $metadataDirs, Reader $annotationReader): DriverInterface
+    public function createDriver(array $metadataDirs, ?Reader $annotationReader = null): DriverInterface
     {
         $driver = $this->driverFactoryToDecorate->createDriver($metadataDirs, $annotationReader);
 
--- a/src/Builder/DriverFactoryInterface.php
+++ b/src/Builder/DriverFactoryInterface.php
@@ -9,5 +9,5 @@ use Metadata\Driver\DriverInterface;
 
 interface DriverFactoryInterface
 {
-    public function createDriver(array $metadataDirs, Reader $annotationReader): DriverInterface;
+    public function createDriver(array $metadataDirs, ?Reader $annotationReader = null): DriverInterface;
 }
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,6 @@
     ],
     "require": {
         "php": "^7.4 || ^8.0",
-        "doctrine/annotations": "^1.14 || ^2.0",
         "doctrine/instantiator": "^1.3.1 || ^2.0",
         "doctrine/lexer": "^2.0 || ^3.0",
         "jms/metadata": "^2.6",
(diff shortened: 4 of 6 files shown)


src/Builder/DocBlockDriverFactory.php
```diff
@@ -26,7 +26,7 @@ public function __construct(DriverFactoryInterface $driverFactoryToDecorate, ?Pa
         $this->typeParser = $typeParser;
     }
 
-    public function createDriver(array $metadataDirs, Reader $annotationReader): DriverInterface
+    public function createDriver(array $metadataDirs, ?Reader $annotationReader = null): DriverInterface
```
mbabker, line 29: This one line is why there is a hard B/C break; widening the annotation reader argument to allow null causes compile errors in implementing classes.

src/Builder/DriverFactoryInterface.php
```diff
@@ -9,5 +9,5 @@
 
 interface DriverFactoryInterface
 {
-    public function createDriver(array $metadataDirs, Reader $annotationReader): DriverInterface;
+    public function createDriver(array $metadataDirs, ?Reader $annotationReader = null): DriverInterface;
```
mbabker, line 12: This one line is why there is a B/C break; widening the annotation reader argument to allow null causes compile errors in classes if they don't do the same.

scyzoryck: I think it should not be a big issues. JMS Bundle seems to not using those classes, so it will affect only people who is using package without Bundle.

src/Builder/DefaultDriverFactory.php
```diff
@@ -54,13 +55,21 @@ public function enableEnumSupport(bool $enableEnumSupport = true): void
...
+
+        $driver = new DriverChain();
+
         if (PHP_VERSION_ID >= 80000) {
-            $annotationReader = new AttributeDriver\AttributeReader($annotationReader);
+            $driver->addDriver(new AttributeDriver($this->propertyNamingStrategy, $this->typeParser, $this->expressionEvaluator));
         }
 
-        $driver = new AnnotationDriver($annotationReader, $this->propertyNamingStrategy, $this->typeParser);
+        if ($annotationReader) {
```
scyzoryck, line 70: I wonder is we should modify `ChainDriver` too?

scyzoryck (commented):
Thanks for contribution!

src/Builder/DefaultDriverFactory.php
mbabker: The instance at line 76 just outside this diff or something about the class itself?

There isn't a prepend method on the chain, so the implementation I have now ended up being the simplest one considering that when metadata paths are given, the factory pushes the YAML and XML drivers up in priority over the annotations driver (and the same goes for the default ordering in the bundle service definitions).

scyzoryck (approved):

scyzoryck:
Looks good for me. Personally I would release it after adding support for ORM 3.0. @goetas what do you think? Should we release it as a new major version?

kevinpapst:
Bump: one of the last major packages that still has a hard dependency on `doctrine/annotations`. 
Would be nice to get rid of it.
Anything I can do to help?

goetas (approved):

scyzoryck:
I will release it on Monday morning.

Steveb-p:
> I think it should not be a big issues. JMS Bundle seems to not using those classes, so it will affect only people who is using package without Bundle.

@scyzoryck We are, in fact, using it with bundle and replacing the implementation with our own. And given that we have stable releases with this serializer as a dependency, we will have projects of our partners breaking.

scyzoryck:
Hello! 
Thanks for feedback! Yes, we are aware about potential affect on some users - but expect that BC break will affect only few of them - with easy fix and easy way to fix that before deploying to production. I'm sorry that your work will be affected too. 

The package itself is not backed by any big companies - it is mainly voluntary work of few people with limited amount of free time. Unfortunately sometimes we need to make the some tough decisions to bring more exciting features and ensure long stability of the project. We always tries to deliver best value and features for others. If you found it useful for your project - it is always a good idea to support this project with contribution or funding :) 

Best!

Steveb-p:
@scyzoryck Understandable :sweat_smile: I kinda just wanted to note that it broke the software that we are giving to others. Given the broad experience spread in our little PHP world, it will take a little more or less for different developers.

I hope I didn't offend you, as my intent is only to note that this particular choice had some impact, nothing more :) I actually deleted a more playful part of the comment when writing, and text is rather bad at conveying intent :)

As for a potential resolution and/or fix for similar things happening in the future, you might consider going the same way as Symfony often does - by removing the argument from the interface altogether, and adding it as a comment. They do that often for things they deprecate.

Similarly maybe to how Symfony changed the argument order in `EventDispatcherInterface` a couple years ago:
https://github.com/nicolas-grekas/symfony/blob/75369dabb8af73b0d0ad7f206d85c08cf39117f8/src/Symfony/Component/EventDispatcher/EventDispatcherInterface.php#L32

scyzoryck:
Thanks for suggestion! I know Symfony approach - however it do not work so well when removing required dependencies - so it would not be super useful in that case 😓

Steveb-p:
Allow me to have a differing opinion on the matter, and respond. Note that those are my personal opinions and I fully value the approach you have taken (and work put into removing annotation dependency). I just don't agree with the notion of "there would always be some impact".

> ...Yeah, there was a chance it was going to impact someone, but there was no code change that wouldn't have.

With all due respect, I disagree with that. This could've been solved without breaking the implementations.

> Thanks for suggestion! I know Symfony approach - however it do not work so well when removing required dependencies - so it would not be super useful in that case 😓

> ...commenting the argument is generally Symfony's signal that a new argument is being added to the signature...

The example I've given is exactly when a required argument was being removed from the interface, and arguments were swapping places (event name argument becoming an optional one).

> ...Fully removing the argument would've introduced static analysis issues (passing more arguments than defined)...

Yes, but no. That could've been solved by adding a [stub file](https://phpstan.org/user-guide/stub-files), where the newer interface declaration could be used to give static analysis correct information about expected class shape. This has been dony in phpstan/phpstan-symfony as well, even specifically for the mentioned EventDispatcher changes.

> ...and Symfony has a history of treating moving dependencies from require to require-dev as soft breaks to avoid breaking apps relying on transient dependencies...

I agree that this is what often happens with Symfony, as those soft-breaks happened to us too. I would not have anything against annotation package removal, as this would indicate that a direct dependency was missing from one's library/project.

I'll be happy to provide you with static analysis compatible interface declaration (including the mentioned stub) if you're willing to take it into consideration.

ebitkov:
Can someone please explain how I re-enable annotation support for the moment, as long as some packages still switching to attributes? The documentation wasn't really clear about it.

Levivb:
@ebitkov I'de suggest requiring `doctrine/annotations` explicitly yourself in your `composer.json`. That way, the JMS Serializer will automatically pick up the `AnnotationReader`

boesing:
Any reason on why this was not a new major? I mean, this is not `boesing/typed-arrays` with 18k overall installs but `jms/serializer` which supposed to be **THE** serializer (along with the symfony one I guess) which has almost 100 million installs overall. 

IMHO there is responsibility for packages like this one and just "yah lets merge and put it in the changelog" is not what I was expecting from a project this size.

Just wanted to drop this here, might be helpful for future releases...

goetas:
As pointed out by @scyzoryck in https://github.com/schmittjoh/serializer/pull/1471#issuecomment-1963918460 this is a project run by volunteers working in their limited free time.

Releasing a new major version would have forced us to upgrade a few other projects such as fos rest, nelmio apidoc, Hateoas bundle... As you can see, it would force us to do a lot more work than what it seems.

I was aware that the changes were a possible BC break for some projects, but was the best trade off option I had in mind given the limited time we have. 

I'm sorry that the bc break impacted some users, hopefully tests were able to catch the issue and no disruption in production happened.

PR #1474, opened 2023-02-14
feat(exclusion): add new option "inheritedGroups" into GroupExclusionStrategy

Xen3r0:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

Questions :
- [ ] How can test attribute `inherited_groups` in Context.php ?
- [ ] Any help for my first contribution ?

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -72,7 +72,13 @@ abstract class Context
         $this->metadataStack = new \SplStack();
 
         if (isset($this->attributes['groups'])) {
-            $this->addExclusionStrategy(new GroupsExclusionStrategy($this->attributes['groups']));
+            $strategy = new GroupsExclusionStrategy($this->attributes['groups']);
+
+            if (isset($this->attributes['inherited_groups']) && is_bool($this->attributes['inherited_groups'])) {
+                $strategy->setInheritedGroups($this->attributes['inherited_groups']);
+            }
+
+            $this->addExclusionStrategy($strategy);
         }
 
         if (isset($this->attributes['version'])) {
@@ -200,6 +206,16 @@ abstract class Context
     public function enableMaxDepthChecks(): self
     {
         $this->attributes['max_depth_checks'] = true;
+
+        return $this;
+    }
+
+    /**
+     * @return $this
+     */
+    public function enableInheritGroups(): self
+    {
+        $this->attributes['inherit_groups'] = true;
 
         return $this;
     }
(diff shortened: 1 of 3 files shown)


src/Context.php
```diff
@@ -72,7 +72,13 @@ public function initialize(string $format, VisitorInterface $visitor, GraphNavig
         $this->metadataStack = new \SplStack();
 
         if (isset($this->attributes['groups'])) {
-            $this->addExclusionStrategy(new GroupsExclusionStrategy($this->attributes['groups']));
+            $strategy = new GroupsExclusionStrategy($this->attributes['groups']);
+
+            if (isset($this->attributes['inherited_groups']) && is_bool($this->attributes['inherited_groups'])) {
```
scyzoryck, line 77: As attributes are private, I guess we do not need to check if it is bool, as it should be enforced by the `enableInheritGroups()` method :)

scyzoryck (commented):
Thanks for your fist contribution! 
Could you also update documentation about exclusions about those changes, please? 
If you need any help, feel free to catch me!

src/Context.php
```diff
@@ -204,6 +210,16 @@ public function enableMaxDepthChecks(): self
         return $this;
     }
 
+    /**
+     * @return $this
+     */
+    public function enableInheritGroups(): self
+    {
+        $this->attributes['inherit_groups'] = true;
```
jdreesen, line 218: It is called `inherited_groups` above, but `inherit_groups` here. Is that right?

Xen3r0:
@scyzoryck @jdreesen can you review again please ?

src/Context.php
```diff
@@ -72,7 +72,13 @@ public function initialize(string $format, VisitorInterface $visitor, GraphNavig
         $this->metadataStack = new \SplStack();
 
         if (isset($this->attributes['groups'])) {
-            $this->addExclusionStrategy(new GroupsExclusionStrategy($this->attributes['groups']));
+            $strategy = new GroupsExclusionStrategy($this->attributes['groups']);
```
goetas, line 75: why not 

```php

$g = new GroupsExclusionStrategy(['a', 'b'], true);
$context->addExclusionStrategy($g);
```
?

In that way we do not need to add the `enableInheritGroups` method in the context. Ideally the context should know the less possible about the various exclusion strategies.

goetas:
I like this because it introduces back https://github.com/schmittjoh/serializer/pull/1071 in a backward compat way

src/Context.php
Xen3r0: Ok, i see, but i can't add 2 GroupsExclusionStrategy, one with inheritedGroups = false and an another one with inheritedGroups = true.

How can i configure this option with JMSerializerBundle for example ?

Thanks for you help

PR #1488, opened 2023-05-29
Allow doctrine instantiator 2.0

dgafka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": "^7.2||^8.0",
         "doctrine/annotations": "^1.13 || ^2.0",
-        "doctrine/instantiator": "^1.0.3",
+        "doctrine/instantiator": "^1.0.3|^2.0",
         "doctrine/lexer": "^1.1 || ^2",
         "jms/metadata": "^2.6",
         "phpstan/phpdoc-parser": "^0.4 || ^0.5 || ^1.0"
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": "^7.2||^8.0",
         "doctrine/annotations": "^1.13 || ^2.0",
-        "doctrine/instantiator": "^1.0.3|^2.0",
+        "doctrine/instantiator": "^1.0.3 || ^2.0",
         "doctrine/lexer": "^1.1 || ^2",
         "jms/metadata": "^2.6",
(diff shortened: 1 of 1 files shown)


dgafka:
[Doctrine instantiator was bumped to 2.0, in order to use PHP 8.1 features, so there are no B/C.](https://github.com/doctrine/instantiator/releases/tag/2.0.0)

composer.json
```diff
@@ -18,7 +18,7 @@
     "require": {
         "php": "^7.2||^8.0",
         "doctrine/annotations": "^1.13 || ^2.0",
-        "doctrine/instantiator": "^1.0.3",
+        "doctrine/instantiator": "^1.0.3|^2.0",
```
jdreesen, line 21:
```suggestion
        "doctrine/instantiator": "^1.0.3 || ^2.0",
```

dgafka: Is still valid syntax?

scyzoryck:
Thanks for contribution! 

I wonder what is a blocker to install it. In the past it was `doctrine/orm`, `doctrine/odm` but I think both already got support for instantiator 2.0 🤔 
https://github.com/schmittjoh/serializer/actions/runs/5112376250/jobs/9190336016?pr=1488
As in general it would be great to have running tests with it before merge :D

composer.json
jdreesen: Yes, both `|` and `||` are valid ([but according to the docs, the latter is the preferred variant](https://getcomposer.org/doc/articles/versions.md#version-range)). Also, everywhere else in the `require` section the `||` syntax is used, and I think we should prefer consistency.

mbabker:
> I wonder what is a blocker to install it.

https://github.com/doctrine/phpcr-odm/pull/844 was only merged to the 2.0 branch.  It just got a beta tag last week.  So unless that gets backported, there'd be no way of installing `doctrine/instantiator` 2.x alongside any `doctrine/phpcr-odm` 1.x release.

scyzoryck:
Ahm! True - I got lost in odm versions :D 

Tested locally with odm 2.0-dev and instantiator `2.0.0`. All looks good, tests are passing.

dgafka:
Thanks Marcin :)

PR #1491, opened 2023-06-02
Require Doctrine Lexer v2 and stop accessing Token properties via ArrayAccess to stop deprecation

ruudk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Doc updated   |no
| BC breaks?    |no 
| Deprecations? |no 
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT

By requiring `doctrine/lexer` v2 we can solve the following deprecation:
```
Accessing Doctrine\Common\Lexer\Token properties via ArrayAccess is deprecated, use the value, type or position property instead (Token.php:104 called by Parser.php:46, https://github.com/doctrine/lexer/pull/79, package doctrine/lexer)
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "php": "^7.2||^8.0",
         "doctrine/annotations": "^1.13 || ^2.0",
         "doctrine/instantiator": "^1.0.3 || ^2.0",
-        "doctrine/lexer": "^1.1 || ^2",
+        "doctrine/lexer": "^2",
         "jms/metadata": "^2.6",
         "phpstan/phpdoc-parser": "^0.4 || ^0.5 || ^1.0"
     },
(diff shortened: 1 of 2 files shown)


scyzoryck (approved):
Thanks for contribution!
The code looks good - I wonder if the dropping the support for `doctrine/lexer` might be a breaking changes. 🤔 I guess we need @goetas expertise to decide on it :)

mbabker:
> The code looks good - I wonder if the dropping the support for `doctrine/lexer` might be a breaking changes.

It should be fine.  The ORM dropped Lexer 1.x support in 2.15.0, and everything but this package in the main app I work in now requires at least 2.0.  So in practice, it's already getting to a point where Lexer 1.0 isn't installable with current dependencies.

ruudk:
Can we merge and tag it please? 💙

goetas (approved):
To me looks good. Require v2 only it is not a breaking change

goetas:
Can you please rebase the pull request?

PR #1495, opened 2023-06-13
Improve null checks in drivers with a delegate

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT

Split off from #1494

The metadata drivers with delegates inconsistently check for null responses, either the check is made after an `assert()` or it's not made at all.  So this PR ensures we have these null checks and early returns before asserting the metadata object is a specific subclass.


scyzoryck:
Thanks for contribution!

PR #1497, opened 2023-06-27
Only add the YamlDriver to the chain when `symfony/yaml` is installed

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT

The `JMS\Serializer\Metadata\Driver\YamlDriver` class requires `Symfony\Component\Yaml\Yaml` to function.  Given `symfony/yaml` is an optional dependency, it shouldn't be arbitrarily included in the driver chain when using the default driver factory and specifying a list of metadata paths.  This PR updates the factory to check for the `Yaml` class before adding the YAML driver to the chain.

### DIFF
--- a/src/Builder/DefaultDriverFactory.php
+++ b/src/Builder/DefaultDriverFactory.php
@@ -19,6 +19,7 @@ use JMS\Serializer\Type\ParserInterface;
 use Metadata\Driver\DriverChain;
 use Metadata\Driver\DriverInterface;
 use Metadata\Driver\FileLocator;
+use Symfony\Component\Yaml\Yaml;
 
 final class DefaultDriverFactory implements DriverFactoryInterface
 {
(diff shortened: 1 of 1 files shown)


tests/Fixtures/AuthorDeprecatedReadOnly.php
```diff
@@ -19,9 +19,10 @@
 class AuthorDeprecatedReadOnly
 {
     /**
-     * @JMS\Serializer\Annotation\ReadOnly
```
scyzoryck, line 22: If I remember correctly the test was setup to avoid issues with not working `ReadOnly` annotation. Is it possible to keep that annotation?

tests/Fixtures/AuthorList.php
```diff
@@ -74,8 +71,7 @@ public function offsetSet($offset, $value)
     /**
      * @see ArrayAccess
      */
-    #[\ReturnTypeWillChange]
-    public function offsetUnset($offset)
+    public function offsetUnset($offset): void
```
scyzoryck, line 74: 👍

tests/Fixtures/AuthorDeprecatedReadOnly.php
mbabker: This comment fits better on #1494 as that's where the change is coming from (as I said, I based this branch on that one for my own sanity 😆).  But, I was running into issues running the tests locally on PHP 8.2 with the aliased class name.  I can try to put that back and see what happens.

scyzoryck (approved):

src/Metadata/Driver/AnnotationDriver.php
```diff
@@ -18,7 +18,7 @@ class AnnotationDriver extends AnnotationOrAttributeDriver
 
     public function __construct(Reader $reader, PropertyNamingStrategyInterface $namingStrategy, ?ParserInterface $typeParser = null, ?CompilableExpressionEvaluatorInterface $expressionEvaluator = null)
     {
-        parent::__construct($namingStrategy, $typeParser, $expressionEvaluator);
+        parent::__construct($namingStrategy, $typeParser, $expressionEvaluator, $reader);
```
goetas, line 21: should this class be deprecated in favor of `AnnotationOrAttributeDriver`?

src/Metadata/Driver/NullDriver.php
```diff
@@ -5,11 +5,23 @@
...
 use Metadata\Driver\DriverInterface;
 
 class NullDriver implements DriverInterface
 {
+    /**
+     * @var PropertyNamingStrategyInterface
+     */
+    private $namingStrategy;
+
+    public function __construct(PropertyNamingStrategyInterface $namingStrategy)
```
goetas, line 20: is this a bc break? (new required param)

goetas: you are using this because now the `AnnotationOrAttributeDriver` can return null (becase of the  `$configured` thing...)

is it worth? what is AnnotationOrAttributeDriver returns the empty metadata as before?

PR #1501, opened 2023-07-26
fix(builder): enableEnumSupport should return self

Anticom:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes/no
| Fixed tickets | -
| License       | MIT

Not sure whether this should be classified as fix or feature.

With this PR the `SerializerBuilder::enableEnumSupport` method now returns `$this` to retain the fluent interface.

### DIFF
--- a/src/SerializerBuilder.php
+++ b/src/SerializerBuilder.php
@@ -524,13 +524,15 @@ final class SerializerBuilder
         return $this;
     }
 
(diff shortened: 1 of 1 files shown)


Anticom:
Relates to #1448

PR #1505, opened 2023-08-03
Fix CS issues

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | not yet
| Fixed tickets | N/A
| License       | MIT

Fixes CS issues I managed to introduce in #1494 that only became evident when combined with #1503.

High CI is still expected to fail due to changes in ORM 2.16.


scyzoryck:
Thanks! :)

PR #1506, opened 2023-08-03
do not return null when no annotations or attributes are found by the…

goetas:
This reverts a small part of https://github.com/schmittjoh/serializer/pull/1494 to minimize the BC break introduced by it.

The plan is to add the NullDriver in the chain of the bundle, and then put back this line


mbabker (approved):

goetas:
I'm merging this change anyway, the test failure seems unrelated

Issue #1512, opened 2023-09-12
Minor Issue from Triggering Autoload Before Expanding Class Names


rossriley:
| Q                | A
| ---------------- | -----
| Bug report?      | yes (possibly)
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Appreciate this is probably an edge case but others may have run into it too. If you use an interactive shell to interact with deserialization via the `DocBlockTypeResolver` the class will try to autoload before it has been properly expanded via the `use` statements within the class.

## Steps required to reproduce the problem

With this simplified version of the setup:

```
<?php

namespace ExampleAPI\Response;

use ExampleAPI\Models\Product;

class ProductCollection
{
    /** @var Product[] */
    public array $data;
}
```

## Expected Result

After deserializing it would be expected that the resolved type would include the `use` statement but the autoloader in PsySH will scan the project for other classes named `Product` and will possibly select the wrong class.

## Possible change

The logic to resolve is here: https://github.com/schmittjoh/serializer/blob/master/src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php#L250

The first few lines use `class_exists('Product')` rather than expanding out from the use statement first. My question is, whilst most people won't run into this problem wouldn't it make sense to expand the class from the use statements first and call `class_exists('ExampleAPI\Models\Product')` first and then fallback to the non namespaced class if the first is not successful?

Happy to PR if you agree.

scyzoryck:
Hi! 
Looking at your explanation - it looks like a bug (edgecase but still bug). Feel free to create an Pull Request with the fix. :) let me know if you'd need any help with preparing the fix. 

Best scyzoryck

goetas:
Fixed in https://github.com/schmittjoh/serializer/pull/1513

PR #1513, opened 2023-09-14
Fix - change the order of instantiation checks

rossriley:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1512
| License       | MIT

Fix for issue described in #1512 all this does is rearrange the resolution logic so when a type is described as eg:

```
@var Product
```

The resolution order now will be:

1. Look for a class in the current namespace called `Product`
2. Parse the `use` statements and look for a class defined as `Namespace\Example\Product`
3. Finally fallback to a class or interface defined in the global namespace `\Product`

Currently steps 1 and 3 are reversed and that effectively means that a precisely defined class is ignored in favour of a more loosely specified one.

### DIFF
--- a/src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php
+++ b/src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php
@@ -249,10 +249,6 @@ final class DocBlockTypeResolver
      */
     private function expandClassNameUsingUseStatements(string $typeHint, \ReflectionClass $declaringClass, $reflector): string
     {
-        if ($this->isClassOrInterface($typeHint)) {
-            return $typeHint;
-        }
-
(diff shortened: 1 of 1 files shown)


scyzoryck:
Just in case `Coding Standards / Coding Standards (7.2) (pull_request)` - fails also on master, so it will be fixed in separate PR :) 
Could you also add some test to make sure that this logic will be not lost in the future? 

Best, scyzoryck.

goetas:
@dgafka  can you please have  look here?

goetas:
@rossriley could you please add a test?

dgafka:
Ye, looks like an edge case with fine fix :)

dgafka (approved):

PR #1519, opened 2023-10-13
Bump minimum dependencies to match what's actually tested

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT

This just bumps the minimum dependency versions in `composer.json` to match the actual versions installed on the PHP 7.2 lowest deps build (https://github.com/schmittjoh/serializer/actions/runs/6449612330/job/17508114478) and normalizes the file's formatting.  The only one that might have a practical impact on folks is if they use `phpstan/phpdoc-parser` at an older release as it's probably constrained by the dev dependency chain here in ways the production dependencies won't be, otherwise a lot of this stuff is somewhat old anyway and with upcoming major releases from a number of tools, the wider ranges eventually start causing extra maintenance burden.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,12 +16,12 @@
         }
     ],
     "require": {
-        "php": "^7.2||^8.0",
-        "doctrine/annotations": "^1.13 || ^2.0",
-        "doctrine/instantiator": "^1.0.3 || ^2.0",
+        "php": "^7.2 || ^8.0",
+        "doctrine/annotations": "^1.14 || ^2.0",
+        "doctrine/instantiator": "^1.3.1 || ^2.0",
         "doctrine/lexer": "^2.0 || ^3.0",
         "jms/metadata": "^2.6",
-        "phpstan/phpdoc-parser": "^0.4 || ^0.5 || ^1.0"
+        "phpstan/phpdoc-parser": "^1.20"
     },
     "suggest": {
         "doctrine/collections": "Required if you like to use doctrine collection types as ArrayCollection.",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -16,12 +16,12 @@
...
     "require": {
-        "php": "^7.2||^8.0",
-        "doctrine/annotations": "^1.13 || ^2.0",
-        "doctrine/instantiator": "^1.0.3 || ^2.0",
+        "php": "^7.2 || ^8.0",
+        "doctrine/annotations": "^1.14 || ^2.0",
+        "doctrine/instantiator": "^1.3.1 || ^2.0",
         "doctrine/lexer": "^2.0 || ^3.0",
         "jms/metadata": "^2.6",
-        "phpstan/phpdoc-parser": "^0.4 || ^0.5 || ^1.0"
```
scyzoryck, line 24: Sound like a bit big issue - I guess not so many people use it as main dependency.

scyzoryck (approved):

composer.json
mbabker: I took a glance at https://packagist.org/packages/phpstan/phpdoc-parser/stats just to see what the impact might be.  [0.4.9's](https://packagist.org/packages/phpstan/phpdoc-parser/stats#0.4.9) a huge outlier and getting a few thousand installs per month on its own (and accounting for most of the 0.x installs in general), but if my math's right, 0.x is only accounting for a little under 3% of all recent installs.  For the 1.x installs, approx. 56% are grabbing the latest minor version, and it looks like approx. 86% are grabbing something matching `^1.20`.  So I don't think this is necessarily as impactful as we might've thought just looking at version ranges.

Issue #1516, opened 2023-10-11
ORM 3.0 Compat


mbabker:
The ORM now has a [3.0 beta release](https://github.com/doctrine/orm/releases/tag/3.0.0-beta1), so compat should be looked at since it's a bit of a bigger release.

I did manage to get it to install with this package after removing the PHPCR-ODM since its dependencies effectively block installing ORM 3.0, and ran the tests.

<details>

<summary>Test Output</summary>

```sh
/usr/local/bin/php "/serializer/vendor/phpunit/phpunit/phpunit" --configuration "/serializer/phpunit.xml.dist" --teamcity
Testing started at 10:49 AM ...
PHPUnit 10.5-dev by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.7
Configuration: /serializer/phpunit.xml.dist


Attribute driver is required.

Attribute driver is required.

Can not define annotation or attribute metadata for internal classes

Short expose syntax not supported on annotations or attribute

Attribute driver is required.

Attribute driver is required.

Can not define annotation or attribute metadata for internal classes

Short expose syntax not supported on annotations or attribute

Error: Class "Doctrine\ORM\Version" not found
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:30


Error: Class "Doctrine\ORM\Mapping\Driver\AnnotationDriver" not found
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:138
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:159
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:25
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:41


Error: Class "Doctrine\ORM\Mapping\Driver\AnnotationDriver" not found
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:138
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:159
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:25
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:51


Error: Class "Doctrine\ORM\Mapping\Driver\AnnotationDriver" not found
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:138
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:159
/serializer/tests/Metadata/Driver/DoctrineDriverTest.php:25
[...]

PR #1520, opened 2023-10-17
Inject an attribute metadata driver into the DoctrineDriver when testing

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | Partially #1516
| License       | MIT

Another small test change that will help with the ORM 3.0 compat.  Without its classes having the `@Annotation` annotation anymore in 3.0, there are a lot of `Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The class "Doctrine\ORM\Mapping\Entity" is not annotated with @Annotation.` style exceptions being thrown.  Changing the `DoctrineDriverTest` to have version conditional behavior and use an annotation or attribute driver as appropriate clears out some of those.

Merging #1471 will help address this in other places in the tests at a later point, but with the hard dependency still in place, those other test classes are not yet updated.


scyzoryck (approved):
Looks good for me! I will merge it on the beginning of the next week with other related PRs. :)

PR #1526, opened 2023-12-04
Allow missing attributes in serialized objects

### LINKED ISSUE #1525: Serialization of objects with missing attribute errors
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| BC Break report? | yes

With 3.28.0, something has changed so that we now end up having an error at runtime due to missing attributes.
We do use OpenAPI attributes to generate API documentation during CI and since we do not need these attributes on production, we do have these in `require-dev` section of composer.
That worked pretty well until 3.27.0 and stopped working with 3.28.0.

The problem is, that there is no check if an attribute actually exists:

https://github.com/schmittjoh/serializer/blob/f24027583f8ad3c1789bff412b039013132316ac/src/Metadata/Driver/AnnotationOrAttributeDriver.php#L331C3-L336

We had something similar with `cuyz/valinor` (https://github.com/CuyZ/Valinor/issues/393), but that was only related to cache files and was not affecting runtime (without caches). 
That was due to the fact that `cuyz/valinor` already kinda handled missing attributes:

https://github.com/romm/Valinor/blob/af98753ba93c2a1ff8b717a5ecfbe3424a41b99f/src/Definition/NativeAttributes.php#L34-L41

1. Create an object to serialize
2. Add an attribute to the class or to a property which does not exist (I've used `#[Test]`)
3. Serialize that object

## Expected Result

* No error is sthrown

## Actual Result

* `Error: Attribute class "Application\Test\TestAsset\Test" not found`

boesing:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Tests pass?   | yes
| Fixed tickets | #1525
| License       | MIT

Fixes #1525 which points to some kind of BC break in v3.28.0 regarding missing attributes.

### DIFF
--- a/phpstan/ignore-by-php-version.neon.php
+++ b/phpstan/ignore-by-php-version.neon.php
@@ -10,7 +10,7 @@ if (PHP_VERSION_ID < 80100) {
     $includes[] = __DIR__ . '/no-enum.neon';
 }
 if (PHP_VERSION_ID >= 80000) {
-    $includes[] = __DIR__ . '/ignore-missing-attributes.neon';
+    $includes[] = __DIR__ . '/ignore-missing-attribute.neon';
 }
 if (PHP_VERSION_ID >= 80100 && PHP_VERSION_ID < 80200) {
     $includes[] = __DIR__ . '/php-81.neon';
--- a/phpstan/ignore-by-php-version.neon.php
+++ b/phpstan/ignore-by-php-version.neon.php
@@ -9,6 +9,9 @@ if (PHP_VERSION_ID < 80000) {
 if (PHP_VERSION_ID < 80100) {
     $includes[] = __DIR__ . '/no-enum.neon';
 }
+if (PHP_VERSION_ID >= 80000) {
+    $includes[] = __DIR__ . '/ignore-missing-attributes.neon';
+}
 if (PHP_VERSION_ID >= 80100 && PHP_VERSION_ID < 80200) {
     $includes[] = __DIR__ . '/php-81.neon';
 }
--- a/src/Metadata/Driver/AnnotationOrAttributeDriver.php
+++ b/src/Metadata/Driver/AnnotationOrAttributeDriver.php
@@ -48,6 +48,8 @@ use Metadata\ClassMetadata as BaseClassMetadata;
 use Metadata\Driver\DriverInterface;
 use Metadata\MethodMetadata;
 
+use function array_filter;
+
 class AnnotationOrAttributeDriver implements DriverInterface
 {
     use ExpressionMetadataTrait;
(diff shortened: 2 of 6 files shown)


src/Metadata/Driver/AnnotationOrAttributeDriver.php
```diff
@@ -309,7 +311,12 @@ protected function getClassAnnotations(\ReflectionClass $class): array
                 static function (\ReflectionAttribute $attribute): object {
                     return $attribute->newInstance();
                 },
-                $class->getAttributes()
+                array_filter(
+                    $class->getAttributes(),
+                    static function (\ReflectionAttribute $attribute): bool {
+                        return class_exists($attribute->getName());
+                    }
+                )
```
mbabker, line 319: The same filtering is needed in this class' `getMethodAnnotations()` and `getPropertyAnnotations()` methods.

boesing: Ah good catch, I've seen that in the `AttributeDriver` but havent adapted in this one.

mbabker (approved):

phpstan.neon.dist
```diff
@@ -21,6 +21,7 @@ parameters:
         - '#^Property JMS\\Serializer\\Tests\\Fixtures\\DocBlockType\\Collection\\CollectionOfClassesWithFullNamespacePath\:\:\$productIds has unknown class JMS\\Serializer\\Tests\\Fixtures\\DocBlockType\\Collection\\JMS\\Serializer\\Tests\\Fixtures\\DocBlockType\\Collection\\Product as its type\.$#'
         - '#^Property JMS\\Serializer\\Tests\\Fixtures\\DocBlockType\\Collection\\CollectionOfInterfacesWithFullNamespacePath\:\:\$productColors has unknown class JMS\\Serializer\\Tests\\Fixtures\\DocBlockType\\Collection\\JMS\\Serializer\\Tests\\Fixtures\\DocBlockType\\Collection\\Details\\ProductColor as its type\.$#'
         - '#^Method JMS\\Serializer\\GraphNavigator\\DeserializationGraphNavigator\:\:resolveMetadata\(\) should return JMS\\Serializer\\Metadata\\ClassMetadata\|null#'
+        - '#^Attribute class JMS\\Serializer\\Tests\\Fixtures\\MissingAttribute does not exist\.$#'
```
scyzoryck, line 24: I think it needs their own file under `phpstan` directory. Plus add logic into `phpstan/ignore-by-php-version.neon.php` that will include this file for some PHP versions.

scyzoryck (commented):
Thanks for contribution! :) Let me know if you need any help with PHPStan issues!. 

Best!

phpstan/ignore-by-php-version.neon.php
```diff
@@ -9,6 +9,9 @@
 if (PHP_VERSION_ID < 80100) {
     $includes[] = __DIR__ . '/no-enum.neon';
 }
+if (PHP_VERSION_ID >= 80000) {
+    $includes[] = __DIR__ . '/ignore-missing-attributes.neon';
```
scyzoryck, line 13:
```suggestion
    $includes[] = __DIR__ . '/ignore-missing-attribute.neon';
```

scyzoryck:
Thanks for contribution! :) 
I will release it on the beginning of the next week.

PR #1536, opened 2024-02-06
Fix serialization of BackedEnum - annotation without name/value 

iKsSs:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

Serialization of Backed Enum, when property has Serializer annotation with specified Enum Class but second parameter (name/value), is not provided was not behaving according to documentation.

### DIFF
--- a/src/Handler/EnumHandler.php
+++ b/src/Handler/EnumHandler.php
@@ -44,7 +44,7 @@ final class EnumHandler implements SubscribingHandlerInterface
         array $type,
         SerializationContext $context
     ) {
-        if (isset($type['params'][1]) && 'value' === $type['params'][1]) {
+        if ((isset($type['params'][1]) && 'value' === $type['params'][1]) || (!isset($type['params'][1]) && is_a($enum, \BackedEnum::class, true))) {
             if (!$enum instanceof \BackedEnum) {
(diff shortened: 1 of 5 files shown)


scyzoryck:
@goetas could you help with review, please? :)

goetas:
Looks good to me. I'm surprised that i forgot to implement it from get go. 

i have left only one minor comment

src/Handler/EnumHandler.php
```diff
@@ -44,7 +44,7 @@ public function serializeEnum(
         array $type,
         SerializationContext $context
     ) {
-        if (isset($type['params'][1]) && 'value' === $type['params'][1]) {
+        if ((isset($type['params'][1]) && 'value' === $type['params'][1]) || (!isset($type['params'][1]) && is_a($enum, \BackedEnum::class, true))) {
```
goetas, line 47: Can `is_a($enum, \BackedEnum::class, true)` be replace with `$enum instanceof \BackedEnum`? seems much more readable to me.

scyzoryck (approved):

Issue #1502, opened 2023-07-26
Question: How to handle de-/serializing unions (using native type hints)?


Anticom:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | no

cc: @dgafka

According to #1330 we should have some capabilities of handling union types.

1. Are those capabilities restricted exclusively to DocBlock annotations?
2. If 1. is true, how would the below snippet have to be defined in order to work?
   ```php
   class ClassWithUnionType
   {
       /**
        * @var string|ComplexType[] $propertyUnderTest
        */
       public string|array $propertyUnderTest;
   }
   ```
   Won't work either.
3. Is implementing a custom `UnionHandler` the only/best way to get this working?
   I could imagine that especially deserialization is tricky since some pattern matching would have to be done in order to infer the actual desired type. Even then ambiguities will remain an issue if complex type definitions overlap.  
 
I thought about implementing a custom handler for this, supporting only two arguments as a union and restricting the first argument to a primitive type to mitigate those pattern matching issues.  
Would you be interested in a contribution to this project for such a restricted union support?

## Steps required to reproduce the problem

```php
class ComplexType
{
    public string $foo;
    public string $bar;
}

class ClassWithUnionType
{
    #[\JMS\Serializer\Annotation\Type('string|array<ComplexType>')]
    public string|array $propertyUnderTest;
}

$serializer = \JMS\Serializer\SerializerBuilder::create()->build();

// serialization
$simpleObject = new ClassWithUnionType();
$simpleObject->propertyUnderTest = 'foo';

$simpleJson = $serializer->serialize($simpleObject, 'json');
var_dump($simpleJson);

// ---

$complexType = new ComplexType();
$complexType->foo = 'foo';
$complexType->bar = 'bar';

$complexObject = new ClassWithUnionType();
$complexObject->propertyUnderTest = [
    $complexType,
];

$complexJson = $serializer->serialize($complexObject, 'json');
var_dump($complexJson);

// deserialization
$simpleJson = <<<JSON
{
    "property_under_test": "foo"
}
JSON;

$simpleObject = $serializer->deserialize($simpleJson, ClassWithUnionType::class, 'json');
var_dump($simpleObject);

// ---

$complexJson = <<<JSON
{
    "property_under_test": [
        {
            "foo": "foo",
            "bar": "bar"
        }
    ]
}
JSON;

[...]

PR #1546, opened 2024-07-03
Union types deserialisation

idbentley:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1502
| License       | MIT

This is just a minor improvement on https://github.com/schmittjoh/serializer/pull/1504 - addressing @goetas feedback, improving type detection and tests.


scyzoryck:
Thanks for taking up this topic. Unfortunately I'm super short in time recently :( 

It looks like there was few issues found by the CI pipeline like: 
> Error: Class 'JMS\Serializer\Tests\Fixtures\DocBlockType\UnionTypedDocBlockProperty' not found

Also, please use `vendor/bin/phpcbf` to format the code :)

idbentley:
@scyzoryck I completely understand.  I need union support for a project I'm working on, so I'm happy to push this forward.  Thanks for taking the time to work with me.

If you can kickoff the workflows again (I don't seem to be able to), I think I've fixed those issues.

idbentley:
I will look into these phpstan issues tomorrow.

idbentley:
Hey @scyzoryck I addressed the `phpstan` issues (sorry, I'm embarrassed I didn't pay more attention to the contributor guide).

However, there were a number of phpstan related issues on a file that is not touched in this pull request: `src/Metadata/Driver/DoctrineTypeDriver.php`.  I can address those too, but I left them out because they seem unrelated to the other code changes.  Let me know if you'd like me to apply that patch before we run through CI again.

idbentley:
Hiya!

I added a PR to only register the UnionHandler on PHPVersion > 8.

Not sure how to get the PHPStan to ignore the `|` usage in UnionHandler though.  It's only used in the return type of the `determineType` function - as a test I tried to use `?string` instead of `string|null`?

It's not clear to me what the `Error: Ignored error pattern #Constructor of class JMS\\Serializer\\Annotation\\.*? has an unused parameter# was not matched in reported errors.` error reported by PHPStan means (that error does appear for me locally in master).

src/Handler/UnionHandler.php
```diff
@@ -0,0 +1,134 @@
...
+     *  ReflectionUnionType::getTypes() returns the types sorted according to these rules:
+     * - Classes, interfaces, traits, iterable (replaced by Traversable), ReflectionIntersectionType objects, parent and self:
+     *     these types will be returned first, in the order in which they were declared.
+     * - static and all built-in types (iterable replaced by array) will come next. They will always be returned in this order:
+     *     static, callable, array, string, int, float, bool (or false or true), null.
+     *
+     * For determining types of primitives, it is necessary to reorder primitives so that they are tested from lowest specificity to highest:
+     * i.e. null, true, false, int, float, bool, string
+     */
+    private function reorderTypes(array $type): array
```
scyzoryck, line 89: Let's move it into `TypedPropertiesDriver`. In that way it will be reorder once & cached, not for each deserialisation.

scyzoryck (changes requested):
Hi! In general it looks good for me as a basic setup. I left one comment to improve performance. 

No need to fix `src/Metadata/Driver/DoctrineTypeDriver.php` in this PR - it requires separate PR. :)

src/Handler/UnionHandler.php
idbentley: Good idea.

idbentley:
Seems like the remaining CI issues are in the `DoctrineTypeDriver.php`.  Let me know if you need anything else from me!

src/Metadata/Driver/TypedPropertiesDriver.php
```diff
@@ -47,6 +47,29 @@ public function __construct(DriverInterface $delegate, ?ParserInterface $typePar
...
+     * For determining types of primitives, it is necessary to reorder primitives so that they are tested from lowest specificity to highest:
+     * i.e. null, true, false, int, float, bool, string
+     */
+    private function reorderTypes(array $type): array
+    {
+        if ($type['params']) {
+            uasort($type['params'], static function ($a, $b) {
+                $order = ['null' => 0, 'true' => 1, 'false' => 2, 'bool' => 3, 'int' => 4, 'float' => 5, 'string' => 6];
+
+                return (array_key_exists($a['name'], $order) ? $order[$a['name']] : 7) <=> (array_key_exists($b['name'], $order) ? $order[$b['name']] : 7);
```
scyzoryck, line 66: I guess something like:
```suggestion
                return $order[$a['name']] ?? 7 <=> $order[$b['name']] ?? 7;
```
would be a bit simpler to understand. :D

scyzoryck (approved):

src/Metadata/Driver/TypedPropertiesDriver.php
scyzoryck: Looks like all comments from the previous review was addressed. 
@idbentley Let me now if it makes sense after that I will merge it to master :)

scyzoryck:
@idbentley thanks for contribution! You are welcome with the next Pull Requests! :) Could you test it with your projects & dev version of the package, please? :) 

Best, Marcin

goetas:
it would be great to have this handler integrated in the bundle as well

idbentley:
@goetas I'm currently working on a PR to expand this functionality to support Objects, arrays, non primitive data types.  I thought it would make sense to get it into documentation and any build targets once that PR is ready as well.

goetas:
@idbentley  I have talked a little with @scyzoryck  about it, deserializing objects used in unions is certainly a good thing to have but I do not have any idea on how to make it work.... how to decide to which class to map an arbitrary data structure? do you have any idea on how to do so?

simPod:
@idbentley this crashes:

image

It calls MetadataFactory::getClassHierarchy() with 'union' input.

calling `serialize` with instance of

```php
final class Foo
{
    public function __construct(
        public A|B $union,
    ) {
```

idbentley:
@simPod that's definitely a bad way of handling the issue, and I'm gonna look into more graceful failure.  However, I think your main problem is that there is currently only support for unions of primitives at this time.  See #1549 to follow progress on my attempt at unions of objects.

simPod:
Yup, it should ignore unsupported cases. Good luck!

PR #1548, opened 2024-07-16
refactor(driver): use instanceof checks for correct phpdoc node

simPod:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes/no
| Fixed tickets | #... 
| License       | MIT

### DIFF
--- a/src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php
+++ b/src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php
@@ -7,6 +7,7 @@ namespace JMS\Serializer\Metadata\Driver\DocBlockDriver;
 use PHPStan\PhpDocParser\Ast\PhpDoc\ParamTagValueNode;
 use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
 use PHPStan\PhpDocParser\Ast\PhpDoc\ReturnTagValueNode;
+use PHPStan\PhpDocParser\Ast\PhpDoc\TypeAliasTagValueNode;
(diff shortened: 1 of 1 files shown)


src/Metadata/Driver/DocBlockDriver/DocBlockTypeResolver.php
```diff
@@ -408,15 +409,15 @@ private function getPhpstanType(\ReflectionClass $declaringClass, string $typeHi
         $self = $this;
 
         foreach ($phpDocNode->children as $node) {
-            if ($node instanceof PhpDocTagNode && '@phpstan-type' === $node->name) {
-                $phpstanType = (string) $node->value;
+            if ($node instanceof PhpDocTagNode && $node->value instanceof TypeAliasTagValueNode) {
+                $phpstanType = $node->value->__toString();
                 preg_match_all(self::PHPSTAN_ARRAY_SHAPE, $phpstanType, $foundPhpstanArray);
-                if (isset($foundPhpstanArray[1][0]) && $foundPhpstanArray[1][0] === $typeHint) {
+                if (isset($foundPhpstanArray[0][0]) && $node->value->alias === $typeHint) {
```
scyzoryck, line 415: Just thinking:
It looks like `$node->value->alias === $typeHint` can be moved to condition in line 412. It looks like we can also use  `preg_match` instead of `preg_match_all`. 🤔 WDYT?

simPod: Yes, done.

PR #1550, opened 2024-07-18
Update JsonSerializationVisitor.php - Make 'skipWhenEmpty' work with all types not just \ArrayObject or array, using empty(..) to decide skipping attribute.

msklenica:
Make 'skipWhenEmpty' work with all types not just \ArrayObject or array, using empty(..) to decide skipping attribute.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | 
| License       | MIT

### DIFF
--- a/src/JsonSerializationVisitor.php
+++ b/src/JsonSerializationVisitor.php
@@ -147,7 +147,7 @@ final class JsonSerializationVisitor extends AbstractVisitor implements Serializ
             return;
         }
 
-        if (true === $metadata->skipWhenEmpty && ($v instanceof \ArrayObject || \is_array($v)) && 0 === count($v)) {
+        if (true === $metadata->skipWhenEmpty && empty($v)) {
             return;
         }
 


src/JsonSerializationVisitor.php
```diff
@@ -147,7 +147,7 @@ public function visitProperty(PropertyMetadata $metadata, $v): void
             return;
         }
 
-        if (true === $metadata->skipWhenEmpty && ($v instanceof \ArrayObject || \is_array($v)) && 0 === count($v)) {
+        if (true === $metadata->skipWhenEmpty && empty($v)) {
```
scyzoryck, line 150: It is not going to work with `ArrayObject`  as `empty` will always return `false` for any object. Try: `var_dump(empty(new \ArrayObject()));`

scyzoryck (commented):
@msklenica could you explain a bit the idea behind this PR, please? It looks like the proposed changes will brake the current logic. 

Best, Marcin.

src/JsonSerializationVisitor.php
msklenica: Thank you @scyzoryck for checking it out! You are right, empty is not suitable on it's own,

What would you think about something like this:

`if (
    true === $metadata->skipWhenEmpty &&
    ((!is_object($v) && empty($v)) ||
        ($v instanceof \ArrayObject && 0 === count($v)))
) {`

PR #1552, opened 2024-07-19
feat: support unions in arrays

simPod:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT


scyzoryck:
@simPod I think we need also E2E test of deserialisation - I think it should already work for simple types :)

idbentley:
My main feedback is that this PR seems rather incomplete.  

The changes to the `@Type` lexer/parser may support the simplest `array<A|B>` test case, but I can think of some edge cases that would cause that type parsing to fail.  The `@Type` typer lexer/parser currently has no union support, so this adds support for `array<A|B>` but there's no support for just `A|B`.

The only tests seem to be of the docblockparser (which is good), but no tests of any serialization/deserialization, no tests of the Type annotation lexer changes.

What is the expected behaviour of `A|B[]` in a docblock type?

Overall, I think this is a great feature, but needs some extra help.

simPod:
This is indeed rather incomplete. I'm now stuck with DeserializationGraphNavigator that does not know how to handle unions.

tests/Metadata/Driver/DocBlockDriverTest.php
```diff
@@ -131,6 +131,37 @@ public function testInferDocBlockMapFromGenericLikeClass()
...
+            ],
+            $m->propertyMetadata['productOrVehicleIds']->type,
+        );
+        self::assertEquals(
+            [
+                'name' => 'array',
+                'params' => [
+                    ['name' => 'int', 'params' => []],
+                    ['name' => Product::class, 'params' => []],
+                    ['name' => Vehicle::class, 'params' => []],
```
scyzoryck, line 150: @simPod It seems like `array<int, string>` and `array<int|string>` will create exactly the same metadata 😓 I would consider using `Union` pseudo type same as for the unions, so it will be handled by the Union Handler :D

PR #1553, opened 2024-07-26
Allow union deserialization using discriminator map

idbentley:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1549
| License       | MIT

Hi @goetas here's a simpler version of PR #1549 - this one only adds support for discriminated complex unions.

### DIFF
--- a/src/Handler/UnionHandler.php
+++ b/src/Handler/UnionHandler.php
@@ -67,44 +67,29 @@ final class UnionHandler implements SubscribingHandlerInterface
             throw new RuntimeException('XML deserialisation into union types is not supported yet.');
         }
 
-        $finalType = null;
-        if (2 === count($type['params'])) {
-            if (!is_array($type['params'][0]) || !array_key_exists('name', $type['params'][0])) {
-                $lookupField = $type['params'][0];
-                $unionMap = $type['params'][1];
-
-                if (!array_key_exists($lookupField, $data)) {
-                    throw new NonVisitableTypeException('Union Discriminator Field \'' . $lookupField . '\' not found in data');
-                }
-
-                $lkup = $data[$lookupField];
-                if (!empty($unionMap)) {
-                    if (array_key_exists($lkup, $unionMap)) {
-                        $finalType = [
-                            'name' => $unionMap[$lkup],
-                            'params' => [],
-                        ];
-                    } else {
-                        throw new NonVisitableTypeException('Union Discriminator Map does not contain key \'' . $lkup . '\'');
-                    }
-                } else {
-                    $finalType = [
-                        'name' => $lkup,
-                        'params' => [],
-                    ];
-                }
+        if (3 === count($type['params'])) {
+            $lookupField = $type['params'][1];
+            if (empty($data[$lookupField])) {
+                throw new NonVisitableTypeException(sprintf('Union Discriminator Field "%s" not found in data', $lookupField));
             }
+
+            $unionMap = $type['params'][2];
+            $lookupValue = $data[$lookupField];
+            if (empty($unionMap[$lookupValue])) {
+                throw new NonVisitableTypeException(sprintf('Union Discriminator Map does not contain key "%s"', $lookupValue));
+            }
+
+            $finalType = [
+                'name' => $unionMap[$lookupValue],
+                'params' => [],
+            ];
+
+            return $context->getNavigator()->accept($data, $finalType);
         }
 
-        if (null !== $finalType && null !== $finalType['name']) {
-            return $context->getNavigator()->accept($data, $finalType);
-        } else {
-            foreach ($type['params'] as $possibleType) {
-                $finalType = null;
-
-                if ($this->isPrimitiveType($possibleType['name']) && $this->testPrimitive($data, $possibleType['name'], $context->getFormat())) {
-                    return $context->getNavigator()->accept($data, $possibleType);
-                }
+        foreach ($type['params'][0] as $possibleType) {
+            if ($this->isPrimitiveType($possibleType['name']) && $this->testPrimitive($data, $possibleType['name'], $context->getFormat())) {
+                return $context->getNavigator()->accept($data, $possibleType);
             }
         }
 
@@ -113,7 +98,7 @@ final class UnionHandler implements SubscribingHandlerInterface
 
     private function matchSimpleType(mixed $data, array $type, Context $context): mixed
     {
-        foreach ($type['params'] as $possibleType) {
+        foreach ($type['params'][0] as $possibleType) {
             if ($this->isPrimitiveType($possibleType['name']) && !$this->testPrimitive($data, $possibleType['name'], $context->getFormat())) {
                 continue;
             }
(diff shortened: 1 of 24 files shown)


idbentley:
Hi @goetas - I wonder if you've had a chance to look at these changes?

src/Handler/UnionHandler.php
```diff
@@ -47,46 +48,90 @@ public function serializeUnion(
...
+                            'params' => [],
+                        ];
+                    }
+                }
+            }
+
+            if (null !== $finalType && null !== $finalType['name']) {
+                return $context->getNavigator()->accept($data, $finalType);
+            } else {
+                foreach ($type['params'] as $possibleType) {
```
goetas, line 102: something is odd with this `foreach` loops, you have here a nested loop the same variable (line 70 and 102). shouldn't they be independent?

goetas, line 102: something is odd with this foreach loops, you have here a nested loop the same variable (line 70 and 102). shouldn't they be independent?

src/Handler/UnionHandler.php
```diff
@@ -47,46 +48,90 @@ public function serializeUnion(
...
+                            $finalType = [
+                                'name' => $propertyMetadata->unionDiscriminatorMap[$lkup],
+                                'params' => [],
+                            ];
+                        } else {
+                            throw new NonVisitableTypeException('Union Discriminator Map does not contain key \'' . $lkup . '\'');
+                        }
+                    } else {
+                        $finalType = [
+                            'name' => $lkup,
```
goetas, line 92: i think that is an anti pattern to leak class name into the JSON representation.... it might be a "quick thing" but it has backfired at me so many times...

are you using this approach? (as example in the "inheritance" discriminator map this feature is intentionally requiring the discriminator map to be not empty)

src/Metadata/Driver/TypedPropertiesDriver.php
```diff
@@ -102,6 +102,10 @@ public function loadMetadataForClass(ReflectionClass $class): ?ClassMetadata
         foreach ($classMetadata->propertyMetadata as $propertyMetadata) {
             // If the inner driver provides a type, don't guess anymore.
             if ($propertyMetadata->type) {
+                if ('union' === $propertyMetadata->type['name']) {
+                    $propertyMetadata->setType($this->reorderTypes($propertyMetadata->type));
```
goetas, line 106: you could potentially encode in the `union` type already the info regarding the discriminator field and map.
advantage of it is that it would be done only once at compile cache metadata time vs at runtime as done now.
it should be much faster.


```php
#[UnionDiscriminator(
  field: 'objectType', 
  map: [
   'author' => 'JMS\Serializer\Tests\Fixtures\MappedDiscriminatedAuthor', 
   'comment' => 'JMS\Serializer\Tests\Fixtures\MappedDiscriminatedComment'
  ]
)]
```
can be translated to 

```php

[
  'type' => 'union',
  'params' => [
      'objectType',
       [
        'author' => 'JMS\Serializer\Tests\Fixtures\MappedDiscriminatedAuthor', 
        'comment' => 'JMS\Serializer\Tests\Fixtures\MappedDiscriminatedComment'
       ]

  ]
]
```

in this way you have all the data ready in the `UnionHandler` and no need to fetch metadata each time


do you see it as doable approach?

goetas (commented):
hi, sorry for the delay in my review. i have left some comments, could you answer my questions please?

after you answer to https://github.com/schmittjoh/serializer/pull/1553#discussion_r1710356286, could you please update the documentation as well? most of what you need is to update `/doc/reference/annotations.rst`

src/Handler/UnionHandler.php
idbentley: This was an error - removing the inner loop is the correct fix.

src/Handler/UnionHandler.php
idbentley: I'm not sure I understand the alternative to putting the className into the type.

This can happen in many ways.  The simplest I can think of is:

```
class SomeClass {
    private MappedDiscriminatedAuthor $data
}
```

will result in such a type:
```
$type = ['name' => 'MappedDiscriminatedAuthor', 'params' => []]
```

I think I must be misunderstanding your question.

goetas:
Your latest version is exactly what I had in mind, nice!

The only thing left in my opinion is to address the potential security issue. With your implementation,

```php
class ComplexDiscriminatedUnion
{
    #[UnionDiscriminator(field: 'type')]
    private DiscriminatedAuthor|DiscriminatedComment $data;
```


```json

{
 "type":  "DOMDocument"
}
```

with the current implementation the serializer will attempt to instantiate a `DOMDocument` class and to set it to `$data`.

In this case `DOMDocument` is harmless, but any class can be passed there.

src/Metadata/Driver/AnnotationOrAttributeDriver.php
```diff
@@ -258,6 +259,8 @@ public function loadMetadataForClass(\ReflectionClass $class): ?BaseClassMetadat
                         $propertyMetadata->xmlAttributeMap = true;
                     } elseif ($annot instanceof MaxDepth) {
                         $propertyMetadata->maxDepth = $annot->depth;
+                    } elseif ($annot instanceof UnionDiscriminator) {
+                        $propertyMetadata->setUnionDiscriminator($annot->field, $annot->map);
```
goetas, line 263: do not forget to update the type as well with the metadata known in this step

idbentley:
@goetas thanks for the feedback. I think I captured all the changes you asked for!

src/Handler/UnionHandler.php
```diff
@@ -47,46 +48,91 @@ public function serializeUnion(
...
+                if (!empty($unionMap)) {
+                    if (array_key_exists($lkup, $unionMap)) {
+                        $finalType = [
+                            'name' => $unionMap[$lkup],
+                            'params' => [],
+                        ];
+                    } else {
+                        throw new NonVisitableTypeException('Union Discriminator Map does not contain key \'' . $lkup . '\'');
+                    }
+                } else {
```
goetas, line 90: this `else` branch still exposes the ability to pass any class name in the json and later that class wiill be instantiated.  I think that this should be removed

goetas:
@idbentley  I took the liberty of pushing the changes my self, could you please have a look?

idbentley:
@goetas Your changes look good to me!

idbentley:
I'm going to be OOTO next week.  Feel free to proceed without me!

src/Metadata/PropertyMetadata.php
```diff
@@ -33,6 +33,16 @@ class PropertyMetadata extends BasePropertyMetadata
      */
     public $serializedName;
 
+    /**
+     * @var string|null
+     */
+    public $unionDiscriminatorField;
```
scyzoryck, line 39: When we moved the code to the type, those fields seems to be not needed anymore. :)

goetas: hmm, can you please clarify? im not sure I understand

PR #1559, opened 2024-09-24
DateTime - Deserialize from more formats (default)

iKsSs:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

This change allows the setting of multiple default deserialization formats of Date.

### DIFF
--- a/src/Handler/DateHandler.php
+++ b/src/Handler/DateHandler.php
@@ -19,6 +19,11 @@ final class DateHandler implements SubscribingHandlerInterface
     private $defaultFormat;
 
     /**
+     * @var array<string>
+     */
+    private $defaultDeserializationFormats;
+
+    /**
      * @var \DateTimeZone
      */
     private $defaultTimezone;
@@ -69,9 +74,17 @@ final class DateHandler implements SubscribingHandlerInterface
         return $methods;
     }
 
-    public function __construct(string $defaultFormat = \DateTime::ATOM, string $defaultTimezone = 'UTC', bool $xmlCData = true)
-    {
+    /**
+     * @param array<string> $defaultDeserializationFormats
+     */
+    public function __construct(
+        string $defaultFormat = \DateTime::ATOM,
+        string $defaultTimezone = 'UTC',
+        bool $xmlCData = true,
+        array $defaultDeserializationFormats = [\DateTime::ATOM]
+    ) {
         $this->defaultFormat = $defaultFormat;
+        $this->defaultDeserializationFormats = $defaultDeserializationFormats;
         $this->defaultTimezone = new \DateTimeZone($defaultTimezone);
         $this->xmlCData = $xmlCData;
     }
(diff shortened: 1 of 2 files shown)


goetas:
Does this addition need to be documented?

iKsSs:
Who can take it further?

src/Handler/DateHandler.php
```diff
@@ -69,9 +74,17 @@ public static function getSubscribingMethods()
...
+     * @param array<string> $defaultDeserializationFormats
+     */
+    public function __construct(
+        string $defaultFormat = \DateTime::ATOM,
+        string $defaultTimezone = 'UTC',
+        bool $xmlCData = true,
+        array $defaultDeserializationFormats = [\DateTime::ATOM]
+    ) {
         $this->defaultFormat = $defaultFormat;
+        $this->defaultDeserializationFormats = $defaultDeserializationFormats;
```
scyzoryck, line 87: If I understand correctly in the past `$defaultFormat` was working for both - serialisation & deserialisation. With the changes that you proposed I can see small BC break - when somebody was using $defaultFormat with different format after updating the package, they will get `DateTime::ATOM` as a deserialisation format what might be small BC. I guess something like might help to keep it consistent with old behaviour when no `defaultDeserializationFormats` are passed. 
```suggestion
        bool $xmlCData = true,
        array $defaultDeserializationFormats = []
    ) {
        $this->defaultFormat = $defaultFormat;
        $this->defaultDeserializationFormats = $defaultDeserializationFormats === [] ?  [$defaultFormat] : $defaultDeserializationFormats;
```

src/Handler/DateHandler.php
```diff
@@ -285,10 +298,14 @@ private function getDeserializationFormats(array $type): array
             return is_array($type['params'][2]) ? $type['params'][2] : [$type['params'][2]];
         }
 
-        return [$this->getFormat($type)];
+        if (isset($type['params'][0])) {
+            return is_array($type['params'][0]) ? $type['params'][0] : [$type['params'][0]];
```
scyzoryck, line 302: Looking at the old code:
```php
    private function getFormat(array $type): string
    {
        return $type['params'][0] ?? $this->defaultFormat;
    }
```
`$type['params'][0]` will be always `string`.

scyzoryck (commented):
@iKsSs In general - great work! It looks like we do not have a documentation for the Handlers - So creating docs in bundle sounds ok for me. 
I left 2 small comments - can you take a look at them please? :) 

Best, Marcin

src/Handler/DateHandler.php
iKsSs: Ou, you are very correct. In my case, it would be BC. Thanks for providing me with a solution for this.
I was thinking about renaming private property `defaultFormat` to `defaultSerializationFormat`, what do you think?

src/Handler/DateHandler.php
iKsSs: Yeah, that's true => I removed redundant code, thanks

src/Handler/DateHandler.php
scyzoryck: Renaming sounds like a good solution 👍

iKsSs: OK, done

scyzoryck:
I will release it on Monday! Thanks for the contribution!

iKsSs:
> I will release it on Monday! Thanks for the contribution!

What Monday? :smile:

scyzoryck:
@iKsSs Sorry, I got some urgent stuff last week. Released today! 🎉

PR #1561, opened 2024-10-11
Do not use single quotes in enum type definition

bobvandevijver:
| Q             | A
| ------------- | ---
| Bug fix?      | Inconsistency fix
| New feature?  | no
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | yes 
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

When enum support was added, it was chosen to use quotes around the type in its type annotation, for example `enum<'Color'>`. However, when compared with the other type annotations such as array and iterable, those quotes are inconsistent.

This PR makes the enum type annotation behaviour similar to the others and deprecated passing a direct string. This deprecation is only triggered during deserialization, as the serialization path has never used the type parameter.

I found this while using NelmioApiDocBundle, see https://github.com/nelmio/NelmioApiDocBundle/issues/2327. I am also making an update for that bundle so it can support both syntaxes.

### DIFF
--- a/tests/Fixtures/ObjectWithEnums.php
+++ b/tests/Fixtures/ObjectWithEnums.php
@@ -12,32 +12,34 @@ use JMS\Serializer\Tests\Fixtures\Enum\Suit;
 class ObjectWithEnums
 {
     /**
-     * @Serializer\Type("enum<'JMS\Serializer\Tests\Fixtures\Enum\Suit', 'name'>")
+     * @Serializer\Type("enum<JMS\Serializer\Tests\Fixtures\Enum\Suit, 'name'>")
      */
     public Suit $ordinary;
 
     /**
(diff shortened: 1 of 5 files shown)


tests/Fixtures/ObjectWithEnums.php
```diff
@@ -12,32 +12,34 @@
 class ObjectWithEnums
```
scyzoryck, line 12: Just thinking out loud - how it is going to work with Attributes? I guess we might miss some test cases for it but it might be used by some users as well.

bobvandevijver: I believe it is just missing tests for Attributes on this particular level. I've said goodbye to annotations quite a while back, so I hit this while using Attributes and verified that for me this solution works.

goetas (approved):
looks good. tbh i forgot that it would work without quotes

scyzoryck:
I will release it on Monday.

PR #1569, opened 2024-11-11
Support value types in serialization

### LINKED ISSUE #1568: Error serializing union types between class and boolean value types
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

As of 3.31.0, if you try to serialize a method that is a union type between a class a boolean value type (e.g. `MyClass|false`), JMS Serializer raises an exception.

This is because the value type [is filtered out by `TypedPropertiesDriver->shouldTypeHint()`](https://github.com/schmittjoh/serializer/blob/6226197b0eadab798b2101d8259f03a8707892d3/src/Metadata/Driver/TypedPropertiesDriver.php#L164-L166) (because `true` and `false` aren't in the allowlist) so the only acceptable value becomes "MyClass".

```php
class SubClass {
}

class TestClass {
    public SubClass|false $falsableProp = false;
}

$serializer = JMS\Serializer\SerializerBuilder::create()->build();
$jsonContent = $serializer->serialize(new TestClass(), 'json');
echo $jsonContent;
```

## Expected Result

```php
{"falsable_prop":false}
```

## Actual Result

Exception thrown:

> JMS\Serializer\Exception\InvalidArgumentException: Value at TestClass is expected to be an object of class SubClass but is of type boolean

## Workarounds

Changing the typehint to `MyClass|bool` fixes this, though means you're opening your type wider than needed.

## Fix

I believe this can be fixed with two changes:
- Adding `'true'` and `'false'` to [the TypedPropertiesDriver whitelist](https://github.com/schmittjoh/serializer/blob/6226197b0eadab798b2101d8259f03a8707892d3/src/Metadata/Driver/TypedPropertiesDriver.php#L71-L83)
- Adding `'true'` and `'false'` as cases to [the graph navigator boolean acceptor](https://github.com/schmittjoh/serializer/blob/master/src/GraphNavigator/SerializationGraphNavigator.php#L157-L159) (without, the navigator treats value types as objects)

I'm happy to raise a PR for this, but I'm not very familiar with the nuances of the codebase so I'm not confident this won't cause other issues.

gregtyler:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes (kinda, but I think this is how you'd expect the library to work)
| Doc updated   | no (not applicable)
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1568
| License       | MIT

Allows serialization of value types (i.e. `true` and `false`, as opposed to `bool`), including in union types (e.g. `string|false`).

Otherwise, tries to create class with FQCN "true" and, for union types, filters out value types as non-primitive.

Fixes #1568

### DIFF
--- a/src/GraphNavigator/DeserializationGraphNavigator.php
+++ b/src/GraphNavigator/DeserializationGraphNavigator.php
@@ -128,6 +128,8 @@ final class DeserializationGraphNavigator extends GraphNavigator implements Grap
 
             case 'bool':
             case 'boolean':
+            case 'false':
+            case 'true':
                 return $this->visitor->visitBoolean($data, $type);
 
             case 'double':
--- a/src/GraphNavigator/SerializationGraphNavigator.php
+++ b/src/GraphNavigator/SerializationGraphNavigator.php
@@ -156,6 +156,8 @@ final class SerializationGraphNavigator extends GraphNavigator
 
             case 'bool':
             case 'boolean':
+            case 'true':
+            case 'false':
                 return $this->visitor->visitBoolean((bool) $data, $type);
 
             case 'double':
--- a/src/Metadata/Driver/TypedPropertiesDriver.php
+++ b/src/Metadata/Driver/TypedPropertiesDriver.php
@@ -75,6 +75,8 @@ class TypedPropertiesDriver implements DriverInterface
             'float',
             'bool',
             'boolean',
+            'true',
+            'false',
             'string',
             'double',
             'iterable',
--- a/tests/Fixtures/DataFalse.php
+++ b/tests/Fixtures/DataFalse.php
@@ -1,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JMS\Serializer\Tests\Fixtures;
+
+class DataFalse
+{
+    public function __construct(
+        public false $data
+    ) {
+    }
+}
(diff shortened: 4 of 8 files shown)


scyzoryck:
@gregtyler I added some end to end tests for serialisation. Unfortunately I was not able to push to your branch :( 
https://github.com/schmittjoh/serializer/commit/ea7f7be4cb6ca2d1ed5e156b57f1413be8936e82

@goetas @gregtyler there is one edge that we need to cover - what if we try to deserialize value `true` into `false` type. Looking at other types it seems like `NonCastableTypeException` should be thrown - we can check type inside `DeserializationVisitorInterface::visitBoolean()` method. WDYT?

gregtyler:
@scyzoryck I've copied your changes onto the PR (I couldn't figure out how to cherry-pick across forks unfortunately).

On the deserialization error, I think the PHP error message caused by type-breaking is already very effective:
> Cannot assign true to property JMS\Serializer\Tests\Fixtures\DataFalse::$data of type false

You could test for that in `visitBoolean` instead, but I think a useful error message would basically say exactly the same thing.

I added that to the test (to expect a TypeError), but I didn't add an expectation on the message because it's PHP internals and varies between versions.

scyzoryck:
Thanks! I will think a bit about exceptions in this case - but it should be merged this week :)

scyzoryck:
Soo, it looks like the handling of incorrect is inconsistent right now in deserialization - as sometimes `\RuntimeException` will be thrown, sometimes `JMS\Serializer\Exception\NonVisitableTypeException` depending on the type of data. 
Current solution sounds good for now, but looks like we need to try to organize it a bit better in the future.

PR #1572, opened 2024-11-21
Allow PHPStan PHPDoc parser 2.0

ruudk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |no
| Doc updated   |no
| BC breaks?    |no 
| Deprecations? |no
| Tests pass?   | yes]
| Fixed tickets | 
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "doctrine/instantiator": "^1.3.1 || ^2.0",
         "doctrine/lexer": "^2.0 || ^3.0",
         "jms/metadata": "^2.6",
-        "phpstan/phpdoc-parser": "^1.20"
+        "phpstan/phpdoc-parser": "^1.20 || ^2.0"
     },
     "suggest": {
         "doctrine/collections": "Required if you like to use doctrine collection types as ArrayCollection.",
@@ -37,6 +37,7 @@
     "require-dev": {
         "ext-pdo_sqlite": "*",
         "doctrine/annotations": "^1.14 || ^2.0",
+        "slevomat/coding-standard": "dev-master#f2cc4c553eae68772624ffd7dd99022343b69c31 as 8.11.9999",
         "doctrine/coding-standard": "^12.0",
         "doctrine/orm": "^2.14 || ^3.0",
         "doctrine/persistence": "^2.5.2 || ^3.0",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -37,17 +37,18 @@
     "require-dev": {
         "ext-pdo_sqlite": "*",
         "doctrine/annotations": "^1.14 || ^2.0",
+        "slevomat/coding-standard": "dev-master#f2cc4c553eae68772624ffd7dd99022343b69c31 as 8.11.9999",
```
ruudk, line 40: Waiting for https://github.com/slevomat/coding-standard/issues/1713

composer.json
```diff
@@ -37,17 +37,18 @@
...
         "doctrine/phpcr-odm": "^1.5.2 || ^2.0",
         "jackalope/jackalope-doctrine-dbal": "^1.3",
         "ocramius/proxy-manager": "^1.0 || ^2.0",
         "phpbench/phpbench": "^1.0",
-        "phpstan/phpstan": "^1.10.57",
+        "phpstan/phpstan": "^2.0",
         "phpunit/phpunit": "^9.0 || ^10.0 || ^11.0",
         "psr/container": "^1.0 || ^2.0",
-        "rector/rector": "^1.0.0",
+        "rector/rector": "dev-main#d593c5af24b9ca2da9b7595125dca713f6032bac as 2.0",
```
ruudk, line 51: Waiting for a tagged release: https://github.com/rectorphp/rector/issues/8815

ruudk:
It seems I have opened Pandora's box 🤯 

@goetas Could you help out? 🙏

goetas:
@ruudk could you clarify which help do you need?

ruudk:
@goetas I guess, I don't need help after all. It wasn't so bad.

ruudk:
Since these 2 are dev dependencies, would it be possible to pin them like I did now? 

Rector will be released on December 12. 

I have no idea about slevomat/coding-standard

Then at least we could remove this blocker for others that want to upgrade to PHPStan 2.

/cc @goetas

goetas:
I have no issue with the changes in the dev dependencies but I ask you kindly to create a new PR with those packages are released.

@scyzoryck art you ok with it?

goetas (approved):

composer.json
```diff
@@ -26,7 +26,7 @@
         "doctrine/instantiator": "^1.3.1 || ^2.0",
         "doctrine/lexer": "^2.0 || ^3.0",
         "jms/metadata": "^2.6",
-        "phpstan/phpdoc-parser": "^1.20"
```
scyzoryck, line 29: I do wonder if we can support both - 1.X & 2.x. Without it people that are using `slevomat/coding-standard` or `rector` will be not able to use new version of serializer. 
At the moment it is a blocker for me as we need to release some bug for the Union types - but I'm fine with releasing it with only 2.x support after releasing bug fix release.

scyzoryck (changes requested):

composer.json
samsonasik: that's not possible, as bc break signature, see change https://github.com/schmittjoh/serializer/pull/1572/files#diff-8daba71112686a0487131b9b0f2347a47887f14a6199c712eee72e587e2eba6e

scyzoryck: In theory we can have a hacky `if(class_exist(ParserConfig::class))`. It will look terrible but most probably it will do the job :)

ruudk: We can do that indeed :)

ruudk:
Anybody an idea about these new PHPStan errors?

samsonasik:
I think you can exclude `Fixtures` directory under excludesPaths

```neon
parameters:
    # other parameters

    excludePaths:
        - */Fixtures/*
```

scyzoryck (approved):

scyzoryck:
Looks good for me :) I will merge it before the next release.

samsonasik (approved):

PR #1582, opened 2024-12-19
Add true/false as primitive types

bobvandevijver:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

The new union handler does not consider `false` or `true` as a primitive type, causing my `array|false` with an array value to be serialised as `true`.


scyzoryck:
Good catch! Thanks!

scyzoryck:
@bobvandevijver - released! Enjoy!

Issue #1584, opened 2025-01-20
Fatal error since >= 3.32.1: `TypeError: TypedPropertiesDriver::shouldTypeHintInsideUnion`


nocive:
After the introduction of https://github.com/schmittjoh/serializer/pull/1571 we experience a fatal error in a previously working codebase while serializing a doctrine entity object:

```
"exception": {
  "class": "TypeError",
  "message": "JMS\\Serializer\\Metadata\\Driver\\TypedPropertiesDriver::shouldTypeHintInsideUnion(): Argument #1 ($reflectionType) must be of type ReflectionNamedType, ReflectionIntersectionType given, called in vendor/jms/serializer/src/Metadata/Driver/TypedPropertiesDriver.php on line 184",
  "code": "0",
  "file": "vendor/jms/serializer/src/Metadata/Driver/TypedPropertiesDriver.php:192",
  "trace": [
    "vendor/jms/serializer/src/Metadata/Driver/TypedPropertiesDriver.php:184",
    "vendor/jms/serializer/src/Metadata/Driver/TypedPropertiesDriver.php:115",
    "vendor/jms/metadata/src/Driver/LazyLoadingDriver.php:38",
    "vendor/jms/metadata/src/MetadataFactory.php:111",
    "vendor/jms/serializer-bundle/Debug/TraceableMetadataFactory.php:42",
    "vendor/jms/serializer/src/GraphNavigator/SerializationGraphNavigator.php:234",
    "vendor/jms/serializer/src/JsonSerializationVisitor.php:101",
    "vendor/jms/serializer/src/GraphNavigator/SerializationGraphNavigator.php:172",
    "vendor/jms/serializer/src/JsonSerializationVisitor.php:101",
    "vendor/jms/serializer/src/GraphNavigator/SerializationGraphNavigator.php:172",
    "vendor/jms/serializer/src/Serializer.php:256",
    "vendor/jms/serializer/src/Serializer.php:163",
    ....
```

I saw that there are already other issues reported related to unions (such as https://github.com/schmittjoh/serializer/issues/1566) but I didn't see any exhibiting the same symptoms so I'm reporting it just in case.

For now we're forced to pin 3.31.1 to work around the issue.

Thank you for the great work.

scyzoryck:
Hi @nocive ! 

Could you share some example how to reproduce the issue? I guess we need some specific property type, but not sure how to reproduce it 😓 

Best, Marcin.

nocive:
[...]

PR #1585, opened 2025-02-02
Do not typehint DNF union types

scyzoryck:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1584
| License       | MIT

Tests coverage based on: https://wiki.php.net/rfc/dnf_types

@goetas I need your opinion on that:
1. Current implementation skips any cases of Disjunctive Normal Form Types.
2. Alternative solution would be to skip only Intersection types, so `(A&B)|D` could be resolved into `D`, when `(A&B)|C|D` into `union<C,D>`

Looking at the use cases like ArrayCollections from Doctrine I guess second option might be more intuitive for users, when first safer as it brings back behavior from prev lib versions. 🤔

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -107,6 +107,10 @@
         <exclude-pattern>tests/*</exclude-pattern>
     </rule>
 
+    <rule ref="Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps">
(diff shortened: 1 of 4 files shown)


goetas:
I'm fine with not supporting such expressions.

I really hope no one uses them! :D

PR #1587, opened 2025-02-20
Add global handlers for all data types

MoviU:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | yes
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| License       | MIT

Allow custom handlers for all types. ("type" => "*").

### DIFF
--- a/doc/handlers.rst
+++ b/doc/handlers.rst
@@ -83,3 +83,50 @@ Skippable Subscribing Handlers
 
 In case you need to be able to fall back to the default deserialization behavior instead of using your custom
 handler, you can simply throw a `SkipHandlerException` from you custom handler method to do so.
+
+Global Subscribing Handlers
+----------------
+You can register global handler, if type is '*'::
+
+    use JMS\Serializer\Handler\SubscribingHandlerInterface;
+    use JMS\Serializer\GraphNavigator;
+    use JMS\Serializer\JsonSerializationVisitor;
+    use JMS\Serializer\JsonDeserializationVisitor;
+    use JMS\Serializer\Context;
+
+    class MyHandler implements SubscribingHandlerInterface
+    {
+        public static function getSubscribingMethods()
+        {
+            return [
+                [
+                    'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
+                    'format' => 'json',
+                    'type' => '*',
+                    'method' => 'handleCustomSerialization',
+                ],
+                [
+                    'direction' => GraphNavigator::DIRECTION_DESERIALIZATION,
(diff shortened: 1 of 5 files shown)


scyzoryck:
Hello. 
What would be a usage case for it? 

Best, Marcin.

MoviU:
Hi,

In some cases, the default serialization behavior needs to be customized. For example, in my current use case, if a DTO is marked with a specific attribute (e.g., #[CustomDTO]), I need to apply custom serialization logic—such as inspecting property attributes and executing specific logic when a custom attribute is found.

My current implementation maps a handler to every DTO within a given namespace. This approach relies heavily on Composer’s autoloading and requires cache revalidation whenever changes occur in that namespace (e.g., when a new DTO is created).

Allowing handlers to apply to all types would enable seamless customization of the default serialization process without the need for complex workarounds.

scyzoryck:
I think adding custom Handler for all classes will create multiple issues - for example with depth exclusions that might not work as expected. 
May event system may solve your issues? https://github.com/schmittjoh/serializer/blob/033c9beab9eb708509a3d400e9f0ffeb2d440e71/doc/event_system.rst#serializerpre_serialize
You can change the type to `*` for any class that needs to handled by your custom Handler.

MoviU:
The issue I encountered with events during deserialization is that they operate on an already deserialized instance.

If the DTO is **readonly**, modifying its properties directly isn't possible. Instead, I would need to create a **new instance** with the updated data.

However, the event system doesn't allow replacing the deserialized result.

Maybe there is another way to achieve this? What do you think?

scyzoryck:
Have you checked `serializer.pre_deserialize` event? It operates on raw data.

MoviU:
Yes, I tried using pre_deserialize also. And I think it still isn't useful in my use-case, since I can't change the data directly

MoviU:
Hi! Any updates?

scyzoryck:
Can you try the following please? 
In `serializer.pre_deserialize` event change type of 
```php
    public function onPreSerializeEnum(PreSerializeEvent $event): void
    {
        $oldType = $event->getType();
        $event->setType('*', $oldType);
    }
```

Register custom handler for `*` type ;-) 

Soo in that way I believe you should achieve your goals. :)

MoviU:
Hi! Any updates?

scyzoryck:
Hi! 
I'm still not convinced to this idea. @goetas any thoughts from your side on that? 

Best.

MoviU:
Hi, @goetas! Any updates?

PR #1599, opened 2025-05-15
Parse list into `list` type instead of `array`

simPod:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc updated   | no
| BC breaks?    | yes
| Deprecations? | no 
| Tests pass?   | yes
| License       | MIT

I noticed in tests we already expect `list<>` input for serialization https://github.com/schmittjoh/serializer/blob/4dc94aa9093f1c0447c10544f5b645dc675bf856/tests/Serializer/JsonSerializationTest.php#L403 therefore I deduced we can use `list` type internally.


scyzoryck:
Sounds reasonable. I will release it on Monday.

PR #1600, opened 2025-05-17
Test for null or empty strings when deserializing dates

gnat42:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | not sure
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #... 
| License       | MIT

Some APIs send empty dates instead of null or omitting the field. For example
```
{
"someDateField": ""
}
```

Currently only `"someDateField": null` works. Changing the null test to an empty test allows for the system to skip deserializing empty date fields.

### DIFF
--- a/src/Handler/DateHandler.php
+++ b/src/Handler/DateHandler.php
@@ -211,7 +211,7 @@ final class DateHandler implements SubscribingHandlerInterface
      */
     public function deserializeDateTimeFromJson(DeserializationVisitorInterface $visitor, $data, array $type): ?\DateTimeInterface
     {
-        if (null === $data) {
+        if (empty($data)) {
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


scyzoryck:
Hi! 
Thanks for the Pull Request! It looks like an interesting edge case. I wonder what is current outcome of the empty string or `0` deserialization? 
Can you cover this case with some additional test cases? I wonder if it makes sens to make it as an optin feature - as it can be potentially and BC for some projects (depending on how it behaves currently). 😓

gnat42:
I've added tests.

I was wondering about the BC, I'm not 100% sure, however I would think that anyone running into this issue may have created their own DateTime handler for deserialization. If they didn't they would get an exception from line 275 of DateHandler.php, which is what happened to us. The change makes those custom handlers obsolete if that is all they do. Otherwise, I'm not sure what behaviour one would expect from this. I get its a change so I'm fine either way but it seems like a safe fix imo

scyzoryck:
Thanks for contribution!

PR #1602, opened 2025-05-19
XML Serialization: fix #1601

### LINKED ISSUE #1601: XML Serialization: SerializedName("ElementName/@AttributeName") with XmlElement does not create an attribute
If a PHP class property is annotated with `#[JMS\XmlElement]` and `#[JMS\SerializedName("ElementName/@AttributeName")]`, the `XmlSerializationVisitor` attempts to create an XML element literally named `ElementName/@AttributeName` instead of creating an element `ElementName` with an attribute `AttributeName`. This results in an "invalid tag name" error because `/` and `@` are not valid characters in XML.

Interestingly, the deserialization process (`XmlDeserializationVisitor`) correctly handles this syntax, mapping an attribute from the XML to the corresponding PHP property using XPath. However, serialization fails to mirror this behavior.

To Reproduce Consider the following PHP class structure:

```php
use JMS\Serializer\Annotation as Serializer;

/** @Serializer\XmlRoot("test-object") */
class MessageSenderDetails
{
    /**
     * @Serializer\SerializedName("FromIdentifier")
     * @Serializer\Type("string")
     * @Serializer\XmlElement(cdata=false)
     */
    public string $fromIdentifier;

    /**
     * @Serializer\SerializedName("FromIdentifier/@SchemeID")
     * @Serializer\Type("string")
     * @Serializer\XmlElement(cdata=false) // This, combined with SerializedName, is key
     */
    public ?string $fromIdentifierSchemeID = null;
}

// --- Test Code ---
// $serializer = SerializerBuilder::create()->build();
// $details = new MessageSenderDetails();
// $details->fromIdentifier = "SomeID";
// $details->fromIdentifierSchemeID = "Scheme123";
//
// $xml = $serializer->serialize($details, 'xml'); // This line throws an exception
```

Expected behavior The serializer should produce XML similar to:

```xml
<test-object>
  <FromIdentifier SchemeID="Scheme123">SomeID</FromIdentifier>
</test-object>
```

Actual behavior: The serialization process throws an exception for the reasons mentioned at the start of this issue.

Environment

 • JMS Serializer version: 3.x/master
 • PHP version: 8.4

As additional context issue arises because `XmlSerializationVisitor::visitProperty` currently does not have specific logic to interpret the `ElementName/@AttributeName` pattern in `serializedName` (when xmlAttribute metadata is false and XmlElement is used) as an instruction to set an attribute on a previously serialized sibling element. Instead, it falls through to the standard element creation logic.

I plan to submit a Pull Request with an implementation for this feature shortly.

NikoGrano:
| Q             | A
| ------------- | ---
| Bug fix?      | no _(/yes)_
| New feature?  | yes
| Doc updated   |no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1601 
| License       | MIT

This PR introduces a new capability to the XML serialization process, allowing properties annotated with `#[XmlElement]` to define attributes on sibling XML elements using the `#[SerializedName("ElementName/@AttributeName")]` syntax.

**Problem:**

Previously, when a property was configured with `#[XmlElement]` and `#[SerializedName("ElementName/@AttributeName")]`, the serializer would attempt to create an XML element literally named `ElementName/@AttributeName`. This would lead to a `DOMException` due to invalid characters in the element name. This behavior was inconsistent with the deserialization process, which correctly interprets this syntax to map an XML attribute to a PHP property. This issue is described in #1601.

**Solution:**

The `XmlSerializationVisitor::visitProperty` method has updated: (_simplified/tl;dr_)
1.  It now checks if a property's `serializedName` contains the `/@` pattern
2.  Conditions are met -> invoke new method `trySerializePropertyAsAttributeOnSiblingElement`.
3.  New method parses the target element name and attribute from `serializedName`.
4. Try to handle namespaces and target element correctly 😅 
5. If the property's value is `null`, the attribute is not rendered.
6.  If the property's value is `null`, the attribute is not rendered, aligning with common XML practices.
7. Implemented two new tests to serialize and assert to expected xml output. (As bonus, trying to deserialize generated xml for consistency, but there is something weird about it when using namespaces, which is not related to this issue)

**Hopefully, this can be checked quickly, so I know how I continue with my own project using this.**


NikoGrano:
Seems like [Scrutinizer](https://scrutinizer-ci.com/g/schmittjoh/serializer/inspections/330f110f-f0fb-443d-90b5-6a7193ae4a26) is somehow broken.

Could somebody restart it or just ignore it...?

tests/Fixtures/ObjectWithElementAttributeSyntax.php
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * @param string $identifierValue
+     * @param string $identifierScheme
+     * @param string $customElementValue
+     * @param string $customElementType
+     * @param string $dataPointValue
+     */
+    public function __construct(
+        $identifierValue,
```
scyzoryck, line 94: We can go with argument types - minimal supported PHP version is 7.4 now :)

NikoGrano:
Thanks for the review!

The deserialization side of this feature already exists, as the serializer passes the name as an [xpath](https://www.php.net/manual/en/simplexmlelement.xpath.php) query. I wasn't aware of this until [Mirgen](https://stackoverflow.com/users/1376033/mirgen) pointed it out in this [answer](https://stackoverflow.com/a/71031000).

However, there is an issue with deserialization when namespaces are involved. Using this example:
https://github.com/schmittjoh/serializer/blob/3ecd0949e04410050a6494e128971491d995680a/tests/Fixtures/ObjectWithElementAttributeSyntax.php#L44-L49

With the current PR, we can serialize `$customElementValue` successfully, but deserializing it results in a null/uninitialized value (depending on typed properties). This only happens when a namespace is present on `#[XmlElement(namespace: ...)]` - **it works fine without namespaces.**

In the tests, we first serialize the entity to XML, compare it to the expected result, then try to deserialize back into an object. We had to add this bypass to set `customElementType` to null, otherwise tests would fail:
https://github.com/schmittjoh/serializer/blob/3ecd0949e04410050a6494e128971491d995680a/tests/Serializer/XmlSerializationTest.php#L662-L664

I'd like to address this deserialization issue separately to keep this PR focused on the serialization side. In my use case, I don't need namespaces, so I'm comfortable with the current implementation. This is an edge case that likely won't affect 99.9% of users, but it would be good to document it or open an issue about it to track the namespace deserialization problem and possible interest fixing it.

For documentation, I'll add it as requested. Since this extends an "accidental feature" that was already working for deserialization, this PR completes the picture by providing serialization support. Do you have specific recommendations for how you'd like the documentation structured? I don't have any clue how to document this.

I'll implement the requested argument updates soonish after I hear back from you.

scyzoryck:
Sounds like a plan! Sounds like we will be able to release it next week :)

NikoGrano:
I continued working on this and got a bit carried away implementing the deserialization side as well, since I could see potential issues arising from incomplete support later.

I've fixed the deserialization issues with namespaces and added implementation to serialize attributes into the current object, essentially adding serialization support to this [answer](https://stackoverflow.com/a/61120963).

While testing the `@` and `@/` style names, I encountered several deserialization issues when using namespaces. I've addressed these in commit [e1334734](https://github.com/schmittjoh/serializer/pull/1602/commits/e1334734991af5ba75c1402f4cb452f4d28e65df).

**TLDR - After the latest changes, this PR now has full support for:**

1. `@AttributeName` syntax: Serializes/deserializes a property as an attribute on the *current* XML element
2. `ElementName/@AttributeName` syntax: Serializes/deserializes a property as an attribute on a *sibling* XML element named `ElementName`

I'll work on documenting this feature now.

Issue #1596, opened 2025-04-18
Undefined array key "params"


Magiczne:
When using `JsonDeserializationVisitor` we encounter an error: `Undefined array key "params"`:

https://github.com/schmittjoh/serializer/blob/master/src/JsonDeserializationVisitor.php on line 101:
```php
        if (!$type['params']) {
            return $data;
        }
```
should check for existence of the params key:
```php
       if (!isset($type['params']) || !$type['params']) {
            return $data;
        }
```

scyzoryck:
Hi! 
Do you have an example how to recreate the issue? At the moment all types should have this property. Missing params property may have impact also in other parts of system.

Magiczne:
Hi. Will try to provide reproduction example or at least tell you more about the setup tomorrow

Magiczne:
@scyzoryck your messages led us to additional investigation on our side, and it was issue inside custom handler implementing `SubscribingHandlerInterface`.

We wre calling `visitArray` manually, and did not pass `params` array - and since there are no requirements to pass that array we thought it is good.

Maybe the check for field existence should be there? Or at least some info in the documentation that `params` even as an empty array

Magiczne:
Yes, phpstan type would also prevent us from having this issue so that's a nice idea

scyzoryck:
PHPstan types merged to master. Closing issue:)

PR #1603, opened 2025-05-30
Params of TypeArray are optional

alexander-schranz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | no
| Fixed tickets | #... 
| License       | MIT

The params where at current state not required and so should be marked as optional. Stumbled over this after getting PHPStan error in @sulu code base.

Related: https://github.com/schmittjoh/serializer/pull/1597

### DIFF
--- a/src/Type/Type.php
+++ b/src/Type/Type.php
@@ -7,7 +7,7 @@ namespace JMS\Serializer\Type;
 /**
  * @internal
  *
- * @phpstan-type TypeArray array{name: string, params: array}
+ * @phpstan-type TypeArray array{name: string, params?: array}
  */
 final class Type
 {


scyzoryck:
Hi! 
I'm sorry for the changes. Internally JMS Serializer always defines `params` and never checks for its existence. Not passing it may cause some `Undefined array key` errors - see: https://github.com/schmittjoh/serializer/issues/1596 
Would it possible to add `params` key with empty array, to Sulu codebase, please? It seems like there is only one its usage.

PR #1606, opened 2025-07-29
Add default class for discriminator by vsychov

ivan1986:
This is update of #707 for latest version

### DIFF
--- a/src/Metadata/Driver/XmlDriver.php
+++ b/src/Metadata/Driver/XmlDriver.php
@@ -124,7 +124,7 @@ class XmlDriver extends AbstractFileDriver
                 $discriminatorFieldName,
                 $discriminatorMap,
                 $discriminatorGroups,
-                !empty($discriminatorDefaultClass) ? $discriminatorDefaultClass : null,
+                $discriminatorDefaultClass ?? null,
             );
         }
 
--- a/src/Metadata/Driver/XmlDriver.php
+++ b/src/Metadata/Driver/XmlDriver.php
@@ -124,7 +124,7 @@ class XmlDriver extends AbstractFileDriver
                 $discriminatorFieldName,
                 $discriminatorMap,
                 $discriminatorGroups,
-                $discriminatorDefaultClass ?? null,
+                $discriminatorDefaultClass ?: null,
             );
         }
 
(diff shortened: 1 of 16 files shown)


src/Metadata/Driver/XmlDriver.php
```diff
@@ -119,7 +120,12 @@ protected function loadMetadataFromFile(\ReflectionClass $class, string $path):
                 $discriminatorGroups[] = (string) $entry;
             }
 
-            $metadata->setDiscriminator($discriminatorFieldName, $discriminatorMap, $discriminatorGroups);
+            $metadata->setDiscriminator(
+                $discriminatorFieldName,
+                $discriminatorMap,
+                $discriminatorGroups,
+                !empty($discriminatorDefaultClass) ? $discriminatorDefaultClass : null,
```
scyzoryck, line 127: Let's make it a bit shorter. 
```suggestion
               $discriminatorDefaultClass ?? null,
```

tests/Serializer/JsonSerializationTest.php
```diff
@@ -120,7 +120,7 @@ protected static function getContent($key)
             $outputs['post'] = '{"type":"post","title":"Post Title"}';
             $outputs['image_post'] = '{"type":"image_post","title":"Image Post Title"}';
             $outputs['image_post_without_type'] = '{"title":"Image Post Title"}';
-            $outputs['garage'] = '{"vehicles":[{"km":3,"type":"car"},{"km":1,"type":"moped"}]}';
+            $outputs['garage'] = '{"vehicles":[{"km":3,"type":"car"},{"km":1,"type":"moped"},{"km":1,"type":"JMS\\\Serializer\\\Tests\\\Fixtures\\\Discriminator\\\Other"}]}';
```
scyzoryck, line 123: Just thinking about the use cases. I do wonder if we should also provide some name like we have in the map - as FQN class names looks weird with other names.

scyzoryck:
Beside one comment the code looks good for me.

tests/Serializer/JsonSerializationTest.php
ivan1986: oh, it second issue with determinator - without this is hard to migrate old code to new with use determinator - need change all code in once

```
'The discriminator field name "%s" of the base-class "%s" conflicts with a regular property of the sub-class "%s".',
```

why can not use exist field in base class as determinator?

For example now in our site we have cart with many products and we use one class CartItem with field type

ivan1986: hm, or all this poll have sence only with implement virtual determinator?

scyzoryck: Good question. May @goetas have some ideas how it should work, as he has much more experience with discriminators.

ivan1986:
@scyzoryck 
any actions for this and #1609  ?

PR #1628, opened 2026-05-27
Fix method not found when Symfony/Component/Clock/DatePoint exists in…

michilehr:
… DateHandler

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #1627 
| License       | MIT

PR #1630, opened 2026-08-21
Keep fractional seconds when formatting DateInterval

semx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc updated   | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT

`DateHandler::format()` builds the ISO 8601 duration from `y`, `m`, `d`, `h`, `i` and `s` only and never reads the `f` property, although `parseDateInterval()` specifically parses fractional seconds into `f`. Fractional seconds are silently dropped on serialization:

```php
$handler = new DateHandler();

$i = new DateInterval('PT10S');
$i->f = 0.5;
$handler->format($i); // "PT10S", 0.5s lost

// round trip
$parsed = /* deserialize */ 'PT0.5S';   // s=0, f=0.5
$handler->format($parsed);              // "P0DT0S" - the value vanishes entirely
```

The `f` property is also populated natively, so intervals coming from `DateTime::diff()` lose their sub-second part the same way:

```php
$a = new DateTimeImmutable('2021-01-01 00:00:00.000000');
$b = new DateTimeImmutable('2021-01-01 00:00:10.500000');
$handler->format($a->diff($b)); // "PT10S" instead of "PT10.5S"
```

This PR renders the seconds component from `s + f` when `f` is non-zero (trimming the trailing zeros of the fixed-point `%.6F` representation, which matches DateInterval's native microsecond resolution), and lets a non-zero `f` alone introduce the `T` separator so `PT0.5S` no longer collapses to `P0DT0S`.

Full suite passes (1139 tests), phpstan and phpcs are clean on the touched files.

### DIFF
--- a/src/Handler/DateHandler.php
+++ b/src/Handler/DateHandler.php
@@ -393,7 +393,7 @@ final class DateHandler implements SubscribingHandlerInterface
             $format .= $dateInterval->d . 'D';
         }
 
-        if (0 < $dateInterval->h || 0 < $dateInterval->i || 0 < $dateInterval->s) {
+        if (0 < $dateInterval->h || 0 < $dateInterval->i || 0 < $dateInterval->s || 0 < $dateInterval->f) {
             $format .= 'T';
         }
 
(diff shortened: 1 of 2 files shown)


scyzoryck:
Hi @semx! 
Can you rebase your branch please?

semx:
Rebased onto current master (the branch now sits on top of #1628). The DateHandler change applies cleanly over the recent DatePoint fix, and the full Date test group is green locally on PHP 8.5. Ready for another look.

scyzoryck:
Thank you for contribution!

